PR #5, opened 2020-07-02
kickstart driver script + build list for pilot setup (WIP)

boegel:
WIP because various bits are still missing:

* [ ] properly configuring EasyBuild (install prefix, RPATH, etc.)
* [ ] actually installing what's specified in the build list
* [ ] tests

### DIFF
--- a/eessi-build-list-2020.06.py
+++ b/eessi-build-list-2020.06.py
@@ -1,0 +1,18 @@
+eessi_2020_06 = {
+    'GROMACS': {
+        '2020.1-foss-2020a-Python-3.8.2': {
+            'except': ['gpu'],
+        },
+    },
+    'OpenFOAM': {
+        '7-foss-2019b': {},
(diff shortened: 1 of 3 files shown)


boegel:
For now the script does little more than "parse" the build list and filter it (kick out installations that are only requested on GPU nodes):

```
$ python3 eessi.py eessi-build-list-2020.06.py
GROMACS
* 2020.1-foss-2020a-Python-3.8.2
OpenFOAM
* 7-foss-2019b
TensorFlow
* 2.2.0-foss-2019b-Python-3.7.4
```

(note how `2.2.0-fosscuda-2019b-Python-3.7.4` is not listed for `TensorFlow`)

boegel:
no longer relevant, see #21 for a working initial build script (with a better solution being worked at in https://github.com/easybuilders/easybuild-framework/pull/3479)

PR #11, opened 2020-09-02
add test script for OpenFOAM 8

boegel:
Single-node test script to verify that installation of `OpenFOAM/8-foss-2020a` in `2020.08` version of pilot repo is working.

A successful run should end with "`Finalising parallel run`".

The script is hardcoded to use only 2 cores, the run should complete in about 10 minutes when using 2 Intel Haswell cores.

It can be changed relatively easily to use `C` cores, by tweaking the values for `X`, `Y` and `Z`, making sure their product is equal to `C`, and that `Y=Z` and `X > Y`.

Examples:

* 2 cores: X=2, Y=1, Z=1
* 16 cores: X=4, Y=2, Z=2
* 20 cores: X=5, Y=2, Z=2
* 36 cores: X=9, Y=2, Z=2

### DIFF
--- a/tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
+++ b/tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
@@ -2,6 +2,10 @@
 
 module load OpenFOAM/8-foss-2020a
 
+# unset $LD_LIBRARY_PATH to avoid breaking system tools (like 'sed')
+# for the software in the EESSI software stack $LD_LIBRARY_PATH is irrelevant (because of RPATH linking)
+unset LD_LIBRARY_PATH
+
 which ssh &> /dev/null
 if [ $? -ne 0 ]; then
     # if ssh is not available, set plm_rsh_agent to empty value to avoid OpenMPI failing over it
--- a/tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
+++ b/tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+
+module load OpenFOAM/8-foss-2020a
+
+which ssh
+if [ $? -ne 0 ]; then
+    # if ssh is not available, set plm_rsh_agent to empty value to avoid OpenMPI failing over it
+    # that's OK, because this is a single-node run
+    export OMPI_MCA_plm_rsh_agent=''
+fi
+
+source $FOAM_BASH
+
+if [ -z $EBROOTOPENFOAM ]; then
+    echo "ERROR: OpenFOAM module not loaded?" >&2
+    exit 1
+fi
+
+export WORKDIR=/tmp/$USER/$$
+echo "WORKDIR: $WORKDIR"
+mkdir -p $WORKDIR
+cd $WORKDIR
+pwd
+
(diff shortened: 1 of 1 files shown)


tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
```diff
@@ -0,0 +1,93 @@
...
+BLOCKMESH_DIMENSIONS="100 40 40"
+# motorBike, 150M cells
+#BLOCKMESH_DIMENSIONS="200 80 80"
+
+# X*Y*Z should be equal to number of available cores
+X=2
+Y=1
+Z=1
+# total number of cores
+NP=2
```
ocaisa, line 35: What about `NP=$( expr $X * $Y * $Z )`

tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
```diff
@@ -0,0 +1,93 @@
...
+# cores per node
+PPN=2
```
ocaisa, line 37: `PPN=$NP`?

ocaisa (changes requested):

tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
boegel: In the long run, no, because `$NP` is the total number of cores, while `$PPN` is the number of cores per node.

Eventually we want to run this multi-node too.

ocaisa: Ok, then I would define `NODES` first and use `$(( NP / $NODES ))` (after checking `$(( NP % NODES ))`)

boegel: Or just `NP=$(($NODES * $PPN))`?

boegel: Ah, no, because X*Y*Z determine *total* number of cores. Ugh.

ocaisa: Ok, this would work without the need for rigorous testing (see https://stackoverflow.com/a/2395294) since it always rounds up:
```
PPN = $(( (NP + NODES -1)/NODES ))
```

tests/OpenFOAM/OpenFOAM-8_motorBike_2M-cells_2cores.sh
```diff
@@ -0,0 +1,102 @@
...
+Z=1
+# number of nodes
+NODES=1
+# total number of cores
+NP=$(($X * $Y * $Z))
+# cores per node
+if [ $NODES -eq 1 ]; then
+    PPN=$NP
+else
+    echo "Don't know how to determine \$PPN value if \$NODES > 1" >&2
```
ocaisa, line 46: `PPN = $(( (NP + NODES -1)/NODES ))`

ocaisa: No need for the if/else in that scenario

ocaisa (approved):

Issue #17, opened 2020-09-19
broken binutils installation in 2020.08 pilot repo


boegel:
The `binutils/2.34-GCCcore-9.3.0` installation in the 2020.08 version of the pilot repo was is not built correctly. It was configured with `--with-sysroot=$EPREFIX` which causes problems like this:

```
/cvmfs/pilot.eessi-hpc.org/2020.08/software/x86_64/intel/haswell/software/binutils/2.34-GCCcore-9.3.0/bin/ld.gold: error: 
cannot open /cvmfs/pilot.eessi-hpc.org/2020.08/compat/x86_64/cvmfs/pilot.eessi-hpc.org/2020.08/compat/x86_64/lib64/libm.so.6: No such file or directory
```

Note the "double tapping" of `$EPREFIX` (`/cvmfs/.../compat/x86_64`).

The `binutils` easyblock included with EasyBuild v4.3.0 does the correct thing (always using `--with-sysroot=/` as configure option), see https://github.com/easybuilders/easybuild-easyblocks/pull/2147 .

Fixing this problem is as simple as rebuilding `binutils/2.34-GCCcore-9.3.0` with EasyBuild v4.3.0 (`module load EasyBuild/4.3.0`, included in the 2020.08 pilot repo).

(hat tip to @terjekv for reporting this)

boegel:
fixed in 2020.09 pilot stack

PR #21, opened 2020-09-25
add script for installing 2020.09 pilot stack

### LINKED ISSUE #13: Filter LD_LIBRARY_PATH
We should set:
```
filter-env-vars=LD_LIBRARY_PATH
```
in our Easybuild configuration, to prevent (system) tools from picking up libraries from modules, which will result in nasty errors.

### LINKED ISSUE #14: Open MPI configuration
The Open MPI installation included in the 2020.08 pilot software stack is configured with `--without-verbs`, so it has no support for high-speed interconnects.

boegel:
fixes #13, #17

WIP because I'm still hoping to fix #14 too...

To use: copy both `EESSI-pilot-2020.09.sh eessi_software_subdir.py` in the same directory, and call `EESSI-pilot-2020.09.sh` after setting up the Gentoo Prefix environment with `startprefix`

c @terjekv, @bedroge


EESSI-pilot-2020.09.sh
```diff
@@ -0,0 +1,162 @@
...
+
+export EASYBUILD_SYSROOT=${EPREFIX}
+
+export EASYBUILD_DEBUG=1
+export EASYBUILD_TRACE=1
+
+export EASYBUILD_RPATH=1
+export EASYBUILD_FILTER_ENV_VARS=LD_LIBRARY_PATH
+
+export EASYBUILD_FILTER_DEPS=Autoconf,Automake,Autotools,binutils,bzip2,gettext,libreadline,libtool,M4,ncurses,XZ,zlib
```
boegel, line 76: More stuff we should consider adding to `--filter-deps`:

* `Bison`
* `cURL`
* `expat`
* `flex`
* `help2man`
* `pkg-config`

bedroge:
Nice, this is very useful! I'm running/testing it on a broadwell node now.

EESSI-pilot-2020.09.sh
bedroge: Do we still need to keep/discuss this somewhere (in an issue)?

bedroge (approved):
Lgtm, and this worked very well for building the pilot software for different architectures. I only had a minor issue with some hardcoded paths to /tmp (ran out of space).

EESSI-pilot-2020.09.sh
boegel: Yes, we should have a discussion on what we (won't) include in this list for the next iteration (2020.10).

We've hit several issues by leveraging the `binutils`, `zlib`, `ncurses` provided by Prefix, for different reasons.
That's not a reason to not filter those, but it does show that relying on stuff from the compat layer could have an impact (changes required in easyblocks/easyconfigs to make it work).

boegel: see #23

PR #22, opened 2020-09-27
add init script for 2020.09 version of pilot repo

boegel:
Ideally we also test the init script itself, by mounting the EESSI pilot repo in a GitHub Actions workflow, and sourcing it...

@bedroge The `init/bash-pilot-2020.09` script should be copied to `/cvmfs/pilot.eessi-hpc.org/2020.09/init/bash`, the `init/eessi_software_subdir_for_host.py` script should be copied into `/cvmfs/pilot.eessi-hpc.org/2020.09/init/`.

### DIFF
--- a/.github/workflows/tests_init.yml
+++ b/.github/workflows/tests_init.yml
@@ -19,7 +19,11 @@ jobs:
 
     - name: install Python packages
       run: |
-        pip install archspec
+        pip install archspec pytest
+
+    - name: unit tests for eessi_software_subdir_for_host.py script
+      run:
+          pytest -v init/test.py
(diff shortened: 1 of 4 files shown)


boegel:
Tests for the Python script pass (see https://github.com/boegel/software-layer/actions/runs/276338176), but are not run here yet because the workflow file is not in the repo...

boegel:
@bedroge This is good to go too imho?

bedroge (approved):
It's too good to go ;-)

bedroge:
Lgtm, and seems to work well.

Issue #29, opened 2020-10-22
Paraview font issues


bedroge:
@JvD007 pointed me at an issue with Paraview, which has some weird font issue when running it inside a Singularity container: all text is not visible, e.g. on buttons and menu bars.

I've been looking into this today, and it seems that our fontconfig module is configured to take fonts from `/usr/share/fonts`, see `$EBROOTFONTCONFIG/etc/fonts/fonts.conf`:
```
<!-- Font directory list -->

	<dir>/usr/share/fonts</dir>
```

This works fine on a normal machine, but our default EESSI pilot Singularity container does not have any fonts installed, and therefore doesn't have this directory.

A quick workaround is to bind mount the host's fonts folder in the container, which solved the issue for me. We could also add some font installation commands to our container recipes. But I don't really like those solutions, as other people who want to build customized containers will probably run into similar issues again.

So, alternatively, we could patch our fontconfig installation, and make it look for fonts in the compatibility layer. We will need to figure out what fonts we need to install then. Is this a acceptable and feasible solution?

bedroge:
fontconfig's configure also accepts some options for setting the default/additional font directory/directories:

```
  --with-default-fonts=DIR
                          Use fonts from DIR when config is busted
  --with-add-fonts=DIR1,DIR2,...
                          Find additional fonts in DIR1,DIR2,...
```

bedroge:
In order to verify what `--with-add-fonts` exactly does, I just did a local installation with that option set to some custom directory. It did what I more or less expected: it adds another `<dir>....</dir>` line to `fonts.conf`.

So I then used an overlayfs to change our current fontconfig installation, and added:
```
        <dir>/cvmfs/pilot.eessi-hpc.org/2020.09/compat/x86_64/usr/share/fonts</dir>
```

I wasn't sure which fonts I had to install, so started with `media-fonts/font-xfree86-type1`. This resulted in ParaView showing weird symbols instead of no letters. Then installed `media-fonts/font-adobe-100dpi`, which did show letters, but it didn't look very nice (unsharp/aliased). As I found something about `dejavu` earlier this afternoon (https://www.cfd-online.com/Forums/openfoam-installation/173514-paraview-4-4-font-problem.html), I finally installed `media-fonts/dejavu`, and now it works/looks perfectly.

bedroge:
Maybe this helps:
https://wiki.gentoo.org/wiki/Fontconfig#Picking_fonts

bedroge:
[...]

PR #31, opened 2020-10-28
Make fontconfig aware of the compatibility layer's fonts directory

bedroge:
See #29. This adds `--with-add-fonts=$EPREFIX/usr/share/fonts` to the `configopts`, to make sure that fontconfig will also look for fonts in the compatibility layer's fonts directory.

@boegel: does `eb` automatically pick up this file when Qt5 gets installed with `--robot`? Or should I add an explicit step that installs fontconfig using this local easyconfig file?

### DIFF
--- a/fontconfig-2.13.92-GCCcore-9.3.0.eb
+++ b/fontconfig-2.13.92-GCCcore-9.3.0.eb
@@ -1,0 +1,39 @@
+easyblock = 'ConfigureMake'
+
+name = 'fontconfig'
+version = '2.13.92'
+
+homepage = 'https://www.freedesktop.org/wiki/Software/fontconfig/'
+
+description = """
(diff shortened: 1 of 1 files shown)


boegel:
> @boegel: does `eb` automatically pick up this file when Qt5 gets installed with `--robot`? Or should I add an explicit step that installs fontconfig using this local easyconfig file?

`eb` only picks up files that are in the robot search path, so something like `eb --robot /path/where/this/easyconfig/is/located` would work.

But the best approach is probably to add a separate explicit `eb /path/to/fontconfig...eb` in the build script, with a comment explaining why we do it like that.

I'm planning to update the build script for EasyBuild v4.3.1 (just released) very soon, so I can tackle that too while I'm at it.

PR #32, opened 2020-10-29
update build script to use EasyBuild v4.3.1 by cleaning up now merged PRs

### LINKED ISSUE #30: Update software install script for 2020.10 pilot
- Add ReFrame
- Update EasyBuild version (when released)
- Add TensorFlow?

boegel:
~~still WIP, needs to be tested (and I'm not 100% confident TensorFlow will work since that hasn't been tested in a Prefix build environment yet)~~

fixes #30

### DIFF
--- a/EESSI-pilot-install-software.sh
+++ b/EESSI-pilot-install-software.sh
@@ -160,6 +160,18 @@ else
     error "EasyBuild not working?!"
 fi
 
+# patch RPATH wrapper to also take into account $LIBRARY_PATH (required for TensorFlow)
+# see https://github.com/easybuilders/easybuild-framework/pull/3495
+echo ">> Patching rpath_args.py script in EasyBuild installation..."
+EB_SCRIPTS=$EBROOTEASYBUILD/easybuild/scripts
+RPATH_ARGS='rpath_args.py'
+cp -a ${EB_SCRIPTS}/${RPATH_ARGS} ${EB_SCRIPTS}/${RPATH_ARGS}.orig
+cd ${TMPDIR}
+curl --silent -OL https://raw.githubusercontent.com/easybuilders/easybuild-framework/develop/easybuild/scripts/${RPATH_ARGS}
+cd - > /dev/null
+cp ${TMPDIR}/${RPATH_ARGS} ${EB_SCRIPTS}/${RPATH_ARGS}
+chmod u+x ${EB_SCRIPTS}/${RPATH_ARGS}
+
 echo_green "All set, let's start installing some software in ${EASYBUILD_INSTALLPATH}..."
 
 # install GCC, using GCC easyblock with workaround for bug introduced in EasyBuild v4.3.1,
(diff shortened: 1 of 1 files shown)


EESSI-pilot-install-software.sh
```diff
@@ -142,94 +142,65 @@ else
...
-#        see https://github.com/EESSI/software-layer/issues/14
-echo ">> Installing properly configured OpenMPI..."
-$EB --from-pr 11387 OpenMPI-4.0.3-GCC-9.3.0.eb --include-easyblocks-from-pr 2188 --robot
+echo ">> Installing OpenMPI..."
+$EB OpenMPI-4.0.3-GCC-9.3.0.eb --robot
 if [[ $? -eq 0 ]]; then
     echo_green "OpenMPI installed, w00!"
 else
     error "Installation of OpenMPI failed, that's not good..."
 fi
```
bedroge, line 197:
```suggestion

# FIXME: fix this in Easybuild by making a template value for sysroot available to easyconfigs and using that in all fontconfig easyconfigs
# for now, we use a custom easyconfig that sets ${EPREFIX}/usr/share/fonts as an additional fonts directory in configopts
echo ">> Installing fontconfig..."
$EB ./fontconfig-2.13.92-GCCcore-9.3.0.eb --robot
if [[ $? -eq 0 ]]; then
    echo_green "Done with fontconfig!"
else
    error "Installation of fontconfig failed!"
fi

```

Use something like this for fontconfig?

boegel: I wonder if we can do better, since this way you need to be aware that you need to copy both the script and the custom easyconfig(s?)...

Should we try downloading it from the repo on the fly in the script?

bedroge: Yeah, maybe that's better.

EESSI-pilot-install-software.sh
```diff
@@ -142,94 +142,79 @@ else
...
-echo ">> Installing Qt5 with extra patch to use zlib provided by Gentoo..."
-$EB --from-pr 11385 --robot
+# install custom fontconfig that is aware of the compatibility layer's fonts directory
+# see https://github.com/EESSI/software-layer/pull/31
+export FONTCONFIG_EC="fontconfig-2.13.92-GCCcore-9.3.0.eb"
+echo ">> Installing custom fontconfig easyconfig (${FONTCONFIG_EC})..."
+cd ${TMPDIR}
+curl -OL https://raw.githubusercontent.com/EESSI/software-layer/master/easyconfigs/${FONTCONFIG_EC}
+cd - > /dev/null
+$EB ${FONTCONFIG_EC} --robot
```
bedroge, line 183: Shouldn't these be in reverse order?

boegel: Ugh, yes, I actually meant to use `$TMPDIR/${FONTCONFIG_EC}` here... Will fix.

boegel:
Fully tested on both `x86_64` (`amd/zen2` with `intel/haswell`, `intel/skylake_avx512` and `x86_64/generic` in the works) and `aarch64` (fully working on `aarch64/graviton2`), so good to go imho...

bedroge (approved):

PR #34, opened 2020-11-04
update init script for 2020.10

boegel:
@bedroge Both of these belong in `/cvmfs/pilot.eessi-hpc.org/2020.10/init/`


bedroge (approved):

bedroge:
I've just published these new files, they should be visible in a few minutes.

PR #43, opened 2020-11-24
Add script to wrap pilot Python3 installation

ocaisa:
This is a straightforward (and hardware portable) way to be able to give Magic Castle the Python it needs to properly configure the JupyterHub service.

### DIFF
--- a/init/eessi_pilot_python3
+++ b/init/eessi_pilot_python3
@@ -1,0 +1,37 @@
+#!/usr/bin/env sh
+EESSI_PILOT_VERSION="2020.10"
+EESSI_PREFIX=/cvmfs/pilot.eessi-hpc.org/$EESSI_PILOT_VERSION
+
+if [ -d $EESSI_PREFIX ]; then
+  EPREFIX=$EESSI_PREFIX/compat/$(uname -m)
+  if [ -d $EPREFIX ]; then
+    # determine subdirectory in software layer
+    # note: eessi_software_subdir_for_host.py will pick up value from $EESSI_SOFTWARE_SUBDIR_OVERRIDE if it's defined!
+    EESSI_PYTHON=$EPREFIX/usr/bin/python3
+    EESSI_SOFTWARE_SUBDIR=$($EESSI_PYTHON ${EESSI_PREFIX}/init/eessi_software_subdir_for_host.py $EESSI_PREFIX)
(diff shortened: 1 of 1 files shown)


ocaisa:
@boegel What do you think about the location of this? I could put it in a MagicCastle subdir to hide it a bit more (as I wouldn't expect anyone to be using it)?

ocaisa:
It does work well for Magic Castle, JupyterHub sees through the wrapper and the actual (arch-specific) python path is picked up.

boegel:
If you just need Python without bells and whistles, why not use the `python3` from the compatibility layer instead?

Also, this script copy-pastes a lot from the `init/bash` script, we should factor that out a bit and have a common script that is run or sourced by the others?

ocaisa:
@boegel I tried that but it's not good enough. JupyterHub has a nice plugin that allows you will to load modules  to change the kernel computing environment, this only works if we use the EESSI python (since it uses a `site.cfg` setup rather than PYTHONPATH)

boegel:
OK, then we should clarify that with a comment in the script why we use this Python specifically...

ocaisa:
This now requires #47

EESSI-pilot-install-software.sh
```diff
@@ -269,6 +269,21 @@ else
...
+echo ">> Installing ReFrame 3.2..."
+$EB ReFrame-3.2.eb --robot
+if [[ $? -eq 0 ]]; then
+    echo_green "ReFrame installed, enjoy!"
+else
+    error "Installation of ReFrame failed, that's a bit strange..."
+fi
+
+echo ">> Installing RStudio-Server 1.3.1093..."
+$EB --from-pr 11764 RStudio-Server-1.3.1093-foss-2020a-Java-11-R-4.0.0.eb --robot
```
ocaisa, line 281: @bedroge RStudio-Server has an OS dep:
```
osdependencies = [
    ('pam-devel', 'libpam0g-dev')
]
``` 
Is that already available in the prefix layer?

boegel:
@ocaisa This needs a sync with current `master` now that #47 is merged?

ocaisa:
@boegel I've removed the changes to the build script since they are a moving target, whereas the rest of the PR is more general.

I'll open an updated PR for the build script changes.

boegel (approved):

PR #47, opened 2020-11-26
Split setting of environment variables out from bash init script

ocaisa:


init/bash
```diff
@@ -1,57 +1,33 @@
...
-          echo "*** Known problems in the ${EESSI_PILOT_VERSION} pilot software stack ***"
-          echo
-          echo "1) No Lmod cache has been generated yet for the module files included"
-          echo "   in the ${EESSI_PILOT_VERSION} pilot software stack."
-          echo "   As a result, 'module' commands may be somewhat slow."
-          echo "   (see also https://github.com/EESSI/software-layer/issues/10)"
-          echo
-          echo
+# The following method should be safe, but might break if file is a symlink
+# (could switch to $(dirname "$(readlink -f "$BASH_SOURCE")") in that case)
```
terjekv, line 10: `readlink -f` does not work on MacOS, so leaving that be might be just as well.

terjekv:
This looks pretty sane to me.  Should we aim to merge and test it until it's solid? Staying clear of `readlink -f ` due to the lack of MacOS support is a good idea.

init/eessi_environment_variables
```diff
@@ -0,0 +1,67 @@
...
+          # Allow for the use of a custom MNS
+          if [ -z ${EESSI_CUSTOM_MODULEPATH+x} ]; then
+            # EESSI_CUSTOM_MODULEPATH not set so we use our defaults
+
+            # Allow for use of alternative module tree shipped with EESSI
+            if [ -z ${EESSI_MODULE_SUBDIR+x} ]; then
+              # EESSI_MODULE_SUBDIR not set
+              EESSI_MODULE_SUBDIR="modules/all"
+            fi
+            EESSI_MODULEPATH=$EESSI_SOFTWARE_PATH/$EESSI_MODULE_SUBDIR
```
ocaisa, line 38: This is a way to support not just alternative MNS, but also domain-specific module trees (I can imagine combining this with domain easystack files)

init/eessi_environment_variables
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+export EESSI_PILOT_VERSION="2020.12"
+export EESSI_PREFIX=/cvmfs/pilot.eessi-hpc.org/$EESSI_PILOT_VERSION
+
+if [ -d $EESSI_PREFIX ]; then
+  echo "Found EESSI pilot repo @ $EESSI_PREFIX!" >> $output
+
+  # Set EPREFIX since that is basically a standard in Gentoo Prefix
+  export EPREFIX=$EESSI_PREFIX/compat/$(uname -m)
```
boegel, line 16: @ocaisa This needs to be tweaked because there's an extra subdirectory `linux/` now, see https://github.com/ocaisa/software-layer/pull/1

boegel (changes requested):

boegel (approved):

boegel:
@bedroge Please "deploy":

```
$ ls /cvmfs/pilot.eessi-hpc.org/2020.12/init/
bash  eessi_environment_variables  eessi_software_subdir_for_host.py
```

PR #50, opened 2020-12-13
update init script for 2020.12 pilot

boegel:


boegel:
@bedroge Please install this at `/cvmfs/pilot.eessi-hpc.org/2020.12/init/bash`?

ocaisa:
Can we update and use #47 instead?

boegel:
Yes, definitely, my bad. Closing this one...

Issue #52, opened 2020-12-16
Use all binaries from the compatibility layer - also at runtime


casparvl:
Currently, when we build in the software layer, we do that on top of the compatibility layer. That means that e.g. using a "system" python would mean: using the python from the compatibility layer.

At runtime, the same module would however use another "system" python: the one from the host. This could present problems...

We could see what happens if we use everything from the compat layer, and learn if that gives problems. Proposal is to do that for `2020.12`.

boegel:
In #54 I'm adding `$EPREFIX/usr/bin` to `$PATH` in the init script.

I' not 100% convinced this is the right way forward, I think we want to have control over which commands get inherited from the compat layer vs which ones come from the host, but for now #54 makes sense (so we can see if we hit any problems with that approach).

Long term, I think only leveraging specific commands from the compat layer (by creating symlinks to `$EPREFIX/usr/bin`) may be better?

boegel:
fixed in #54

Issue #46, opened 2020-11-26
certificate issues with curl provided through EasyBuild


boegel:
I ran into problems with the `cURL` we install as a module with EasyBuild. It remembers the path to the certificates it found in the host OS, and when you then try using it in a different OS it fails.

For example (I hit this when installing Bioconductor):

```
Warning: download failed
  web resource path: https://experimenthub.bioconductor.org/metadata/experimenthub.sqlite3
  local file path: /tmp/eb-074tby_7/RtmpWxUxrk/BiocFileCache/300674bca9b85_experimenthub.sqlite3
  reason: error setting certificate verify locations:
  CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
```

This was with a `cURL` installed on CentOS 8, which was then being used in Debian 10.
Reinstalling the `cURL` module in Debian 10 fixed the issue (but then it's probably broken in CentOS).

I think our options here are:

* Just using the `curl` provided in the compatibility layer (so adding it to `--filter-deps` in the EasyBuild configuration).
  That would likely involve making sure it's *always* there, not just when building software (in a `startprefix` env), so maybe we need to create a small symlink farm behind an `eessi-env` module that adds specific tools to `$PATH` (like `curl`, but also like `ld` & co, see #26).

* We configure our own `cURL` installation with EasyBuild correctly to use the certificates from the compatibility layer.

boegel:
This fixed, for now, because i) we include `cURL` in `--filter-deps` (see #51) and ii) we use the `curl` command provided by the compat layer also at runtime (see #54).

PR #54, opened 2021-01-07
add location of commands provided by compat layer to $PATH

### LINKED ISSUE #26: gcc/ld errors when compiling with EESSI modules
I was trying to compile some C code, and ran into the following issues:
```
$ gcc hello.c -o hello
/usr/bin/ld.gold: --push-state: unknown option
/usr/bin/ld.gold: use the --help option for usage information
collect2: error: ld returned 1 exit status

$ which -a ld.gold
/usr/bin/ld.gold
/bin/ld.gold
```
We somehow need to instruct gcc to use the one from the compatibility layer.

boegel:
fixes #26 and #52

Also relevant for #46 (as long as we only use `cURL` from compat layer, that problem shouldn't occur); in the `2020.12` revision we added `cURL` to `--filter-deps` (see #51)

### DIFF
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -23,6 +23,11 @@ if [ -d $EESSI_PREFIX ]; then
 
   # Set EPREFIX since that is basically a standard in Gentoo Prefix
   export EPREFIX=$EESSI_PREFIX/compat/$EESSI_OS_TYPE/$EESSI_CPU_FAMILY
+
+  # add location of commands provided by compat layer to $PATH;
+  # see https://github.com/EESSI/software-layer/issues/52
+  export PATH=$EPREFIX/usr/bin:$PATH
+
   export EESSI_EPREFIX=$EPREFIX
   if [ -d $EESSI_EPREFIX ]; then
 


init/eessi_environment_variables
```diff
@@ -23,6 +23,11 @@ if [ -d $EESSI_PREFIX ]; then
 
   # Set EPREFIX since that is basically a standard in Gentoo Prefix
   export EPREFIX=$EESSI_PREFIX/compat/$EESSI_OS_TYPE/$EESSI_CPU_FAMILY
+
+  # add location of commands provided by compat layer to $PATH;
+  # see https://github.com/EESSI/software-layer/issues/52
+  export PATH=$EPREFIX/usr/bin:$PATH
```
bedroge, line 29: Just to verify: do we only need `$EPREFIX/usr/bin`? The `startprefix` script also adds `/(s)bin` and `/usr/sbin` (and  `/usr/local/(s)bin` and `/opt/bin`, which we don't have in our installation).

boegel: I would start with only `$EPREFIX/usr/bin`, since that's where all the tools are that we've been having trouble with (`ld`, `curl`, `python`).

Long-term, I'm not convinced this is the best way forward, but as suggested in #52 it makes sense to try this approach and see if we hit any trouble with it.

bedroge: Okay!

bedroge (approved):

PR #61, opened 2021-01-29
Allow Lmod initialisation to also be silent

ocaisa:
Currently we get a line on every login
```
Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/2020.12/software/x86_64/amd/zen2/.lmod/lmodrc.lua
```
which I think comes from here


boegel (approved):
lgtm

PR #62, opened 2021-01-29
Reload modules when opening a subshell

ocaisa:
When a new shell is opened, if it is started from an initial shell there will be some modules already loaded, we need to do a `reload` to ensure that these modules take priority. A `refresh` means that (in the current configuration) the prefix `bin` get's priority in JupyterHub.

Not sure if I should instead be checking if Lmod is already initialized (and do nothing in that case)


ocaisa:
@bedroge Can we add this for `2021.03`?

bedroge:
Sure, I'll add them to the repo.

bedroge (approved):

PR #65, opened 2021-02-04
Gromacs v2

casparvl:
Created a GROMACS reframe test that is as generic as we can make it right now. It consists of 2*3 tests:

- Running GROMACS in the CVMFS container (i.e. with singularity exec). Use the tag `-t container` to start this.
- Running GROMACS on a system where CVMFS is mounted natively. Use the tag `-t native` to start this.

Additionally, each of those two options has 3 test 'sizes':
- Single node, 10k simulation steps
- Four nodes, 40k simulation steps
- Ten nodes, 100k simulation steps

To run e.g. the container test, on a single node, the call would be:

`reframe --config-file=config/settings.py --checkpath eessi-checks/applications/ -r -v -t container -t single`

Note that for now the following still has to be adjusted for each local site:

- The `.../reframe/config/settings.py` (see the ReFrame documentation). You'll need to specify at least an environment `container` to run the container-based tests. This environment should provide the singularity command, and either an MPI that can be used to launch mpi processes with `mpirun`, or select `srun` as launcher if your SLURM supports `PMIx`.
- `SHAREDSPACE` and `LOCALSPACE` in `shared_alien_cache_minimal.sh` should point to a cache on a shared disk (available from all nodes) and on a local disk.
- `self.num_tasks_per_node` in `gromacs.py`

You can run with e.g.
```
reframe --config-file=config/settings.py --checkpath eessi-checks/applications/ -r -v -t container -t single
reframe --config-file=config/settings.py --checkpath eessi-checks/applications/ -r -v -t native -t single
```


casparvl:
Got superseeded by https://github.com/EESSI/software-layer/pull/115/files , which is essentially the same test case, but in the 'library of tests' format of `ReFrame`. I've taken out the container-based test in that newer version, since I don't think we'll really keep supporting those (too complicated). If we do, we can always come back to this PR and cherry-pick the parts we need.

PR #66, opened 2021-02-05
Make sure that /dev/stdout is writable if needed

ocaisa:
There are corner cases (like when `su -` into an account) that `/dev/stdout` is not writable, rather than get permission denied errors we can instead silently make things silent


ocaisa:
I'm not 100% sure this is a good idea

boegel:
@ocaisa How should we move forward with this?

boegel:
Alternate approach implemented in #415

PR #76, opened 2021-03-02
Additional packages for next pilot stage: code-server (not on ppce64le), ReFrame, RStudio-Server

ocaisa:
Add additional packages for next pilot stage:

- 2 packages for Magic Castle + EESSI related to JupyterHub (code-server for VS Code in browser and RStudio Server)
- ReFrame for testing


boegel:
@ocaisa Have you tested the installation of these on top of the existing EESSI software stack?
There may be issues there that we need to sort out first (some of which may be specific to `aarch64` or `ppc64le`)...

ocaisa:
I'm testing them on AMD now. Is the AWS setup ready for wider use yet?

boegel:
@ocaisa Yes, we can give you access to resources in AWS for testing things, let's sync on that via the EESSI Slack.

ocaisa:
This was tested on `x86_64` (SkyLake) and `aarch64`(Graviton). @boegel also tested on `ppcle64` so I think this should be good to go

boegel:
Tested on `power8` with test install of 2021.02 software layer, works fine, so I expect no problems.

boegel (approved):

PR #75, opened 2021-02-27
update software installation script for EESSI 2021.02 pilot

boegel:
Overview of changes:

* updates for EasyBuild v4.3.3 (fixes #70)
* also use `{linux,macos}` subdirectory in `2021.02/software/` (fixes #71)
* change check for Prefix environment (relying on start of `$PATH` value is problematic, sometimes additional paths are included at front)
* fix path to Lmod init script (changes due to https://github.com/EESSI/gentoo-overlay/pull/9?)
* filter additional dependencies (to use what's already provided in compat layer)
  * *not* filtering `Bison`, due to problems with `Qt5` using `yacc` from compat layer (see #49)
* stop using EasyBuild bootstrap script, use `pip install` + `eb --install-latest-eb-release`

(WIP, still testing it...)

PR #78, opened 2021-03-24
trivial update of software layer install script for 2021.03 pilot version

boegel:
Still testing on top of the freshly built `2021.03` compat layer, but should be fine after cleanup that was done in #75 and #77...


ocaisa (approved):

boegel:
Fully tested on AMD Rome now, so good to merge...

Issue #79, opened 2021-03-30
failing 'rstan' installation due to timezone issue?!


boegel:
During the installation of `R-4.0.0-foss-2020a.eb` in the `2021.03` version of the software layer, the installation of `rstan` is failing with:

```
** R
** inst
** byte-compile and prepare package for lazy loading
Error in as.POSIXlt.character(x, tz, ...) :
  character string is not in a standard unambiguous format
Error: unable to load R code in package rstan
Execution halted
ERROR: lazy loading failed for package rstan
```

Others have reported this issue with `rstan` too, see https://github.com/stan-dev/rstan/issues/612.

The problem is easy to reproduce manually when starting `R`:

```
$ R
> as.POSIXct("1970-01-01 00:00.00 UTC")
Error in as.POSIXlt.character(x, tz, ...) :
  character string is not in a standard unambiguous format
```

Workaround in R:

```
> as.POSIXct("1970-01-01 00:00.00 UTC", tz='UTC')
[1] "1970-01-01 UTC"
```

Workaround using `$TZ`:

```
$ TZ=UTC R
> as.POSIXct("1970-01-01 00:00.00 UTC")
[1] "1970-01-01 UTC"
```

I worked around this in 86e5ca9 by setting `$TZ` to `UTC` when installing `R`.

boegel:
This is related to not having `/etc/localtime` anymore in the `2021.03` compat layer, while it is there in `2020.12`:

```
$ cat /cvmfs/pilot.eessi-hpc.org/2020.12/compat/linux/x86_64/etc/localtime
TZif2-00TZif2-00
<-00>0
```

The problem can be fixed by symlinking the missing file to the host:

```
ln -s /etc/localtime /cvmfs/pilot.eessi-hpc.org/2021.03/compat/linux/x86_64/etc/localtime
```

bedroge:
This has been solved in https://github.com/EESSI/compatibility-layer/pull/105.

PR #84, opened 2021-04-07
Update RStudio-Server installation

ocaisa:
It was not working due to permission related issues

### DIFF
--- a/EESSI-pilot-install-software.sh
+++ b/EESSI-pilot-install-software.sh
@@ -313,7 +313,7 @@ check_exit_code $? "${ok_msg}" "${fail_msg}"
 echo ">> Installing RStudio-Server 1.3.1093..."
 ok_msg="RStudio-Server installed, enjoy!"
 fail_msg="Installation of RStudio-Server failed, might be OS deps..."
-$EB RStudio-Server-1.3.1093-foss-2020a-Java-11-R-4.0.0.eb --robot
+TZ=UTC $EB --from-pr 12544 RStudio-Server-1.3.1093-foss-2020a-Java-11-R-4.0.0.eb --robot
 check_exit_code $? "${ok_msg}" "${fail_msg}"
 
 echo ">> Installing OSU-Micro-Benchmarks 5.6.3..."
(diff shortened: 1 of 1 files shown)


EESSI-pilot-install-software.sh
```diff
@@ -313,7 +313,7 @@ check_exit_code $? "${ok_msg}" "${fail_msg}"
 echo ">> Installing RStudio-Server 1.3.1093..."
 ok_msg="RStudio-Server installed, enjoy!"
 fail_msg="Installation of RStudio-Server failed, might be OS deps..."
-$EB RStudio-Server-1.3.1093-foss-2020a-Java-11-R-4.0.0.eb --robot
+TZ=UTC $EB --from-pr 12544 RStudio-Server-1.3.1093-foss-2020a-Java-11-R-4.0.0.eb --robot
```
ocaisa, line 316: The `TZ` thing is just precautionary, I've no idea if it is really required

boegel: It's not required, only needed for very specific cases like rstan (see #79), so please drop it.

ocaisa: Removed, however I did find on Google that this can be an issue for RStudio

boegel: link?

ocaisa: Nope, can't find it now, it was mentioned on one of the links I followed related to se-linux and the error that this change solves:
```
rserver[7485]: ERROR system error 13 (Permission denied) [path: /var/run/rstudio-server, target-dir: ]; OCCURRED AT rstudio::core::Error rstudio::core::FilePath::createDirectory(const string&) const src/cpp/shared_core/FilePath.cpp:648; LOGGED FROM: int main(int, char* const*) src/cpp/server/ServerMain.cpp:555
```
something about a timeout of something due to it  not being executable because of se-linux and setting `TZ` as being a way of getting around that

ocaisa: Found it https://github.com/rstudio/rstudio/issues/4937#issuecomment-643060730

ocaisa:
Screenshot 2021-04-08 13 38 12

ocaisa:
I tested this by (re)building on top of the EESSI stack, and it works on Magic Castle.

boegel (approved):

PR #100, opened 2021-04-26
Allow rebuilding the module tree via some envvars

### LINKED ISSUE #89: module avail always returns 0 for an existing module dir but non-existing module
While working on #86, I noticed that the `module avail <whatever>` command always returned 0 for me, even if a module for `<whatever>` does not exist:

```
[EESSI pilot 2021.03] $ module avail EasyBuild/4.5.0
No module(s) or extension(s) found!
Use "module spider" to find all possible modules and extensions.
Use "module keyword key1 key2 ..." to search for all possible modules matching
any of the "keys".


[EESSI pilot 2021.03] $ echo $?
0
[EESSI pilot 2021.03] $ module avail EasyBuild/4.3.4

 /cvmfs/pilot.eessi-hpc.org/2021.03/software/linux/x86_64/intel/haswell/modules/all 
   EasyBuild/4.3.4

Use "module spider" to find all possible modules and extensions.
Use "module keyword key1 key2 ..." to search for all possible modules matching
any of the "keys".


[EESSI pilot 2021.03] $ echo $?
0
[EESSI pilot 2021.03] $ module avail ThisDoesNotExist
No module(s) or extension(s) found!
Use "module spider" to find all possible modules and extensions.
Use "module keyword key1 key2 ..." to search for all possible modules matching
any of the "keys".


[EESSI pilot 2021.03] $ echo $?
0
```

When building a new stack from scratch, you don't notice this, because the `$MODULEPATH` then still points to a non-existing directory, in which case `module avail` does return 1:
```
[EESSI pilot 2021.03] $ export MODULEPATH=/does/not/exist
[EESSI pilot 2021.03] $ module avail EasyBuild
Lmod has detected the following error:  module avail is not
possible. MODULEPATH is not set or not set with valid paths.



[EESSI pilot 2021.03] $ echo $?
1
```

But if you need to update EasyBuild, as was necessary for #86, this is annoying.
A possible solution would be to use `module spider` or `module is-avail`.

ocaisa:
Fixes #89
Requires ~~https://github.com/easybuilders/easybuild-framework/pull/3659~~ to work


ocaisa:
Some example usages:
```
EESSI_REBUILD_PATH=/project/def-sponsor00/ocaisa/easybuild bash EESSI-pilot-install-software.sh
```
```
EESSI_REBUILD_PATH=/project/def-sponsor00/ocaisa/easybuild EESSI_REBUILD_MODULEPATH=$EESSI_REBUILD_PATH/modules/all/Core EESSI_REBUILD_MNS=HierarchicalMNS bash EESSI-pilot-install-software.sh
```

EESSI-pilot-install-software.sh
```diff
@@ -138,14 +138,29 @@ echo ">> Setting up \$MODULEPATH..."
...
+
 if [[ -z ${MODULEPATH} ]]; then
     error "Failed to set up \$MODULEPATH?!"
 else
     echo_green ">> MODULEPATH set up: ${MODULEPATH}"
 fi
 
-echo ">> Checking for EasyBuild module..."
+REQ_EB_VERSION='4.3.4'
+echo ">> Checking for EasyBuild module (required version is ${REQ_EB_VERSION})..."
```
ocaisa, line 163: To be honest since a specific version is required we should be trying to load that version, not the default. If we fail, we should get eb to install the latest release, and if that is too recent use the latest version to install the required version.

ocaisa: We can do that in this block with `module avail easybuild/${REQ_EB_VERSION}`, then using a temporary pip installation to _either_ give us the required version _or_ the latest release if that is not available.

bedroge: https://github.com/EESSI/software-layer/issues/89 this is probably also relevant here

ocaisa: @bedroge You are right, I will fix this and use `module load` instead

EESSI-pilot-install-software.sh
```diff
@@ -62,14 +62,14 @@ export EPREFIX=${EESSI_PREFIX}/compat/${EESSI_OS_TYPE}/${EESSI_CPU_FAMILY}
...
+EASYBUILD='EasyBuild'
 if [[ "$1" == "--generic" || "$EASYBUILD_OPTARCH" == "GENERIC" ]]; then
     echo_yellow ">> GENERIC build requested, taking appropriate measures!"
     DETECTION_PARAMETERS="$DETECTION_PARAMETERS --generic"
     GENERIC=1
     EB='eb --optarch=GENERIC'
 fi
 
-
 # make sure we're in Prefix environment by which path to 'bash' command
```
ocaisa, line 73: Is this check really good enough? My impression is that `$EPREFIX/startprefix` does more than just change your `$PATH`. To really check we should check what they do:
```
if [[ ${SHELL#/cvmfs/pilot.eessi-hpc.org/2021.03/compat/linux/x86_64} != ${SHELL} ]]
then
        echo "You appear to be in prefix already (SHELL=${SHELL})" > /dev/stderr
        exit -1
elif [[ ${SHELL#/cvmfs/pilot.eessi-hpc.org/2021.03/compat/linux/x86_64} != ${SHELL} ]] &&
         [[ ${EPREFIX-unset} == '/cvmfs/pilot.eessi-hpc.org/2021.03/compat/linux/x86_64' ]]
then
        echo "You appear to be in stacked prefix already (EPREFIX=${EPREFIX})" > /dev/stderr
        exit -1
fi
```
This also makes the shebang of this file wrong, it should be ` #!/usr/bin/env bash`

ocaisa:
This (also) covered what was merged in #134, before I go to the effort of resolving the (current) conflict is there a chance it will be merged anytime soon? @bedroge @boegel

EESSI-pilot-install-software.sh
```diff
@@ -107,22 +110,34 @@ echo ">> Setting up \$MODULEPATH..."
...
 fi
 
-REQ_EB_VERSION='4.4.1'
-
-echo ">> Checking for EasyBuild module..."
-ml_av_easybuild_out=$TMPDIR/ml_av_easybuild.out
-module avail 2>&1 | grep -i easybuild/${REQ_EB_VERSION} &> ${ml_av_easybuild_out}
+echo ">> Checking for ${EB_MODULE_NAME} module (required version is ${REQ_EB_VERSION})..."
+ml_show_easybuild_out=$TMPDIR/ml_show_easybuild.out
+module show ${EB_MODULE_NAME}/${REQ_EB_VERSION} &> ${ml_show_easybuild_out}
```
ocaisa, line 136: @boegel I see you used grep to get around https://github.com/EESSI/software-layer/issues/89, this approach also works, not sure which  is better...

ocaisa:
@bedroge @boegel Kicking the tyres on this again, I will soon have to update it again, it would be nice if it got included soon...

ocaisa:
I think I would like to change _where_ we install modules to actually use the naming scheme in the path, so `modules/EasyBuildMNS` rather than `modules` as the default. This will allow us to easily ship multiple schemes and have a simple setting for our initialisation script.

boegel:
@ocaisa Is this still relevant?

ocaisa:
@boegel Yes, still relevant but requires yet another update at this point.

ocaisa:
Closing this, can return to it at some point

PR #106, opened 2021-05-04
Tensorflow and Horovod tests

casparvl:
This ReFrame test contains TensorFlow and Horovod tests for both CPU and GPU. In EESSI, we don't have GPU-based modules yet, nor do we have Horovod, but the tests have been checked using a local module environment. Once we have GPUs/Horovod in EESSI, these test can be used right away. Until that time, tags can be used to select a subset:
- `-n TensorFlow2Native` to only select the TensorFlow tests (no Horovod)
- `-t cpu` to only select the cpu based tests
(note that you can combine both arguments)

In order to run the test, run e.g. (in the `tests/reframe` directory):

```
reframe --config-file=config/settings.py --checkpath eessi-checks/applications/tensorflow2.py -r --prgenv=foss --performance-report -n TensorFlow2Native -t cpu
```
where your `settings.py` at least specifies a partition that has 
- `num_cpus` defined
- `foss` as a valid programming environment
- Specify the foss environment in the config's `environments` dictionary

If your only valid programming environment in your `settings.py` is `foss`, you may ommit `--prgenv=foss` from the run command


casparvl:
This PR is now outdated and is superseded by https://github.com/EESSI/software-layer/pull/122 where I recreated the TensorFlow and Horovod tests with ReFrame's library of tests pattern

PR #111, opened 2021-05-21
add script to create tarball to ingest on build node

boegel:
First stab at a script that facilitates creating the tarball to be ingested on a build node after building/installing additional software.

Far from perfect, because:

* It assumes that the `fuse-overlayfs` upper directory is located at `/tmp/$USER/EESSI/overlay-upper`
* Only deals with adding of new files/directories (except for Lmod cache)
* Hard filters out the `EasyBuild` software installation, because `*.pyc` files are created when EasyBuild is running, resulting in things like:
```
$ find /tmp/ec2-user/EESSI/overlay-upper/2021.03/software/linux/aarch64/generic/software/EasyBuild/ -type f
/tmp/ec2-user/EESSI/overlay-upper/2021.03/software/linux/aarch64/generic/software/EasyBuild/4.3.4/lib/python3.8/site-packages/easybuild/easyblocks/e/__pycache__/elpa.cpython-38.opt-1.pyc
/tmp/ec2-user/EESSI/overlay-upper/2021.03/software/linux/aarch64/generic/software/EasyBuild/4.3.4/lib/python3.8/site-packages/easybuild/easyblocks/e/__pycache__/.wh.elpa.cpython-38.opt-1.pyc.281473148148112
/tmp/ec2-user/EESSI/overlay-upper/2021.03/software/linux/aarch64/generic/software/EasyBuild/4.3.4/lib/python3.8/site-packages/easybuild/easyblocks/q/__pycache__/quantumespresso.cpython-38.opt-1.pyc
/tmp/ec2-user/EESSI/overlay-upper/2021.03/software/linux/aarch64/generic/software/EasyBuild/4.3.4/lib/python3.8/site-packages/easybuild/easyblocks/q/__pycache__/.wh.quantumespresso.cpython-38.opt-1.pyc.281473142905232
```

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -e
+
+if [ $# -ne 2 ]; then
+    echo "ERROR: Usage: $0 <pilot version (example: 2021.03)> <CPU arch subdir (example: x86_64/amd/zen2)" >&2
+    exit 1
+fi
+pilot_version=$1
+cpu_arch_subdir=$2
+
+tmpdir=`mktemp -d`
+echo ">> tmpdir: $tmpdir"
+
+os="linux"
+cvmfs_repo="/cvmfs/pilot.eessi-hpc.org"
+
+software_dir="${cvmfs_repo}/${pilot_version}/software/${os}/${cpu_arch_subdir}"
+if [ ! -d ${software_dir} ]; then
+    echo "Software directory ${software_dir} does not exist?!" >&2
+    exit 2
+fi
+
+overlay_upper_dir="/tmp/$USER/EESSI/overlay-upper"
+
+software_dir_overlay="${overlay_upper_dir}/${pilot_version}/software/${os}/${cpu_arch_subdir}"
+if [ ! -d ${software_dir_overlay} ]; then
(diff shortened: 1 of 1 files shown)


create_tarball.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -e
+
```
terjekv, line 4: Do we want to do something like:

for pilot_version in $( ls -rt1d ${cvmfs_repo}  ); do
     for cpu_arch_subdir = $( ls -rt1d ${cvmfs_repo}/${pilot_version} ); do
         make_tarball( $pilot_version, $cpu_arch_dir )
    done
done

boegel: No, that would create a huge tarball for every CPU subdir, we only want to tar up the new stuff, to avoid having to deal with huge tarballs.

terjekv: I was thinking of implementing `make_tarball` the same way as the original code, using data from upper, but essentially building tarballs for every arch subdir for every pilot version the script finds. Also, we can now add a way to upload data to the staging bucket. I'll write some test code. :)

terjekv: https://github.com/EESSI/infrastructure/issues/25

PR #115, opened 2021-06-10
Gromacs ReFrame test for EESSI, with seperate library test

casparvl:
Before running ReFrame with the test in this PR
- make sure that `tests/reframe` is in your `PYTHONPATH`
- adjust the attached settings.py for your system (adapt partition names, number of CPU cores, number of GPUs etc)
- make sure you have a new enough ReFrame installation (with `3.5.0` the `required` keyword didn't work properly for me, `3.6.2` did)

This test should run on the EESSI software stack, but would also run on a locally installed GROMACS. Skipping of GPU tests on CPU partitions and vice versa only works as long as the module name contains `cuda` as substring whenever it is a GPU module (i.e. probably won't work in a hierarchical module system).


tests/reframe/testlib/applications/gromacs/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Derived tests must specify the variables ``num_tasks``, ``num_tasks_per_node``, ``num_cpus_per_task``, ``nsteps`` and ``modules``.
+    Note that a sufficiently large ``nsteps`` needs to be defined in order for GROMACS to pass the load balancing phase.
+    As a rough estimate: 10000 steps would generally be ok for 24 tasks, a 100000 steps for 240 tasks, etc.
+    '''
+
+    num_tasks = required
+    num_tasks_per_node = required
+    num_cpus_per_task = required
+    nsteps = required
```
jjotero, line 18: This variable needs to be declared first (`num_tasks`, `num_tasks_per_node` and so on were declared in the `rfm.RegressionTest` class already.)
```suggestion
    nsteps = variable(int)
```
I've created an issue (https://github.com/eth-cscs/reframe/issues/2011) to raise an error when this is attempted.

tests/reframe/eessi-checks/applications/gromacs.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @rfm.run_after('init')
+    def apply_module_info(self):
+        self.s, self.e, self.m = self.module_info
+        self.valid_systems = [self.s]
+        self.modules = [self.m]
+        self.valid_prog_environs = [self.e]
+
+    @rfm.run_after('init')
+    def set_test_scale(self):
```
jjotero, line 29: This comment is just a note on the style for the test parameterisation :)
If you're not too bothered about how the name of the test gets printed out, I would write the parameter `scale` as
```python
    scale = parameter([
        ('singlenode', 10000, 1),
        ('small', 40000, 4),
        ('large', 100000, 10),
    ])
```
so then this hook just becomes
```python
@run_after('init')
def set_test_scale(self):
    scale_variant, self.nsteps, self.num_nodes = self.scale
    self.tags.add(scale_variant)
```
In any case, the naming of the test with long parameters should be fixed in the future, since there is an open PR on that.

tests/reframe/eessi-checks/applications/gromacs.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @rfm.run_after('init')
+    def requires_gpu(self):
+        self.requires_cuda = False
+        if re.search("(?i)cuda", self.m) is not None:
+            self.requires_cuda = True
+
+    @rfm.run_after('setup')
+    def check_gpu_presence(self):
+        self.gpu_list = [ dev.num_devices for dev in self.current_partition.devices if dev.device_type == 'gpu' ]
```
jjotero, line 49: It feels to me that you might be using this hook in many other tests too. If so, you can move this `check_gpu_presence` function into a hook utility module and then bind the external function into the class as a hook. To bind it, you can either do
```python
import my_hook_utils as hu

class MyTest(rfm.RunOnlyTest):
    ...
    @run_after('setup')
    def check_gpu_presence(self):
        hu.check_gpu_presence(self)
```
or as a one-liner
```python
import my_hook_utils as hu

class MyTest(rfm.RunOnlyTest):
    ...
    run_after('setup')(bind(hu.check_gpu_presence))
```

jjotero (commented):
Looks pretty good! Just one comment on the variables though. In order to use the `required` keyword in the tests, the class attribute must be declared first with the `variable` builtin as 
```python
nsteps = variable(int)
```
If you don't give this a value, it is set as `required` by default. If you set a regular Python attribute as `required`, you won't get all the implicit checking.

tests/reframe/testlib/applications/gromacs/__init__.py
casparvl: Ah, I see! I indeed copied what you did for `num_tasks` in https://github.com/eth-cscs/reframe/pull/1996/ , but failed to realize that this was actually a special case since the `variable` was already declared as part of the `ReFrame` framework. Cool, ok, I'll replace by `variable` to get the type checking :)

tests/reframe/testlib/applications/gromacs/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    modules = required
```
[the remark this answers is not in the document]
casparvl: Hm, fair point. My idea when designing this class was to create a library test for just running a GROMACS module. But indeed, nothing specific is in this library that would make this a 'run-only' test. The person deriving from the test might technically build GROMACS as part of the test, and this would still be a valid library test to derive from. I wouldn't recommend it, but you're right: there's no use in _forcing_ the use of a module here.

tests/reframe/eessi-checks/applications/gromacs.py
casparvl: Agreed, much cleaner. I'm not to worried about the test names: they are a bit messy anyway if you use the automatic module discovery tbh... :)

tests/reframe/eessi-checks/applications/gromacs.py
casparvl: Oh, I like this idea! And it gives me another idea to: I can do the same thing for the requires_cuda test. The regex makes some pretty strong assumption (namely that `cuda` is in the name), but that only works for EB toolchains like `fosscuda` and `intelcuda`. It would fail e.g. with `gompic`. Moving it out of the test and into a `utils` kind of thing allows us to easily improve and/or adjust that check in a single place if things change in the future.

tests/reframe/testlib/applications/gromacs/__init__.py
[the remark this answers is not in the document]
casparvl: No, wait, let me get back to this: I derived the library test from RunOnlyRegressionTest. Doesn't _that_ essentially make a module required?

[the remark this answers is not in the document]
jjotero: Not necessarily. One could have installed gromacs in the system manually. As long as the executable `gmx_mpi` is there, this test can run.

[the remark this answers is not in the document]
casparvl: Fair point. So used to managing things with modules that I sometimes forget what happens on non-HPC systems =)

jjotero:
This looks pretty good to me now! One last comment on the hook-importing from `hooks.py` module. If you think that you're gonna have a few hooks that will always get used together in the test, you can bundle them up into a mixin class and then just do multiple inheritance on the tests. For example:
```python
class RequiresGpusAndCuda(rfm.RegressionMixin):
    # Skip testing GPU-based modules on CPU-based nodes    
    @run_after('setup')
    def skip_gpu_test_on_cpu_nodes(self):
        hooks.skip_gpu_test_on_cpu_nodes(self)

    # Skip testing CPU-based modules on GPU-based nodes
    # (though these would run fine, one is usually not interested in them)
    @run_after('setup')
    def skip_cpu_test_on_gpu_nodes(self):
       hooks.skip_cpu_test_on_gpu_nodes(self)
```
and you feed this into your derived GROMACS test with multiple inheritance.

This is more of a matter of taste than anything though. I think that both multiple inheritance and explicit hook binding work absolutely fine in this example.

casparvl:
Interesting approach with the multiple inheritance. It makes the tests even more compact, but also slightly less descriptive. I think for now, I'd prefer the more verbose and descriptive approach of keeping the hooks in explicitely, rather than through inheritance.

casparvl:
This PR is now outdated, I recreated the test using the ReFrame 'library of tests' design https://github.com/EESSI/software-layer/pull/122

boegel (changes requested):
I wouldn't include the input file itself, only a script to download it from somewhere, and a SHA256 checksum to ensure we got the right file.

boegel:
We should also have a short README file, both in the `test/reframe` subdirectory, and into the test-specific subdirectory (`<whatever>/gromacs/`), with some basic general information + test-specific info (short description of input files, type of run, etc.)

tests/reframe/config/settings.py
```diff
@@ -0,0 +1,86 @@
...
+            'modules_system': 'tmod',
+	    'hostnames': ['login', 'int'],
+            'partitions': [
+                {
+                    'name': 'cpu',
+                    'scheduler': 'slurm',
+                    'launcher': 'srun',
+                    'access':  ['-p cpu'],
+                    'environs': ['builtin'],
+                    'processor': {
```
casparvl, line 15: This _can_ now be taken out, since ReFrame supports autodetection of CPU architecture (not yet of GPU arch though).
https://reframe-hpc.readthedocs.io/en/stable/configure.html?highlight=auto%20detection#auto-detecting-processor-information

tests/reframe/eessi_utils/hooks.py
```diff
@@ -0,0 +1,26 @@
+import reframe as rfm
+import eessi_utils.utils as utils
+
+def skip_cpu_test_on_gpu_nodes(test: rfm.RegressionTest):
+    '''Skip test if GPUs are present, but no CUDA is required'''
+    skip = ( utils.is_gpu_present(test) and not utils.is_cuda_required(test) )
+    if skip:
+        print("GPU is present on this partition, skipping CPU-based test")
+        test.skip_if(True)
+
```
casparvl, line 10: Maybe we can use the logic from flex-alloc-nodes to skip tests that are too large for the current system. Might not work well on PBS though. 

Logic can be taken from here https://github.com/eth-cscs/reframe/blob/8ecb01eb742cfa61f9ef320ed3c5af9d9908dece/reframe/core/schedulers/__init__.py#L412

casparvl: Scratch that: we want to tag, because for the CI on the build nodes, we _only_ want to run the single node tests anyway. It doesn't make sense to already spend resources on multi-node tests at this stage.

casparvl:
Discussing with Victor, it might be better to build an EESSI test implementation on top of their GROMACS library test: https://github.com/eth-cscs/reframe/blob/master/hpctestlib/sciapps/gromacs/benchmarks.py

This library tests implements multiple tests (6 different input files for now). We probably want to run 1 test case in CI, but maybe more in the monitoring.

casparvl:
Closing this PR, as we have one on top of the CSCS test library now: https://github.com/EESSI/software-layer/pull/156

PR #122, opened 2021-07-05
TensorFlow and Horovod test

casparvl:
I recreated the TensorFlow tests from https://github.com/EESSI/software-layer/pull/106 in the library-of-tests fromat, and reusing the same standard `hooks` as I've set up for e.g. `GROMACS`.

To run this test case:
- Make sure that `tests/reframe` is in your `PYTHONPATH` (so that the `eessi_utils` is found)
- Adjust the attached `settings.py` for your system (adapt partition names, #CPU cores, #sockets, # GPUs)
- Make sure you have a new enough ReFrame installation (`3.6.2` or newer should work)
- Make sure you have a flat module naming scheme (the `find_modules` logic is used based on the assumption of a flat module naming scheme). Note that it _does not_ have to be the EESSI software stack per se - it could be some locally installed modules, provided that `module av Horovod` and `module av TensorFlow` return you respective Horovod and TensorFlow modules.

After that, simply run e.g.

```
PYTHONPATH=$PYTHONPATH:$(pwd) reframe --config-file=config/settings_cartesius.py --checkpath eessi-checks/applications/tensorflow2.py -r --performance-report
```

in the `software-layer/tests/reframe` directory.


boegel:
@casparvl Is this still relevant, especially with https://github.com/EESSI/test-suite/pull/38?

casparvl:
Closing this PR. We now have a native TensorFlow distributed test through https://github.com/EESSI/test-suite/pull/38 . This PR (#122) has a TensorFlow test which uses Horovod for distribution. That's nice, if you want to test Horovod, but we can always re-implement that later in the EESSI test suite.

PR #129, opened 2021-08-27
add script to start build node environment

boegel:
This will make the procedure documented at https://eessi.github.io/docs/software_layer/build_nodes a bit easier to follow, we can mostly change it to:

```
git clone https://github.com/EESSI/software-layer.git
cd software-layer
./start_build_node_env.sh
```

You need to clone the `software-layer` repository anyway to get the software install script...

TODO: allow to easily use a different directory than `/tmp`, since that has significant limitations on some systems.
Maybe we need to change the `start_build_node_env.sh` script to make it require a single argument, a path that can be used to create temporary directories? Usage would then become:

```
$ ./start_build_node_env.sh /tmp
```

### DIFF
--- a/start_build_node_env.sh
+++ b/start_build_node_env.sh
@@ -1,8 +1,13 @@
 #!/bin/bash
 
-# create temporary directories
-export EESSI_TMPDIR=/tmp/$USER/EESSI
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <path for temporary directories>" >&2
+    exit 1
+fi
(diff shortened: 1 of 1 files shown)


bedroge (approved):
This is very useful! :-)

Tried it on my laptop, and worked like a charm.

terjekv:
Also, +1 for very useful environment variable names. WRITABLE vs READONLY is a lot more friendly than some of the alternatives. :)

PR #156, opened 2021-12-15
Gromacs EESSI on top of CSCS GROMACS library test

casparvl:
This PR contains an implementation of the GROMACS library test that comes with ReFrame. We 

- Usa a standard set of hooks to skip e.g. GPU tests on CPU nodes, set number tasks per node equal to the number of cores per node, etc
- Tag all tests cases with `monitoring` and a single test case with `CI`
- Parametrize for different node counts (1,2,8,16) and tag accordingly with `singlenode`, `n_small`, `n_medium`, `n_large`.
- Skip running with `-nb gpu` on CPU based nodes.


casparvl:
To test:
- Checkout ReFrame-3.9.2 from github (e.g. to `~/reframe`)
- Checkout `software-layer` with this change (e.g. to `~/software-layer`)
- If running on the EESSI Magic Castle test cluster, it should run with the `config/settings_magic_castle.py` config file for ReFrame. On other cluster, adapt the virtual partitions in that config file to match your own cluster. Be sure to list gpu's as `devices` with name `gpu`. Specify the processor `num_cores` config item for each partition. Modify the `access` item in order to match the scheduler arguments needed for your cluster.
- Run with e.g. `PYTHONPATH=$PYTHONPATH:~/reframe/:~/software-layer/tests/reframe/ reframe -C config/settings_magic_castle.py -c eessi-checks/applications/ -r -t CI -t n_small --performance-report` (the `PYTHONPATH` should contain the right dirs in order to find the `hpctestlib` from ReFrame, and the `utils` and `hooks` from the `~/software-layer/tests/reframe/eessi_utils` directory)

tests/reframe/config/settings.py
```diff
@@ -0,0 +1,87 @@
+site_configuration = {
+    'systems': [
+        {
+            'name': 'example_system',
```
hmeiland, line 4: this looks like cartesius ;) let's make sure we have here a default partition with a launcher that is local e.g. mpirun; so that it does not rely on a non-eessi environment (like slurm)

tests/reframe/config/settings_cartesius.py
```diff
@@ -0,0 +1,119 @@
+site_configuration = {
```
hmeiland, line 1: do we want deployment specific files in here?

tests/reframe/config/settings_magic_castle.py
```diff
@@ -0,0 +1,92 @@
+site_configuration = {
```
hmeiland, line 1: let's then call this system magic castle and have the config sync with the deployment in EESSI, like partition names

tests/reframe/config/system_properties.py
```diff
@@ -0,0 +1,2 @@
+
+ncorespernode=16
```
hmeiland, line 2: how is this file used?

tests/reframe/eessi-checks/prgenv/mpi.py
```diff
@@ -0,0 +1,20 @@
...
+class MpiHelloWorld(rfm.RegressionTest):
+    def __init__(self):
+        # We don't define these here to keep tests generic
+        # Sensible systems & programming environments should be defined in your site configuration file
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+
+        self.sourcepath = 'mpi_hello_world.c'
+        self.maintainers = ['casparvl']
+        self.num_tasks_per_node = -2
```
hmeiland, line 18: whiy is this definition, if you redefine later as 16?

tests/reframe/eessi-checks/prgenv/mpi.py
```diff
@@ -0,0 +1,20 @@
...
+# Try to use an import to define all site-specific things
+# import system_properties
+
+@rfm.simple_test
+class MpiHelloWorld(rfm.RegressionTest):
+    def __init__(self):
+        # We don't define these here to keep tests generic
+        # Sensible systems & programming environments should be defined in your site configuration file
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
```
hmeiland, line 14: shouldn't we need to set this to load openmpi and gcc; it will not have anything by default

hmeiland (changes requested):
some code in here for gpu; while there is not gpu support in EESSI today; let's make sure we review this again when it becomes available....

tests/reframe/config/settings_cartesius.py
casparvl: No, sorry, didn't realize this was still in there :) I'll take it out.

tests/reframe/config/settings_magic_castle.py
casparvl: > have the config sync with the deployment in EESSI

Not sure what you mean with this. I'm not sure how 'standardized'the magic castle deployment was and if e.g. partition names will be the same next time. 

In any case, it _does_ serve as a 'neutral' (i.e. not related to a particular institute) example of a valid configuration file. I'll add some notes in the file to indicate that it is an example.

tests/reframe/config/settings.py
casparvl: Yep, I'll take it out. We have the Magic Castle settings file as an example now.

tests/reframe/config/system_properties.py
casparvl: it isn't anymore. It was a solution I had before ReFrame had support for describing processor properties in the config file... I took it out now

tests/reframe/eessi-checks/prgenv/mpi.py
casparvl: All of that is provided by loading the module. A programming environment is only needed if you want to run tests that e.g. compile and run code. Then, you can define a programming environment to load a fixed collection of modules, and test that collection. (think about hello-world tests where you test e.g. GCC and intel compilers, some basic flags, etc)

tests/reframe/eessi-checks/prgenv/mpi.py
casparvl: what a mess: another file that I used to test something... I'll go through the whole file list of the PR and check if there are any other leftover files :)

casparvl:
We'll move the tests in a separate repo. Closing in favor of https://github.com/EESSI/test-suite/pull/9

PR #159, opened 2021-12-16
fix init script for archspec 0.1.3

### LINKED ISSUE #142: host detection script is broken due to changes in archspec 0.1.3
The `archspec.cpu.host` function was changed in archspec 0.1.3 (see https://github.com/archspec/archspec/pull/53) to return the best matching generic CPU microarchitecture level if there's no exact match.

These levels are values like `x86-64-v4`, and have a `'generic'` value for vendor, see https://github.com/archspec/archspec-json/pull/31/files.

This causes trouble where we expect to find an actual vendor string like `GenuineIntel` (which we map to `intel`) or `AuthenticAMD` (which we map to `amd`), for example in `det_host_triple` in `init/eessi_software_subdir_for_host.py` or in `eessi_software_subdir_for_host.py`.

boegel:
fixes #142 (although I really think that the `archspec` API could be improved to avoid all this copy-pasting...)

We have largely the same logic in `eessi_software_subdir.py` (used by build script) and `init/eessi_software_subdir_for_host.py` (used by `init/bash` script), we should fix that too at some point...

The build script should leverage the script from `init/` (but that's a follow-up PR imho...)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8]
+        python: [3.6, 3.7, 3.8, 3.9, '3.10']
       fail-fast: false
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


bedroge:
@boegel should we also add Python 3.9 to the tests (especially since we're using that in our Prefix)?

boegel:
@bedroge Makes sense, done, also for Python 3.10

bedroge (approved):
lgtm

PR #166, opened 2022-02-04
rough starting point for eessi-init command to be installed through pip

hmeiland:
checking the proposed location and concept;


ocaisa:
There's now a repository for this at https://github.com/EESSI/eessi-cli (still quite basic, but we've at least started)

PR #170, opened 2022-02-10
proposal for config file and script for managing datasets

hmeiland:
adding datasets yaml file, tarball creator and tree for scripts with WRF benchmark scripts


boegel:
I don't think this is going anywhere...

Actually setting up a separate repository for hosting datasets should be carefully considered, since it would have large implications in terms of maintenance, and also using it needs to be done with care w.r.t. CernVM-FS cache, etc.

If we want to go through with something like this, there should be a clear demand/incentive, and a detailed discussion on how to structure & maintain such a repository.

So, closing for now.

Issue #108, opened 2021-05-17
Enabling end-user GPU support and ABI-compatible overrides


ocaisa:
# Accelerator support

In `2021.03` we added the machinery to enable adding GPU support to EESSI by adding a new trusted directory ( where the GPU runtime libraries are expected to be found) to the compatability loader. This approach doesn't do any sanity-checking though and leaves the door open to things not working (for example because the driver is too old for the CUDA shipped with EESSI). I think we can do better than this:

- Add a _single_ variant symlink to the CMVFS config that points (by default) to `/opt/eessi`
    - Since the pilot version number occurs right at the root, this will have to be something like `/cvmfs/pilot.eessi-hpc.org/host_injections`
- Create a script that controls the structure underneath this folder
    - Driver libraries are related to the compatability layer so I would suggest, e.g., `/cvmfs/pilot.eessi-hpc.org/host_injections/2021.03/compat/linux/x86_64/lib` (and this will be added as a trusted glibc dir for the compatability layer)
    - The script will check that driver libraries are adequate to support the requested stack (and tell you what to do if they are not)
    - It will search for all the libraries required to use the CUDA-enable EESSI stack (see [recent OpenSSL easyblock](https://github.com/easybuilders/easybuild-easyblocks/pull/2429) PR for the machinery to do this)
    - It will place symlinks to the libraries in the designated subfolder
    - It will perform sanity checks to verify that the setup works, e.g.,
      ```
      LD_LIBRARY_PATH=/cvmfs/pilot.eessi-hpc.org/host_injections/2021.03/compat/linux/x86_64/lib my_cuda_verification_exec
      ```
- The script can tell the user how to add all the CUDA-enabled modules to their environment
    - I think CUDA-enabled _modules_ should be installed in a separate path, for a flat naming scheme then making the modules available just means an additional `module use ...`. For a hierarchical scheme, it should mean setting an additional environment variable.
    - EasyBuild has all the machinery needed to support this, we just need to build this configuration into our hooks.

# RPATH overrides

We can also use this approach to allow for controlled ABI-compatible overrides of the software EESSI provides:

- For RPATH injections, this is related to the software installations themselves so should follow the existing subdirectory structure of EESSI underneath the symlink, e.g., `/cvmfs/pilot.eessi-hpc.org/host_injections/2021.03/software/linux/x86_64/intel/skylake_avx512/rpath_overrides/{software}/{version?}/lib`
[...]

Issue #69, opened 2021-02-08
Fail to download CUDA using the python in $EPREFIX


ocaisa:
Ran into an issue when trying to install CUDA using the Python from `$EPREFIX`:
```
== FAILED: Installation ended unsuccessfully (build directory: /tmp/CUDAcore/11.1.1/system-system): build failed (first 300 chars): Unexpected error occurred when trying to download https://developer.download.nvidia.com/compute/cuda/11.1.1/local_installers/cuda_11.1.1_455.32.00_linux.run to /project/def-sponsor00/easybuild/sources/c/CUDAcore/cuda_11.1.1_455.32.00_linux.run: signed integer is greater than maximum (took 0 sec)
```
It looks like this is a bug in Python 3.8 (from https://bugs.python.org/issue42853)

boegel:
See also https://github.com/easybuilders/easybuild-framework/issues/3455

ocaisa:
Building some CUDA examples also fails, I think this might be due to the fact that the prefix GCC compiler is ahead of the `foss` GCC (GCC10 versus GCC9):
```
[ocaisa@gpu-node1 reduction]$ make HOST_COMPILER=$(which g++)
>>> GCC Version is greater or equal to 4.7.0 <<<
/project/def-sponsor00/easybuild/software/CUDAcore/11.1.1/bin/nvcc -ccbin /cvmfs/pilot.eessi-hpc.org/2020.12/software/x86_64/amd/zen2/software/GCCcore/9.3.0/bin/g++ -I../../Common -m64 --std=c++17 -gencode arch=compute_70,code=sm_70 -gencode arch=compute_75,code=sm_75 -gencode arch=compute_80,code=sm_80 -gencode arch=compute_86,code=sm_86 -gencode arch=compute_86,code=compute_86 -o reduction_kernel.o -c reduction_kernel.cu
/cvmfs/pilot.eessi-hpc.org/2020.12/compat/linux/x86_64/usr/include/math.h: In function 'int iseqsig(_T1, _T2)':
/cvmfs/pilot.eessi-hpc.org/2020.12/compat/linux/x86_64/usr/include/math.h:1374:39: error: expected ')' before numeric constant
 1374 |   typedef decltype (__MATH_EVAL_FMT2 (__x, __y)) _T3;
      |                                   ~   ^~~
      |                                       )
```

ocaisa:
[...]

PR #172, opened 2022-05-12
Add CUDA support [WIP]

huebner-m:
This WIP PR follows up on the work done in the [Hackathons](https://github.com/EESSI/hackathons/tree/05_gpu).
The following features have been implemented:
- Check that NVIDIA drivers are installed on host (via `nvidia-smi`)
- Get latest CUDA compat libs from NVIDIA website (automatically picking the correct OS version etc.)
- Check that `host_injections` is a writable path
- Install CUDA compat libs in `host_injections` (using the appropriate rpm or deb files and tools)
- Check if CUDA is already installed as a module
- Check if disk space in `host_injections` is sufficient to install CUDA
- Install CUDA using EasyBuild
- Add test based on CUDA samples to check if GPU support works
- Add EasyBuild hook to tag software that depends on CUDA with the property `gpu` (based on https://github.com/easybuilders/JSC/blob/2022/Custom_Hooks/eb_hooks.py#L335)
- Add Lmod plugin that hides software with the property `gpu` if CUDA is not installed

Open tasks:
- [ ] Only download/install CUDA compat libs from NVIDIA website if necessary
- [ ] Update Lmod version to make plugin work, also see: https://github.com/TACC/Lmod/issues/552
- [ ] Ship CUDA samples with EESSI
- [ ] Ship rpm/deb tools to install CUDA compat libs with EESSI's compat layer

Related issues:
- [Enabling end-user GPU support and ABI-compatible overrides](https://github.com/EESSI/software-layer/issues/108)
- [CUDA support](https://github.com/EESSI/compatibility-layer/issues/71)
- [Fail to download CUDA using the python in $EPREFIX](https://github.com/EESSI/software-layer/issues/69)


eb_hooks.py
```diff
@@ -7,7 +7,7 @@
 from easybuild.tools.systemtools import AARCH64, POWER, get_cpu_architecture
 
 EESSI_RPATH_OVERRIDE_ATTR = 'orig_rpath_override_dirs'
-
+CUDA_ENABLED_TOOLCHAINS = ["pmvmklc", "gmvmklc", "gmvapich2c", "pmvapich2c"]
```
ocaisa, line 10: This list is not complete (nvompi,...). It doesn't really matter since we control what is included as a toolchains in EESSI

eb_hooks.py
```diff
@@ -41,13 +41,32 @@ def get_rpath_override_dirs(software_name):
...
 
+def inject_gpu_property(ec):
+    ec_dict = ec.asdict()
+    # Check if CUDA is in the dependencies, if so add the GPU Lmod tag
+    if (
+        "CUDA" in [dep[0] for dep in iter(ec_dict["dependencies"])]
+        or ec_dict["toolchain"]["name"] in CUDA_ENABLED_TOOLCHAINS
+    ):
+        key = "modluafooter"
+        value = 'add_property("arch","gpu")'
```
ocaisa, line 52: I think `gpu` is a recognised property in Lmod so a good choice for now. Once we add AMD support it will get more complicated.

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+current_dir=$(dirname $(realpath $0))
+
+# Get arch type from EESSI environment
+eessi_cpu_family="${EESSI_CPU_FAMILY:-x86_64}"
+
+# Get OS family
+# TODO: needs more thorough testing
+os_family=$(uname | tr '[:upper:]' '[:lower:]')
+
+# Get OS version
```
ocaisa, line 30: @boegel Does EB do this already, can we hook into that?

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+  echo "Your NVIDIA driver version is too old, please update first.."
+  exit 1
+fi
+
+
+# Check if the CUDA compat libraries are installed and compatible with the target CUDA version
+# if not find the latest version of the compatibility libraries and install them
+
+# get URL to latest CUDA compat libs, exit if URL is invalid
+latest_cuda_compat_url="$(./get_latest_cuda_compatlibs.sh ${os} ${ver} ${eessi_cpu_family})"
```
ocaisa, line 87: Don't use `./` here, instead use something like `$(dirname "$BASH_SOURCE")/` as this could be run from another directory

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+driver_cuda_version=$(nvidia-smi  -q --display=COMPUTE | grep CUDA | awk 'NF>1{print $NF}' | sed s/\\.//)
+eessi_cuda_version=$(LD_LIBRARY_PATH=/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/compat/:$LD_LIBRARY_PATH nvidia-smi  -q --display=COMPUTE | grep CUDA | awk 'NF>1{print $NF}' | sed s/\\.//)
+if [ "$driver_cuda_version" -gt "$eessi_cuda_version" ]; then  echo "You need to update your CUDA compatability libraries"; fi
+
+# Check if our target CUDA is satisfied by what is installed already
+# TODO: Find required CUDA version and see if we need an update
+
+# If not, grab the latest compat library RPM or deb
+# download and unpack in temporary directory, easier cleanup after installation
+mkdir -p tmp
```
ocaisa, line 114: Use `mktemp -d` instead here and store it in `$tmp`

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+cd tmp
```
ocaisa, line 115: Store `$PWD` so that you can get back from `cd $tmp`

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+compat_file=${latest_cuda_compat_url##*/}
+wget ${latest_cuda_compat_url}
+
+# Unpack it
+# (the requirements here are OS dependent, can we get around that?)
+# (for rpms looks like we can use https://gitweb.gentoo.org/repo/proj/prefix.git/tree/eclass/rpm.eclass?id=d7fc8cf65c536224bace1d22c0cd85a526490a1e)
+# (deb files can be unpacked with ar and tar)
+file_extension=${compat_file##*.}
+if [[ ${file_extension} == "rpm" ]]; then
+  rpm2cpio ${compat_file} | cpio -idmv
```
ocaisa, line 125: We need to have these supported in the compat layer (and ensure they are being used from there)

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+latest_cuda_compat_url="$(./get_latest_cuda_compatlibs.sh ${os} ${ver} ${eessi_cpu_family})"
+ret=$?
+if [ $ret -ne 0 ]; then
+  echo $latest_cuda_compat_url
+  exit 1
+fi
+
+# Create a general space for our NVIDIA compat drivers
+if [ -w /cvmfs/pilot.eessi-hpc.org/host_injections ]; then
+  mkdir -p /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia
```
ocaisa, line 96: Make sure you can write to that dir first?

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+
+# return to initial dir
+cd $current_dir
+
+###############################################################################################
+###############################################################################################
+# Install CUDA
+# TODO: Can we do a trimmed install?
+# if modules dir exists, load it for usage within Lmod
+if [ -d /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/modules/all ]; then
```
ocaisa, line 163: I would probably use `$EESSI_SOFTWARE_PATH` here but swap out `versions` for `host_injections`, that is going to be a lot more flexible. You can still check for the existence of `<dir>/software/CUDA/<version>` later in the Lmod plugin

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,186 @@
...
+  #   (CUDA is a binary installation so no need to worry too much about this)
+  # TODO: The install is pretty fat, you need lots of space for download/unpack/install (~3*5GB), need to do a space check before we proceed
+  avail_space=$(df --output=avail /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/ | tail -n 1 | awk '{print $1}')
+  if (( ${avail_space} < 16000000 )); then
+    echo "Need more disk space to install CUDA, exiting now..."
+    exit 1
+  fi
+  # install cuda in host_injections
+  module load EasyBuild
+  eb --installpath=/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/ CUDA-${install_cuda_version}.eb
```
ocaisa, line 182: Check return value?

gpu_support/test_cuda
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Test CUDA
+module use /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/modules/all/
+module load CUDA
```
ocaisa, line 5: Check return value?

gpu_support/test_cuda
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Test CUDA
+module use /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/modules/all/
```
ocaisa, line 4: Check path existence?

gpu_support/test_cuda
```diff
@@ -0,0 +1,39 @@
...
+module load CUDA
+tmp_dir=$(mktemp -d)
+cp -r $EBROOTCUDA/samples $tmp_dir
```
ocaisa, line 7: CUDA no longer ships samples, there is a plan to ship the (compiled) CUDA samples with EESSI

gpu_support/test_cuda
```diff
@@ -0,0 +1,39 @@
...
+
+if [ $? -eq 0 ] 
+then
+  # Set the color variable
+  green='\033[0;32m'
+  # Clear the color after that
+  clear='\033[0m'
+  echo -e ${green}
+  echo "Congratulations, your GPU is working with EESSI!"
+  echo "  - To build CUDA enabled modules use /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/ as your EasyBuild prefix"
```
ocaisa, line 21: If we use a different path, we'll need to tell them to use our EasyBuild hook

init/SitePackage.lua
```diff
@@ -0,0 +1,29 @@
...
+			return true
+		end
+	end
+	return ok, err
+end
+
+local function visible_hook(modT)
+	local frameStk  = require("FrameStk"):singleton()
+	local mt        = frameStk:mt()
+	local cudaDir = exists('/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/modules/')
```
ocaisa, line 20: Oh, now I see how using `EESSI_SOFTWARE_PATH` might be hard for this. Can we use `os.getenv` here?

ocaisa (changes requested):
Encouraging!

gpu_support/add_gpu_support.sh
ocaisa: Whoops, sorry, did not see the `-w`

gpu_support/add_gpu_support.sh
ocaisa: Can we can do that already in the current pilot @bedroge ?

init/SitePackage.lua
ocaisa: We also probably also need an ENVVAR setting to force display of these modules

gpu_support/test_cuda
ocaisa: Actually in general they will need to use the hook as the `gpu` property needs to be added to the module files they build.

ocaisa:
There are indications that we may be allowed to ship the CUDA runtime with EESSI, that would mean we wouldn't (necessarily) need to install CUDA _unless_ people want to actually build their own software on top of EESSI.

I would go with this PR as is right now, but in a future pilot we should make that installation optional (making another code branch that only creates the software installation directory so that the Lmod plugin will still work). In that scenario though, we will need some major tweaking to the CUDA module shipped with EESSI, it would need conditionals based on what is available on the host.

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,192 @@
...
+ret=$?
+if [ $ret -ne 0 ]; then
+  echo $latest_cuda_compat_url
+  exit 1
+fi
+
+# Create a general space for our NVIDIA compat drivers
+if [ -w /cvmfs/pilot.eessi-hpc.org/host_injections ]; then
+  mkdir -p /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia
+  cd /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia
```
ocaisa, line 97: Before dropping in here, I would store where we started from and return us there at the end of the script (or at any earlier exit point)

gpu_support/add_gpu_support.sh
```diff
@@ -0,0 +1,192 @@
...
+
+if [ ! -e latest ] ; then
+  echo "Symlink to latest cuda compat lib version is broken, exiting now..."
+  exit 1
+fi
+
+# Create the space to host the libraries
+mkdir -p /cvmfs/pilot.eessi-hpc.org/host_injections/${EESSI_PILOT_VERSION}/compat/${os_family}/${eessi_cpu_family}
+# Symlink in the path to the latest libraries
+if [ ! -d "/cvmfs/pilot.eessi-hpc.org/host_injections/${EESSI_PILOT_VERSION}/compat/${os_family}/${eessi_cpu_family}/lib" ]; then
```
ocaisa, line 152: What should we do if it _does_ exist? We should probably check that it is the symlink we expected and error if it is not.

ocaisa: `host_injections_dir` is `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia` so I would set that early and use it where possible (like around L95).

Similarly, `/cvmfs/pilot.eessi-hpc.org/host_injections/${EESSI_PILOT_VERSION}/compat/${os_family}/${eessi_cpu_family}` is used in a few places, I would have that as a variable as early as possible and then reuse that.

gpu_support/add_gpu_support.sh
ocaisa: Ah, sorry, you have that

gpu_support/add_gpu_support.sh
ocaisa: In general, using `$EESSI_SOFTWARE_PATH` with the `versions`/`host_injections` swap is a good way to give a default way to extend the stack that can be accounted for in our initialisation scripts.

gpu_support/add_gpu_support.sh
bedroge: It's not there yet. I can try to look into installing it, together with some other packages/updates (Lmod?). Especially updates are not so straightforward, though.

ocaisa: It would be interesting to add the additional packages for now, but I would probably hold off on updates and just wait for the next pilot.

huebner-m: > We also probably also need an ENVVAR setting to force display of these modules

In which scenario do you want to force display the modules? The assumption is that if `cudaDir` exists, CUDA itself is also available and only then modules with the `gpu` tag will be shown.

ocaisa: I was thinking of login nodes without GPUs, but I guess that scenario is still covered. @kostrzewa was the one who originally raised this so perhaps he see's a reason.

Now that I think about though, what is actually required is an option to skip the various verification steps for that use case, so that you can enable CUDA even from a node without GPUs.

huebner-m: Not sure why my comment ended up here...

Shouldn't it be enough to `module use cuda_modules_path` to enable CUDA on nodes without GPUs?

ocaisa: What I mean is, what if someone wanted to run `add_gpu_support.sh` from a login node? This would fail because there is no `nvidia-smi` and there is no way to override the checks.

gpu_support/add_gpu_support.sh
huebner-m: Addressed in commit https://github.com/EESSI/software-layer/pull/172/commits/01844c6e30ba43902769bf3157e12d9d40cb204b

EESSI-pilot-install-software.sh
```diff
@@ -108,6 +108,9 @@ module --force purge
 # ignore current $MODULEPATH entirely
 module unuse $MODULEPATH
 module use $EASYBUILD_INSTALLPATH/modules/all
+if [ -d /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/modules/all ]; then
```
ocaisa, line 111: This directory should be `${EESSI_SITE_MODULEPATH}`?

init/eessi_environment_variables
```diff
@@ -51,7 +55,7 @@ if [ -d $EESSI_PREFIX ]; then
           fi
 
           if [ -d $EESSI_MODULEPATH ]; then
-            export EESSI_MODULEPATH=$EESSI_MODULEPATH
+            export EESSI_MODULEPATH="$EESSI_MODULEPATH $eessi_gpu_software_path"
```
ocaisa, line 58:
```suggestion
            export EESSI_MODULEPATH=$EESSI_MODULEPATH
```

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,152 @@
+# Drop into the prefix shell or pipe this script into a Prefix shell with
```
ocaisa, line 1:
```suggestion
#!/bin/bash

# Drop into the prefix shell or pipe this script into a Prefix shell with
```

gpu_support/add_amd_gpu_support.sh
```diff
@@ -0,0 +1,12 @@
+cat << EOF
```
ocaisa, line 1:
```suggestion
#!/bin/bash

cat << EOF
```

ocaisa (changes requested):
I think it's time to remove the WIP label and get someone else feedback on this!

gpu_support/add_gpu_support.sh
ocaisa: @huebner-m I think we still need this boolean right?

huebner-m: Yes, the boolean has been added in commit https://github.com/EESSI/software-layer/pull/172/commits/5f82658ee4ea9e5e038ab5d34dacc8b2f8b4b708

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,187 @@
...
+  if command -v nvidia-smi > /dev/null 2>&1; then
+    nvidia-smi > /dev/null 2>&1
+    if [ $? -ne 0 ]; then
+      echo "nvidia-smi was found but returned error code, exiting now..." >&2
+      exit 1
+    fi
+    echo "nvidia-smi found, continue setup."
+  else
+    echo "nvidia-smi not found, exiting now..." >&2
+    exit 1
```
ocaisa, line 22:
```suggestion
    echo
    echo "If you do not have a GPU on this device but wish to force the installation,"
    echo "please set the environment variable INSTALL_WO_GPU=true"
    exit 1
```

ocaisa (changes requested):
Allow using an environment variable to skip GPU checks

eb_hooks.py
ocaisa, line 10:
```suggestion
CUDA_ENABLED_TOOLCHAINS = ["fosscuda", "gcccuda", "gimpic", "giolfc", "gmklc", "golfc", "gomklc", "gompic", "goolfc", "iccifortcuda", "iimklc", "iimpic", "intelcuda", "iomklc", "iompic, "nvompic", "nvpsmpic"]
```
That's the full list of currently supported CUDA enabled toolchains in the EB release

EESSI-pilot-install-software.sh
```diff
@@ -108,6 +108,10 @@ module --force purge
 # ignore current $MODULEPATH entirely
 module unuse $MODULEPATH
 module use $EASYBUILD_INSTALLPATH/modules/all
+if [ ! -z "${EESSI_SITE_MODULEPATH}" ]; then
+  echo_green "Add ${EESSI_SITE_MODULEPATH} to \$MODULEPATH for GPU support!"
```
ocaisa, line 112:
```suggestion
  echo_green "Add ${EESSI_SITE_MODULEPATH} to \$MODULEPATH for site installations support!"
```

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,190 @@
...
+
+# get URL to latest CUDA compat libs, exit if URL is invalid
+latest_cuda_compat_url="$($(dirname "$BASH_SOURCE")/get_latest_cuda_compatlibs.sh ${os} ${ver} ${eessi_cpu_family})"
+ret=$?
+if [ $ret -ne 0 ]; then
+  echo $latest_cuda_compat_url
+  exit 1
+fi
+latest_driver_version="${latest_cuda_compat_url%-*}"
+latest_driver_version="${latest_driver_version##*_}"
```
ocaisa, line 110:
```suggestion
latest_driver_version="${latest_driver_version##*-}"
```
I needed this change for this to work for me

huebner-m: This has been addressed in your PR (https://github.com/EESSI/software-layer/pull/172/commits/75ce85042420e5fc050811590b2115a6344da01b) with a small modification in commit https://github.com/EESSI/software-layer/pull/172/commits/17b7662dee1e3c8fdec5aa954b60037df5462f88

gpu_support/README.md
```diff
@@ -0,0 +1,10 @@
+# How to add GPU support
+The collection of scripts in this directory enables you to add GPU support to your setup.
+Note that currently this means that CUDA support can be added for Nvidia GPUs. AMD GPUs are not yet supported (feel free to contribute that though!).
+To enable the usage of CUDA in your setup, simply run the following script:
+```
+./add_nvidia_gpu_support.sh
+```
+## Prerequisites and tips
+* You need write permissions to `/cvmfs/pilot.eessi-hpc.org/host_injections` (which by default is a symlink to `/opt/eessi` but can be configured in your CVMFS config file to point somewhere else). If you would like to make a system-wide installation you should change this in your configuration to point somewhere on a shared filesystem.
+* If you want to install CUDA on a node without GPUs (e.g. on a login node where you want to be able to compile your CUDA-enabled code), you should `export INSTALL_WO_GPU=true` in order to skip checks and tests that can only succeed if you have access to a GPU.
```
ocaisa, line 10:
```suggestion
* If you want to install CUDA on a node without GPUs (e.g. on a login node where you want to be able to compile your CUDA-enabled code), you should `export INSTALL_WO_GPU=true` in order to skip checks and tests that can only succeed if you have access to a GPU. This approach is not recommended as there is a chance the CUDA compatibility library installed is not compatible with the existing CUDA driver on GPU nodes (and this will not be detected).
```

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,180 @@
...
+    fi
+
+    if [ ${latest_driver_version//./} -gt ${eessi_driver_version//./} ]; then
+      install_compat_libs=true
+    else
+      echo "CUDA compat libs are up-to-date, skip installation."
+    fi
+
+    if [ "${install_compat_libs}" == true ]; then
+      source $(dirname "$BASH_SOURCE")/install_cuda_compatlibs.sh $latest_cuda_compat_url
```
ocaisa, line 140:
```suggestion
      bash $(dirname "$BASH_SOURCE")/install_cuda_compatlibs.sh $latest_cuda_compat_url
```
I think we don't need to `source` this anymore

ocaisa (changes requested):

ocaisa:
@huebner-m and I were discussing this today and we came up with a plan for how to implement the CUDA support:

- Add an eb hook for the CUDA build that tweaks the software installation directory for CUDA to be `${EESSI_SOFTWARE_PATH/versions/host_injections}` but keeping the module installed in the normal directory
- The eb hook also injects a `modluafooter` that checks that the installation folder exists (and maybe some other criteria), and errors otherwise with a message about how to use the CUDA install script
- The install script is stopped after the install before it creates a module for CUDA. The sanity check step is run separately (with `--sanity-check-only` since the module exists).

This means that CUDA is treated by EESSI the same way it is seen by users, if CUDA is in the right place under `host_injections` things will work. With this approach, rpaths all point into `host_injections` so need for other kinds of tweaking.

ocaisa:
It is possible to install the compat libraries without needing to worry about the OS: https://docs.nvidia.com/deploy/cuda-compatibility/index.html#manually-installing-from-runfile

This is currently being looked at in EasyBuild: https://github.com/easybuilders/easybuild-easyconfigs/pull/15892

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,185 @@
...
+# Check that the CUDA driver version is adequate
+# (
+#  needs to be r450 or r470 which are LTS, other production branches are acceptable but not
+#  recommended, below r450 is not compatible [with an exception we will not explore,see
+#  https://docs.nvidia.com/datacenter/tesla/drivers/#cuda-drivers]
+# )
+# only check first number in case of multiple GPUs
+if [[ "${install_wo_gpu}" != "true" ]]; then
+  driver_version=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader | tail -n1)
+  driver_version="${driver_version%%.*}"
```
ocaisa, line 48: This should be `driver_major_version`

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,203 @@
...
+  fi
+  rm -rf ${tmpdir}
+fi
+
+# install p7zip, this will be used to install the CUDA compat libraries from rpm
+# the rpm and deb files contain the same libraries, so we just stick to the rpm version
+module avail 2>&1 | grep -i p7zip &> /dev/null
+if [[ $? -eq 0 ]]; then
+  echo "p7zip module found! No need to install p7zip again, proceeding with installation of compat libraries"
+else
```
ocaisa, line 95: Probably best to just error out here, `p7zip` should be part of the software layer (it isn't yet but it will be)

init/SitePackage.lua
```diff
@@ -0,0 +1,48 @@
...
+			modT.isVisible = false
+		end
+	end
+end
+
+local function cuda_load_hook(t)
+	local frameStk  = require("FrameStk"):singleton()
+	local simpleName = string.match(t.modFullName, "(.-)/")
+	if string.match(simpleName, 'CUDA') ~= nil then
+		local cudaDir = string.gsub(os.getenv('EESSI_SOFTWARE_PATH') or "", 'versions', 'host_injections')
```
ocaisa, line 34: This is probably not specific enough, should include `CUDA` and even the CUDA version (since we know what it is).

One issue with this is that this means you can't mix different software from different `EESSI_SOFTWARE_PATH` even though there is no practical problem with this.

ocaisa (changes requested):

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,204 @@
...
+###############################################################################################
+###############################################################################################
+# Install CUDA
+# TODO: Can we do a trimmed install?
+# if modules dir exists, load it for usage within Lmod
+cuda_install_dir="${EESSI_SOFTWARE_PATH/versions/host_injections}"
+mkdir -p ${cuda_install_dir}
+# only install CUDA if specified version is not found
+if [ -d ${cuda_install_dir}/software/CUDA/${install_cuda_version} ]; then
+  echo "CUDA module found! No need to install CUDA again, proceeding with tests"
```
ocaisa, line 67: This is not quite accurate, you've tested for the software here, not the module

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,204 @@
...
+  avail_space=$(df --output=avail ${cuda_install_dir}/ | tail -n 1 | awk '{print $1}')
+  if (( ${avail_space} < 16000000 )); then
+    echo "Need more disk space to install CUDA, exiting now..."
+    exit 1
+  fi
+  # install cuda in host_injections
+  module load EasyBuild
+  # we need the --rebuild option, since the module file is shipped with EESSI
+  tmpdir=$(mktemp -d)
+  eb --rebuild --installpath-modules=${tmpdir} --installpath=${cuda_install_dir}/ CUDA-${install_cuda_version}.eb
```
ocaisa, line 81: Right now this makes testing difficult as the actual CUDA module in EESSI is not available. You might be better off here checking if the CUDA module exists and if so prefixing this command with  `EASYBUILD_INSTALLPATH_MODULES=${tmpdir} `

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,204 @@
...
+    exit 1
+  fi
+  rm -rf ${tmpdir}
+fi
+
+# Install p7zip, this will be used to install the CUDA compat libraries from rpm.
+# The rpm and deb files contain the same libraries, so we just stick to the rpm version.
+# If p7zip is missing from the software layer (for whatever reason), we need to install it.
+# This has to happen in host_injections, so we check first if it is already installed there.
+module use ${cuda_install_dir}/modules/all/
```
ocaisa, line 94: Maybe do this conditionally (i.e, only if this directory exists)

gpu_support/add_nvidia_gpu_support.sh
ocaisa, line 81:
```suggestion
  # we need the --rebuild option and a random dir for the module if the module file is shipped with EESSI
  if [ -f ${EESSI_SOFTWARE_PATH}/modules/all/CUDA/${install_cuda_version}.lua ]; then
    tmpdir=$(mktemp -d)
    extra_args="--rebuild --installpath-modules=${tmpdir}"
  fi
  eb ${extra_args} --installpath=${cuda_install_dir}/ CUDA-${install_cuda_version}.eb
```

gpu_support/add_nvidia_gpu_support.sh
ocaisa: You are right to leave this for now, it's required if anyone wants to test this (since `p7zip` is still not part of EESSI)

gpu_support/test_cuda.sh
```diff
@@ -0,0 +1,54 @@
...
+ret=$?
+if [ $ret -ne 0 ]; then
+  echo "Could not load CUDA even though modules path exists..."
+  exit 1
+fi
+tmp_dir=$(mktemp -d)
+cp -r $EBROOTCUDA/samples $tmp_dir
+current_dir=$PWD
+cd $tmp_dir/samples/1_Utilities/deviceQuery
+make HOST_COMPILER=$(which g++) -j
```
ocaisa, line 21: `g++` here needs to be coming from either the compat layer or EESSI to ensure we are using the correct loader. Since we already load the CUDA module, we might as well load GCC above.

gpu_support/test_cuda.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Test CUDA
+cuda_install_dir="${EESSI_SOFTWARE_PATH/versions/host_injections}"
+if [ -d ${cuda_install_dir}/modules/all ]; then
+  module use ${cuda_install_dir}/modules/all/
+else
+  echo "Cannot test CUDA, modules path does not exist, exiting now..."
+  exit 1
+fi
+module load CUDA
```
ocaisa, line 11: We should probably load the specific version of CUDA here.

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,204 @@
...
+module avail 2>&1 | grep -i p7zip &> /dev/null
+if [[ $? -eq 0 ]]; then
+  echo "p7zip module found! No need to install p7zip again, proceeding with installation of compat libraries"
+else
+  # install p7zip in host_injections
+  module load EasyBuild
+  eb --robot --installpath=${cuda_install_dir}/ p7zip-${install_p7zip_version}.eb
```
ocaisa, line 101: Unfortunately you don't have the same configuration as EESSI when you run this command, so it goes off and (unnecessarily) builds `binutils`. You need:
```
export EASYBUILD_IGNORE_OSDEPS=1
export EASYBUILD_SYSROOT=${EPREFIX}
export EASYBUILD_RPATH=1
export EASYBUILD_FILTER_ENV_VARS=LD_LIBRARY_PATH
export EASYBUILD_FILTER_DEPS=Autoconf,Automake,Autotools,binutils,bzip2,cURL,DBus,flex,gettext,gperf,help2man,intltool,libreadline,libtool,Lua,M4,makeinfo,ncurses,util-linux,XZ,zlib
export EASYBUILD_MODULE_EXTENSIONS=1
```

ocaisa:
For me to get this working out of the box _right now_ I needed
```
diff --git a/gpu_support/add_nvidia_gpu_support.sh b/gpu_support/add_nvidia_gpu_support.sh
index 025a1be..9cf4b70 100755
--- a/gpu_support/add_nvidia_gpu_support.sh
+++ b/gpu_support/add_nvidia_gpu_support.sh
@@ -76,9 +76,12 @@ else
   fi
   # install cuda in host_injections
   module load EasyBuild
-  # we need the --rebuild option, since the module file is shipped with EESSI
-  tmpdir=$(mktemp -d)
-  eb --rebuild --installpath-modules=${tmpdir} --installpath=${cuda_install_dir}/ CUDA-${install_cuda_version}.eb
+  # we need the --rebuild option and a random dir for the module if the module file is shipped with EESSI
+  if [ -f ${EESSI_SOFTWARE_PATH}/modules/all/CUDA/${install_cuda_version}.lua ]; then
+    tmpdir=$(mktemp -d)
+    extra_args="--rebuild --installpath-modules=${tmpdir}"
+  fi
+  eb ${extra_args} --installpath=${cuda_install_dir}/ CUDA-${install_cuda_version}.eb
   ret=$?
   if [ $ret -ne 0 ]; then
     echo "CUDA installation failed, please check EasyBuild logs..."
@@ -97,6 +100,12 @@ if [[ $? -eq 0 ]]; then
   echo "p7zip module found! No need to install p7zip again, proceeding with installation of compat libraries"
 else
   # install p7zip in host_injections
+  export EASYBUILD_IGNORE_OSDEPS=1
+  export EASYBUILD_SYSROOT=${EPREFIX}
+  export EASYBUILD_RPATH=1
+  export EASYBUILD_FILTER_ENV_VARS=LD_LIBRARY_PATH
+  export EASYBUILD_FILTER_DEPS=Autoconf,Automake,Autotools,binutils,bzip2,cURL,DBus,flex,gettext,gperf,help2man,intltool,libreadline,libtool,Lua,M4,makeinfo,ncurses,util-linux,XZ,zlib
+  export EASYBUILD_MODULE_EXTENSIONS=1
   module load EasyBuild
   eb --robot --installpath=${cuda_install_dir}/ p7zip-${install_p7zip_version}.eb
   ret=$?
```

ocaisa:
The discussion today at the [EESSI Community Meeting](https://eessi.github.io/docs/meetings/2022-09-amsterdam/), led to the following design suggestion/comments:

- According to the CUDA EULA we can distribute the CUDA runtime. It looks like this maps more or less to the `lib` directory under the CUDA installation, even if not the [EULA essentially gives a whitelist](https://docs.nvidia.com/cuda/eula/index.html#attachment-a).
- We can create post install hook for the CUDA installation that replaces anything not in the whitelist with a symlink to the equivalent location under `host_injections`. When the `host_injections` path has a CUDA install, CUDA is fully capable, otherwise it is just the runtime (i.e., we only _ship_ the runtime).
- CUDA will be downgraded to a build dependency (or simply excluded from the final module file). Our use of rpath means software built with CUDA will still work (since they will link to the runtime, which we will ship)
- Lmod will need to be smart on two fronts. It will need a hook that looks at the Lmod `gpu` property and check for the compat library before allowing the module to load (with a helpful error if it needs to fail). For the CUDA module itself, it will also need to refuse to load it unless the symlinks inside have been resolved.
- For the verification that CUDA is actually working, we will need to _ship_ a compiled `deviceQuery` (or whatever), since there is no guarantee the `nvcc` compiler will be available.

The benefit of this approach is that we only install the CUDA SDK if the user actually wants it. It will greatly speed up the GPU support script since there will be need for any `eb` installations at all (by default).

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,75 @@
...
+  fi
+fi
+
+###############################################################################################
+# Install CUDA
+cuda_install_dir="${EESSI_SOFTWARE_PATH/versions/host_injections}"
+mkdir -p ${cuda_install_dir}
+if [ "${install_cuda}" != false ]; then
+  bash $(dirname "$BASH_SOURCE")/cuda_utils/install_cuda.sh ${install_cuda_version} ${cuda_install_dir}
+fi
```
ocaisa, line 65: Let's break this into separate script (and PR) since it will be needed by https://github.com/EESSI/software-layer/pull/212

You also need to check the exit code on the creation of `cuda_install_dir` since this may fail

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,75 @@
...
+###############################################################################################
+# Prepare installation of CUDA compat libraries, i.e. install p7zip if it is missing
```
ocaisa, line 67: You can drop stuff related to p7zip because of #212 (and that also means we can drop `prepare_cuda_compatlibs.sh` entirely)

gpu_support/add_nvidia_gpu_support.sh
```diff
@@ -0,0 +1,75 @@
...
+$(dirname "$BASH_SOURCE")/cuda_utils/prepare_cuda_compatlibs.sh ${install_p7zip_version} ${cuda_install_dir}
+###############################################################################################
+# Try installing five different versions of CUDA compat libraries until the test works.
+# Otherwise, give up
+bash $(dirname "$BASH_SOURCE")/cuda_utils/install_cuda_compatlibs_loop.sh ${cuda_install_dir} ${install_cuda_version}
+
+cuda_version_file="/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/version.txt"
```
ocaisa, line 74:
```suggestion
cuda_version_file="/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/eessi_compat_version.txt"
```
I also think that this creation should be part of `install_cuda_compatlibs_loop.sh` and we should put the supported CUDA version in there according to the compat libs, not the version we need (will help us to avoid unnecessary updates in the future).

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+install_cuda_version=$1
+cuda_install_dir=$2
+
+# TODO: Can we do a trimmed install?
```
ocaisa, line 6: This is done now via your hook

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+install_cuda_version=$1
+cuda_install_dir=$2
```
ocaisa, line 4: General CUDA installation is done via #212 now so I don't think you need this argument. This script is only about installing the CUDA package under `host_injections`...but changing the name of the script to reflect that is probably a good idea.

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# TODO: Can we do a trimmed install?
+# Only install CUDA if specified version is not found.
+# This is only relevant for users, the shipped CUDA installation will
+# always be in versions instead of host_injections and have symlinks pointing
+# to host_injections for everything we're not allowed to ship
+if [ -f ${cuda_install_dir}/software/CUDA/${install_cuda_version}/EULA.txt ]; then
```
ocaisa, line 11: The if/else is still good, except we should be checking under the `host_injections` path. This will allow us to skip any check on available space etc.

ocaisa: You should construct `cuda_install_dir` rather than take it as an argument

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
...
+  echo "CUDA software found! No need to install CUDA again, proceeding with tests"
+else
+  # - as an installation location just use $EESSI_SOFTWARE_PATH but replacing `versions` with `host_injections`
+  #   (CUDA is a binary installation so no need to worry too much about this)
+  # The install is pretty fat, you need lots of space for download/unpack/install (~3*5GB), need to do a space check before we proceed
+  avail_space=$(df --output=avail ${cuda_install_dir}/ | tail -n 1 | awk '{print $1}')
+  if (( ${avail_space} < 16000000 )); then
+    echo "Need more disk space to install CUDA, exiting now..."
+    exit 1
```
ocaisa, line 20: This is a tricky one, we need space for sources, space for the build, space for the install but people can choose where to put all these. I guess we leave it as is for now, but allow people to set an envvar to override this check (and tell them that envvar in the error message)

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
...
+  fi
+  # install cuda in host_injections
+  module load EasyBuild
+  # we need the --rebuild option and a random dir for the module if the module file is shipped with EESSI
+  if [ -f ${EESSI_SOFTWARE_PATH}/modules/all/CUDA/${install_cuda_version}.lua ]; then
```
ocaisa, line 25: If this script is standalone, we'll need to guarantee `EESSI_SOFTWARE_PATH` is defined.

gpu_support/cuda_utils/install_cuda.sh
```diff
@@ -0,0 +1,39 @@
...
+    tmpdir=$(mktemp -d)
+    extra_args="--rebuild --installpath-modules=${tmpdir}"
+  fi
+  eb ${extra_args} --installpath=${cuda_install_dir}/ CUDA-${install_cuda_version}.eb
+  ret=$?
```
ocaisa, line 30: Let's import the bash functions defined in `utils.sh` and use them throughout (where appropriate)

ocaisa:
@huebner-m Let's branch out a separate PR for the script to install CUDA under `host_injections`

gpu_support/cuda_utils/install_cuda.sh
ocaisa: Also, we should allow for a forced installation to override this check

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+if [[ $driver_cuda_version =~ ^[0-9]+$ ]]; then
+  if [ "$driver_cuda_version" -gt "$eessi_cuda_version" ]; then  echo "You need to update your CUDA compatability libraries"; fi
+fi
+
+# If not, grab the latest compat library RPM or deb
+# Download and unpack in temporary directory, easier cleanup after installation
+tmpdir=$(mktemp -d)
+cd $tmpdir
+compat_file=${libs_url##*/}
+wget ${libs_url}
```
ocaisa, line 32: Is `wget` guaranteed by the compat layer? We may need to switch to `curl`

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Unpack it
+# rpm files are the default for all OSes
+# Keep support for deb files in case it is needed in the future
+file_extension=${compat_file##*.}
+if [[ ${file_extension} == "rpm" ]]; then
+  # p7zip is installed under host_injections for now, make that known to the environment
+  if [ -d ${cuda_install_dir}/modules/all ]; then
+    module use ${cuda_install_dir}/modules/all/
+  fi
```
ocaisa, line 43: You can drop this

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+  # Load p7zip to extract files from rpm file
+  module load p7zip
```
ocaisa, line 45: The assumption here is that the EESSI environment has already been sourced

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
+
+libs_url=$1
+cuda_install_dir=$2
+
+current_dir=$(dirname $(realpath $0))
+host_injections_dir="/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia"
+host_injection_linker_dir=${EESSI_EPREFIX/versions/host_injections}
```
ocaisa, line 8: The assumption here is that the EESSI environment has already been sourced

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+  echo "${host_injection_linker_dir}/lib should point to ${host_injections_dir}/latest/compat"
+  exit 1
+fi
+
+# return to initial dir
+cd $current_dir
+
+echo
+echo CUDA driver compatability drivers installed for CUDA version:
+echo ${cuda_dir/cuda-/}
```
ocaisa, line 92: I'd drop the CUDA version supported into `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/eessi_compat_version.txt.tmp
` in a way that makes it easy to compare to the target CUDA version

If we verify the installation in `install_cuda_compatlibs_loop.sh`, then we can remove the `.tmp` extension

ocaisa: We could also use the `.tmp` files to check whether we need to bother installing a certain compat lib or not.

gpu_support/cuda_utils/install_cuda_compatlibs_loop.sh
```diff
@@ -0,0 +1,102 @@
...
+if [ $ret -ne 0 ]; then
+  echo "Couldn't find current URLs of the CUDA compat libraries, instead got:"
+  echo $cuda_compat_urls
+  exit 1
+fi
+
+# loop over the compat library versions until we get one that works for us
+keep_driver_check=1
+# Do a maximum of five attempts
+for value in {1..5}
```
ocaisa, line 21: We can do better here. We should probably keep going until we hit a compat version that won't satisfy our `install_cuda_version` requirement. Using `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/eessi_compat_version.txt.tmp` and `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/eessi_compat_version.txt` should be able to help us here.

gpu_support/cuda_utils/install_cuda_compatlibs_loop.sh
```diff
@@ -0,0 +1,102 @@
...
+      install_compat_libs=true
+    else
+      echo "CUDA compat libs are up-to-date, skip installation."
+    fi
+
+    if [ "${install_compat_libs}" == true ]; then
+      bash $(dirname "$BASH_SOURCE")/install_cuda_compatlibs.sh ${latest_cuda_compat_url} ${cuda_install_dir}
+    fi
+
+    if [[ "${install_wo_gpu}" != "true" ]]; then
```
ocaisa, line 62: Should this be an upper case envvar? And we can check for a GPU already  via the existence/running of `nvidia-smi`, so if this is not set but that fails we can already exit and tell them what they need to do

gpu_support/cuda_utils/install_cuda_compatlibs_loop.sh
```diff
@@ -0,0 +1,102 @@
...
+      bash $(dirname "$BASH_SOURCE")/test_cuda.sh "${install_cuda_version}"
+      if [ $? -eq 0 ]
+      then
+        exit 0
+      else
+        echo
+        echo "It looks like your driver is not recent enough to work with that release of CUDA, consider updating!"
+        echo "I'll try an older release to see if that will work..."
+        echo
```
ocaisa, line 71: Start using `utils.sh` here

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+  # p7zip is installed under host_injections for now, make that known to the environment
+  if [ -d ${cuda_install_dir}/modules/all ]; then
+    module use ${cuda_install_dir}/modules/all/
+  fi
+  # Load p7zip to extract files from rpm file
+  module load p7zip
+  # Extract .cpio
+  7z x ${compat_file}
+  # Extract lib*
+  7z x ${compat_file/rpm/cpio}
```
ocaisa, line 49: We should pipe the output of these to `/dev/null` (or perhaps adopt the behaviour of `EESSI_SILENT`). We should also be checking the exit codes (there's some bash function out there for that).

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
...
+# Add a symlink that points the latest version to the version we just installed
+ln -sfn ${cuda_dir} latest
+
+if [ ! -e latest ] ; then
+  echo "Symlink to latest cuda compat lib version is broken, exiting now..."
+  exit 1
+fi
+
+# Create the space to host the libraries
+mkdir -p ${host_injection_linker_dir}
```
ocaisa, line 77: We should always check exit codes on our commands, should like a function that does that for us is needed

gpu_support/cuda_utils/install_cuda_compatlibs_loop.sh
```diff
@@ -0,0 +1,102 @@
...
+      fi
+    else
+      echo "Requested to install CUDA without GPUs present, so we skip final tests."
+      echo "Instead we test if module load CUDA works as expected..."
+      if [ -d ${cuda_install_dir}/modules/all ]; then
+        module use ${cuda_install_dir}/modules/all/
+      else
+        echo "Cannot load CUDA, modules path does not exist, exiting now..."
+        exit 1
+      fi
```
ocaisa, line 81: Drop this, no need for the module use, CUDA-Samples is shipped with EESSI. Our Lmod hook should cause the load of (a specific version of) `CUDA-Samples` (not `CUDA` since we only deal with compat libs here) to fail unless the compat libs are in place (i.e. Lmod should ensure the existence of `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/eessi_compat_version.txt`)

gpu_support/cuda_utils/install_cuda_compatlibs_loop.sh
```diff
@@ -0,0 +1,102 @@
...
+      module load CUDA
+      ret=$?
+      if [ $ret -ne 0 ]; then
+        echo "Could not load CUDA even though modules path exists..."
+        exit 1
+      else
+        echo "Successfully loaded CUDA, you are good to go! :)"
+        echo "  - To build CUDA enabled modules use ${EESSI_SOFTWARE_PATH/versions/host_injections} as your EasyBuild prefix"
```
ocaisa, line 89: This is not required, then can build where they like (but it is a very sensible location!)

gpu_support/cuda_utils/prepare_cuda_compatlibs.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
ocaisa, line 1: This script can be dropped

gpu_support/cuda_utils/test_cuda.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
ocaisa, line 1: This script can be greatly simplified, just load `CUDA-Samples` and see does `deviceQuery` run

ocaisa: We can expand the testing from there.

ocaisa:
@huebner-m Break the two compat libs scripts off into another PR, the testing script into another, the lmod hook into another and the docs into another (we can finalise those once the others are ready)

gpu_support/cuda_utils/install_cuda_compatlibs.sh
ocaisa: Yes it is (right now):
```
[ocaisa@gpunode1 ~]$ which wget
/cvmfs/pilot.eessi-hpc.org/versions/2021.12/compat/linux/x86_64/usr/bin/wget
```

gpu_support/cuda_utils/install_cuda_compatlibs.sh
```diff
@@ -0,0 +1,92 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ocaisa: It's fine as is, as long as the first thing we do is source the EESSI environment

gpu_support/cuda_utils/install_cuda.sh
ocaisa: It prefer the we ship the EULA text so I think we should check for an expectect broken symlink here:
```
if [[ -L "${cuda_install_dir}/software/CUDA/bin/nvcc" && -e "${cuda_install_dir}/software/CUDA/bin/nvcc" ]]; then
```

ocaisa:

PR #184, opened 2022-09-09
Adjustments to scripts necessary for bot PR24

trz42:
This PR contains a few changes to make the software layer compatible with [EESSI/eessi-bot-software-layer/PR24](https://github.com/EESSI/eessi-bot-software-layer/pull/24).

This can be used as a base branch for adding new software packages to the EESSI pilot via the bot.

- The parameters for EESSI-pilot-install-software.sh and the handling of GENERIC option have been changed. Also a note where more sections for additional packages has been added.
- The README.md has been updated (newer pilot version).
- The 'exit 2' when `/tmp` directory does not support extended attributes or the command 'attr' was not found has been commented out.
- In 'install_software_layer.sh' the parameter "$@" was added to running the script "EESSI-pilot-install-software.sh".

### DIFF
--- a/build_container.sh
+++ b/build_container.sh
@@ -26,7 +26,7 @@ if [ $? -eq 0 ]; then
     attr -s test -V test $testfile > /dev/null
     if [ $? -ne 0 ]; then
         echo "ERROR: $EESSI_TMPDIR does not support extended attributes!" >&2
-       exit 2
+        #exit 2
     else
         rm $testfile
     fi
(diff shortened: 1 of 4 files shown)


build_container.sh
```diff
@@ -26,7 +26,7 @@ if [ $? -eq 0 ]; then
     attr -s test -V test $testfile > /dev/null
     if [ $? -ne 0 ]; then
         echo "ERROR: $EESSI_TMPDIR does not support extended attributes!" >&2
-       exit 2
+        #exit 2
```
boegel, line 29: Rather than commenting this out, we should just drop this `attr` check, should be OK thanks to https://github.com/EESSI/filesystem-layer/pull/125

I'll do that in a follow-up PR

boegel (approved):

build_container.sh
boegel: see #189

PR #186, opened 2022-09-15
Feature archdetect

hmeiland:
alternative for archspec:
- removes requirement for python
- provides more flexibility on which cpu features are -really- relevant

not enabled by default, requires env variable EESSI_USE_ARCHDETECT to be set to use.


boegel:
PR replaced by #187

PR #187, opened 2022-09-15
add support in init script for using 'archdetect' alternative to archspec (only if $EESSI_USE_ARCHDETECT is set to '1')

hmeiland:
alternative for archspec
- removes requirement for python
- be able to maintain which flags are important for architecture determination
- provides detection for all EESSI supported cpu archs 
- breaks graviton2 detection in favor of neoverse-n1
- breaks graviton3 detection in favor of neoverse-v1
- 
not enabled by default: require EESSI_USE_ARCHDETECT to be used (for now)


init/eessi_archdetect.sh
```diff
@@ -0,0 +1,74 @@
...
+# x86_64/generic
+# x86_64/intel/haswell
+# x86_64/intel/skylake_avx512
+# x86_64/amd/zen2
+# x86_64/amd/zen3
+# aarch64/generic
+# ppc64le/generic
+# ppc64le/power9le
+
+ARGUMENT=${1:-none}
```
boegel, line 13: Avoid hardcoding `/proc/cpuinfo` path and `uname -m` in script, so we can unit test it easily in CI:

```
PROC_CPUINFO=${EESSI_PROC_CPUINFO:-/proc/cpuinfo}
MACHINE_TYPE=${EESSI_MACHINE_TYPE:-$(uname -m)}
```

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,74 @@
...
+  fi
+
+  if [ ${MACHINE_TYPE} == "x86_64" ]; then
+    # check for vendor info, if available, for x86_64
+    CPUINFO_VENDOR_FLAG=$(grep -m 1 ^vendor_id /proc/cpuinfo)
+    [[ $CPUINFO_VENDOR_FLAG =~ .*GenuineIntel* ]] && CPU_VENDOR=intel
+    [[ $CPU_VENDOR_FLAG =~ .*AuthenticAMD* ]] && CPU_VENDOR=amd
+
+    CPU_FLAGS=$(grep -m 1 ^flags /proc/cpuinfo)
+    [[ $CPU_FLAGS =~ .*avx2* ]] && HAS_AVX2=true
```
boegel, line 40: We need to protect ourselves from not picking up `$HAS_AVX2` that is set outside of this script, by declaring these local?

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,74 @@
...
+    [[ $CPU_FLAGS =~ .*avx512f* ]] && HAS_AVX512F=true
+    [[ $CPU_FLAGS =~ .*avx512vl* ]] && HAS_AVX512VL=true
+    [[ $CPU_FLAGS =~ .*avx512ifma* ]] && HAS_AVX512IFMA=true
+    [[ $CPU_FLAGS =~ .*avx512_vbmi2* ]] && HAS_AVX512_VBMI2=true
+    [[ $CPU_FLAGS =~ .*avx512_vnni* ]] && HAS_AVX512_VNNI=true
+    [[ $CPU_FLAGS =~ .*avx512fp16* ]] && HAS_AVX512FP16=true
+    [[ $CPU_FLAGS =~ .*vaes* ]] && HAS_VAES=true
+
+    [[ ${CPU_VENDOR} == "intel" ]] && [[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=haswell 
+    [[ ${CPU_VENDOR} == "intel" ]] && [[ ${HAS_AVX512F} ]] && CPU_TYPE=skylake_avx512
```
boegel, line 51:
```bash
if [ ${CPU_VENDOR} == "intel" ]; then
  [[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=haswell 
  [[ ${HAS_AVX512F} ]] && CPU_TYPE=skylake_avx512

elif [ ${CPU_VENDOR} == "amd" ]; then
   ...
fi
```

boegel (changes requested):

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+  # fallback path
+  CPU_PATH="${MACHINE_TYPE}/generic"
+
+  if [ ${MACHINE_TYPE} == "aarch64" ]; then
+    echo ${CPU_PATH}
+    exit
+  fi
+
+  if [ ${MACHINE_TYPE} == "ppc64le" ]; then
+    PROC_CPUINFO=${EESSI_PROC_CPUINFO:-/proc/cpuinfo}
```
boegel: This should move up, to only have it once

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+    CPU_FLAGS=$(grep -m 1 -i ^cpu ${PROC_CPUINFO})
+    [[ $CPU_FLAGS =~ .*POWER9* ]] && HAS_POWER9=true
```
boegel: Add quotes, helps with readability?

```shell
[[ $CPU_FLAGS =~ ".*POWER9*" ]] && HAS_POWER9=true

File: init/eessi_archdetect.sh [hunk 6]
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+# current pathways implemented in EESSI
```
boegel, line 3: mention `2021.12 pilot`?

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+# x86_64/generic
+# x86_64/intel/haswell
+# x86_64/intel/skylake_avx512
+# x86_64/amd/zen2
+# x86_64/amd/zen3
+# aarch64/generic
```
boegel, line 9: `graviton2` + `graviton3` is missing here, should also be covered by tests

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+  if [ ${MACHINE_TYPE} == "x86_64" ]; then
+    # check for vendor info, if available, for x86_64
+    PROC_CPUINFO=${EESSI_PROC_CPUINFO:-/proc/cpuinfo}
+    CPUINFO_VENDOR_FLAG=$(grep -m 1 -i ^vendor ${PROC_CPUINFO})
+    [[ $CPUINFO_VENDOR_FLAG =~ .*GenuineIntel* ]] && CPU_VENDOR=intel
+    [[ $CPUINFO_VENDOR_FLAG =~ .*AuthenticAMD* ]] && CPU_VENDOR=amd
+
+    CPU_FLAGS=$(grep -m 1 -i ^flags ${PROC_CPUINFO})
+    [[ $CPU_FLAGS =~ .*avx2* ]] && HAS_AVX2=true
+    [[ $CPU_FLAGS =~ .*fma* ]] && HAS_FMA=true
```
boegel: We need to be a bit more careful here, `fma` could be (in theory) a partial match in another CPU feature, so spaces before/after are needed here (while taking into account that `fma` may be listed last, so no trailing space in that case...)

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+    [[ $CPU_FLAGS =~ .*avx512fp16* ]] && HAS_AVX512FP16=true
+    [[ $CPU_FLAGS =~ .*vaes* ]] && HAS_VAES=true
+
+    if [ ${CPU_VENDOR} == "intel" ]; then
+      [[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=haswell 
+      [[ ${HAS_AVX512F} ]] && CPU_TYPE=skylake_avx512
+      # [[ ${HAS_AVX512IFMA} ]] && [[ ${HAS_AVX512_VBMI2} ]] && CPU_TYPE=icelake_avx512
+      # [[ ${HAS_AVX512_VNNI} ]] && [[ ${HAS_AVX512VL} ]] && [[ ${HAS_AVX512FP16} ]] && CPU_TYPE=sapphire_rapids_avx512
+    elif [ ${CPU_VENDOR} == "amd" ]; then
+      [[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=zen2
```
boegel, line 60: Maybe we should add a comment here to make this a bit easier to find when people look for `rome` or `milan`:

```shell
[[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=zen2  # Rome
[[ ${HAS_AVX2} ]] && [[ ${HAS_FMA} ]] && CPU_TYPE=zen3  # Milan, Milan-X

File: init/eessi_environment_variables [hunk 10]
```diff
@@ -25,9 +25,15 @@ if [ -d $EESSI_PREFIX ]; then
   if [ -d $EESSI_EPREFIX ]; then
 
     # determine subdirectory in software layer
-    # note: eessi_software_subdir_for_host.py will pick up value from $EESSI_SOFTWARE_SUBDIR_OVERRIDE if it's defined!
-    export EESSI_EPREFIX_PYTHON=$EESSI_EPREFIX/usr/bin/python3
-    export EESSI_SOFTWARE_SUBDIR=$($EESSI_EPREFIX_PYTHON ${EESSI_INIT_DIR_PATH}/eessi_software_subdir_for_host.py $EESSI_PREFIX)
+    if [ -z $EESSI_USE_ARCHDETECT ]; then
+      # if archdetect is enabled, use internal code
+      export EESSI_SOFTWARE_SUBDIR=$(${EESSI_INIT_DIR_PATH}/eessi_archdetect.sh cpupath)
+      echo "archdetect says ${EESSI_SOFTWARE_SUBDIR}"
```
boegel, line 31: If we have this `echo`, we should also have it for `archspec` below, so it's always clear which mode is used

boegel (changes requested):

init/eessi_archdetect.sh
hmeiland: done

init/eessi_archdetect.sh
hmeiland: done

init/eessi_archdetect.sh
hmeiland: with changes to wildcard: done

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,109 @@
...
+      [[ $EESSI_CPU_TYPE ]] && CPU_PATH="${MACHINE_TYPE}/${EESSI_CPU_VENDOR}/${EESSI_CPU_TYPE}"
+    else
+      CPU_FLAGS=$(grep -m 1 -i ^features ${PROC_CPUINFO} | sed 's/$/ /g')
+      [[ $CPU_FLAGS =~ " asimd " ]] && EESSI_HAS_ASIMD=true
+      [[ $CPU_FLAGS =~ " svei8mm " ]] && EESSI_HAS_SVEI8MM=true
+
+      [[ ${EESSI_HAS_ASIMD} ]] && EESSI_CPU_TYPE=graviton2
+      [[ ${EESSI_HAS_SVEI8MM} ]] && EESSI_CPU_TYPE=graviton3
+      [[ $EESSI_CPU_TYPE ]] && CPU_PATH="${MACHINE_TYPE}/${EESSI_CPU_TYPE}"
+    fi
```
boegel, line 49: I would never return `graviton2` as a result here, since then we get get `graviton2` as a result while we're actually on a `thunderx2`, for example. We should break here with `archspec`, and symlink `aarch64/arm/neoverse-n1` to `aarch64/graviton2` in EESSI pilot 2021.12, and then the reverse in the next EESSI pilot (so it still works with `archspec`).

```shell
  if [ ${MACHINE_TYPE} == "aarch64" ]; then
    CPUINFO_VENDOR_FLAG=$(grep -m 1 -i ^vendor ${PROC_CPUINFO})
    [[ $CPUINFO_VENDOR_FLAG =~ "ARM" ]] && EESSI_CPU_VENDOR=arm

   if [ ${EESSI_CPU_VENDOR} == "arm" ]; then
      CPU_FLAGS=$(grep -m 1 -i ^flags ${PROC_CPUINFO} | sed 's/$/ /g')
   else
      CPU_FLAGS=$(grep -m 1 -i ^features ${PROC_CPUINFO} | sed 's/$/ /g')
   fi
   [[ $CPU_FLAGS =~ " asimd " ]] && EESSI_HAS_ASIMD=true
   [[ $CPU_FLAGS =~ " svei8mm " ]] && EESSI_HAS_SVEI8MM=true

   [[ ${EESSI_HAS_ASIMD} ]] && EESSI_CPU_TYPE=neoverse-n1
   [[ ${EESSI_HAS_SVEI8MM} ]] && EESSI_CPU_TYPE=neoverse-v1
   [[ $EESSI_CPU_TYPE ]] && CPU_PATH="${MACHINE_TYPE}/${EESSI_CPU_VENDOR}/${EESSI_CPU_TYPE}"
```

boegel (changes requested):

init/eessi_archdetect.sh
```diff
@@ -0,0 +1,81 @@
...
+  fi
+
+  echo "should not see this...something weird going on..."
+  echo "please mail output of lscpu to me..."
+}
+
+if [ ${ARGUMENT} == "none" ]; then
+  echo usage: $0 cpupath
+  exit
+elif [ ${ARGUMENT} == "cpupath" ]; then
```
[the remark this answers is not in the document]
hmeiland: no, currently this is specific for cpupath, hence the slashes and the usage in building the software path
this way we will have future output choices

init/eessi_archdetect.sh
hmeiland: done

init/eessi_archdetect.sh
hmeiland: implemented

init/eessi_environment_variables
hmeiland: done

lexming:
As discussed, I suggest to define all supported CPU architectures in some data structure independent from the script checking the CPU architecture of the host system. This approach should be easier to maintain as the script will _rarely_ need any changes and adding new CPU archs can be done by expanding the list of definitions.

Please check https://gist.github.com/lexming/fd249a4b1a779c6b19a8f5bcb736e170 for a working barebones implementation. If you agree with this suggestion I can prepare a PR to this PR to integrate it.

hmeiland:
> As discussed, I suggest to define all supported CPU architectures in some data structure independent from the script checking the CPU architecture of the host system. This approach should be easier to maintain as the script will _rarely_ need any changes and adding new CPU archs can be done by expanding the list of definitions.
> 
> Please check https://gist.github.com/lexming/fd249a4b1a779c6b19a8f5bcb736e170 for a working proof-of-concept implementation. If you agree with this suggestion I can prepare a PR to this PR to integrate it.

added as arch_detect; tuned for edge cases (eg support both "Vendor ID" and "vendor_id"... switching tests to this code.

lexming:
@hmeiland thanks for the update, I went through the code and we can now easily move the list of specs for each architecture into their own files. In this way the script is totally independent from the spec data.

Please see my PR https://github.com/hmeiland/software-layer/pull/2. If you agree with my changes and merge it, it will update this PR.

lexming:
@hmeiland I've gone through this PR one more time and it looks fine on my side. All I can think of as improvement at this point is to enhance the logging and comments in the script.

I made the following changes in PR https://github.com/hmeiland/software-layer/pull/3:
* add a `log` function to simplify logging, we can now easily use different logging levels
* add `-h` (print help) and `-d` (debug level) arguments to the script
* add more comments and fix formatting
* fix `get_cpuinfo()` to be fully case insensitive
* generalize `check_flags()` to `check_allinfirst()`, the new function works in the same way

You will see that the diff looks like I rewrote most of the script, but that's mainly because I standardize the indentation to 4 spaces.

lexming (approved):
LGTM

init/arch_specs/eessi_arch_x86.spec
```diff
@@ -0,0 +1,6 @@
+# x86_64 CPU architecture specifications
+# Software path in EESSI 	| Vendor ID 	| List of defining CPU features
+"x86_64/intel/haswell"		"GenuineIntel"	"avx2 fma" 		# Intel Haswell, Broadwell
+"x86_64/intel/skylake_avx512"	"GenuineIntel"	"avx2 fma avx512f"	# Intel Skylake, Cascade Lake
```
bartoldeman, line 4: If I read this correctly this could be misidentified as Xeon Phi Knights Landing.
Skylake and Cascade Lake have `avx512bw avx512vl avx512dq`, which Knights Landing doesn't have.

bartoldeman (changes requested):

init/arch_specs/eessi_arch_x86.spec
hmeiland: the conscious idea was to look for the -minimal- required cpu features to have EESSI set of binaries run optimally, in contrast to archspec which tries take the (over?) complete set of features. The current combination of features is just a hopefully sane start, which work for the current set of cpu's in the test scenarios. Please do provide a cpuinfo from KL so we can add it. Also let's test verify that the current codes indeed do not run on KL... Thanks!

bartoldeman: I no longer have access to a KNL node but https://colfaxresearch.com/knl-avx512/ has this:
```
vega@lyra% cat /proc/cpuinfo
...
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cm
ov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdt
scp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aper
fmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 fma cx16 xtp
r pdcm sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdr
and lahf_lm abm 3dnowprefetch ida arat epb xsaveopt pln pts dtherm tpr_shadow vnmi 
flexpriority ept vpid fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms avx512f rdseed 
adx avx512pf avx512er avx512cd
...```

Knights Landing is getting pretty rare so there is no reason to compile with only support for `avx512f`; excluding bw/vl/dq isn't good for performance. It just shouldn't misidentify. Haswell `avx2` code will run fine on KNL, just without optimal performance.

Comment (hmeiland) [reply init/arch_specs/eessi_arch_x86.spec]: skylake_avx512 updated to include '''AVX512F, AVX512BW, AVX512CD, AVX512DQ, AVX512VL'''

**Review by bartoldeman — APPROVED** [review]
lgtm

**Review by lexming — DISMISSED** [review]
LGTM

File: init/eessi_environment_variables [hunk 14]
```diff
@@ -25,9 +25,16 @@ if [ -d $EESSI_PREFIX ]; then
   if [ -d $EESSI_EPREFIX ]; then
 
     # determine subdirectory in software layer
-    # note: eessi_software_subdir_for_host.py will pick up value from $EESSI_SOFTWARE_SUBDIR_OVERRIDE if it's defined!
-    export EESSI_EPREFIX_PYTHON=$EESSI_EPREFIX/usr/bin/python3
-    export EESSI_SOFTWARE_SUBDIR=$($EESSI_EPREFIX_PYTHON ${EESSI_INIT_DIR_PATH}/eessi_software_subdir_for_host.py $EESSI_PREFIX)
+    if [ -z $EESSI_USE_ARCHDETECT ]; then
```
boegel, line 28: @hmeiland The condition is wrong here: currently, if `$EESSI_USE_ARCHDETECT` is set (then `-z` is false), you're using `archspec` rather the `archdetect` as intended...

Let's use a more explicit check, and only enable archdetect when `$EESSI_USE_ARCHDETECT` is set to `1`:

```suggestion
    if [ "$EESSI_USE_ARCHDETECT" == "1" ]; then
```

We should also make the archdetect mode honor `$EESSI_SOFTWARE_SUBDIR_OVERRIDE` btw, but that can be done in a follow-up PR

boegel (changes requested):

boegel (approved):

boegel:
I've tested this in various systems, works as expected:

* default of using `archspec` is unchanged, for now;
* `eessi_archdetect.sh` bash script is used instead of `archspec` only if `$EESSI_USE_ARCHDETECT` is set to `1`;
* on an Azure AMD Milan(-X) system (`HB120rs_v3`), `archdetect` correctly returns `x86_64/amd/zen3`, while the `archspec` version we have in the EESSI pilot version 2021.12 returns `x86-64/amd/zen2` (see also https://github.com/archspec/archspec-json/issues/38):
  * using `archspec` (default):
    ```
    Singularity> source /cvmfs/pilot.eessi-hpc.org/versions/2021.12/init/bash
    Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2021.12!
    archspec says x86_64/amd/zen2
    Using x86_64/amd/zen2 as software subdirectory.
    Using /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/modules/all as the directory to be added to MODULEPATH.
    Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/.lmod/lmodrc.lua
    Initializing Lmod...
    Prepending /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen2/modules/all to $MODULEPATH...
    Environment set up to use EESSI pilot software stack, have fun!
    [EESSI pilot 2021.12] $
    ```
  * using `archdetect` (opt-in via `$EESSI_USE_ARCHDETECT`):
    ```
    Singularity> EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2021.12/init/bash
    Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2021.12!
    2022-11-08 17:17:59 [INFO] cpupath: best match for host CPU: x86_64/amd/zen3
    archdetect says x86_64/amd/zen3
    Using x86_64/amd/zen3 as software subdirectory.
    Using /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen3/modules/all as the directory to be added to MODULEPATH.
    Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen3/.lmod/lmodrc.lua
    Initializing Lmod...
    Prepending /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/x86_64/amd/zen3/modules/all to $MODULEPATH...
    Environment set up to use EESSI pilot software stack, have fun!
    [EESSI pilot 2021.12] $
    ```
* On Graviton 3, `archdetect` returns `aarch64/arm/neoverse-v1` (similar to what the next version of `archspec` will do, cfr. https://github.com/archspec/archspec-json/pull/57), which requires that we symlink that subdir to `aarch64/graviton3`:
   ```
   Singularity> EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2021.12/init/bash
   Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2021.12!
   2022-11-08 16:58:55 [INFO] cpupath: best match for host CPU: aarch64/arm/neoverse-v1
   archdetect says aarch64/arm/neoverse-v1
   Using aarch64/arm/neoverse-v1 as software subdirectory.
   ERROR: EESSI software layer at /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/arm/neoverse-v1 not found!
   ```
   ```
   Singularity> cd /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64
   Singularity> ls
   generic  graviton2  graviton3
   Singularity> mkdir arm
   Singularity> cd arm
   Singularity> ln -s ../graviton3 neoverse-v1
   ```
   ```
   Singularity> EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2021.12/init/bash
   Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2021.12!
   2022-11-08 16:59:39 [INFO] cpupath: best match for host CPU: aarch64/arm/neoverse-v1
   archdetect says aarch64/arm/neoverse-v1
   Using aarch64/arm/neoverse-v1 as software subdirectory.
   Using /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/arm/neoverse-v1/modules/all as the directory to be added to MODULEPATH.
   Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/arm/neoverse-v1/.lmod/lmodrc.lua
   Initializing Lmod...
   Prepending /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/arm/neoverse-v1/modules/all to $MODULEPATH...
   Environment set up to use EESSI pilot software stack, have fun!
   [EESSI pilot 2021.12] $
   ```
   * With the `archspec` version in the 2021.12 EESSI compat layer, we're falling back to `aarch64/graviton2`:
     ```
     Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2021.12!
     WARNING: Ignoring unknown target "graviton3"
     archspec says aarch64/graviton2
     Using aarch64/graviton2 as software subdirectory.
     Using /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/graviton2/modules/all as the directory to be added to MODULEPATH.
     Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/graviton2/.lmod/lmodrc.lua
     Initializing Lmod...
     Prepending /cvmfs/pilot.eessi-hpc.org/versions/2021.12/software/linux/aarch64/graviton2/modules/all to $MODULEPATH...
     Environment set up to use EESSI pilot software stack, have fun!
     [EESSI pilot 2021.12] $
     ```

I've also tested this on:

* on Intel Skylake, `archdetect` returns `x86_64/intel/skylake_avx512`;
* on AMD Rome, `archdetect` returns `x86_64/amd/zen2`;
* on Graviton 2, `archdetect` returns `aarch64/arm/neoverse-n1` (which should symlink to `aarch64/graviton2` in EESSI pilot 2021.12);
* on POWER9, `archdetect` returns `ppc64le/power9le`;

PR #189, opened 2022-10-07
remove check for support for extended attributes from script for using build container

boegel:
The `attr` check isn't needed anymore thanks to https://github.com/EESSI/filesystem-layer/pull/125, and `attr` is often not available anyway...


trz42 (approved):
Looks good. Cleaner than just commenting out the exit!

PR #190, opened 2022-10-15
enhance install script to check exit code when installing EasyBuild

boegel:
We were assuming that the EasyBuild installation always worked, leading to confusing output being produced by the install script...


trz42 (approved):
Looks good.

PR #197, opened 2022-11-08
add snakemake on 2020a

jpecar:
I'll leave this here, will look into 2021a tomorrow and then we can decide which one do we want to ingest


bedroge:
The pilot repository is no longer supproted, so I'm going to close this old PR. Snakemake is available in the production repository (https://github.com/EESSI/software-layer/pull/474).

PR #203, opened 2022-11-12
enhance EasyBuild hooks to fix installation of MetaBAT in EESSI

boegel:
This is required for #202, in conjunction with https://github.com/easybuilders/easybuild-easyconfigs/pull/16624 (but the latter doesn't have to be merged strictly speaking)


ocaisa (approved):
LGTM

ocaisa:
Ah, you requested my review, so that's a yes. Merging...

PR #212, opened 2022-12-16
Add CUDA support to software_layer

ocaisa:
This is part of a logical splitting of #172 to make it a bit more manageable

~~Requires #228~~


EESSI-pilot-install-software.sh
```diff
@@ -394,6 +394,34 @@ $EB --from-pr 15885 OpenBLAS-0.3.15-GCC-10.3.0.eb --robot
...
+# install CUDA (uses eb_hooks.py to only install runtime)
+cuda_ec="CUDA-11.3.1.eb"
+echo ">> Installing $cuda_ec..."
+ok_msg="$cuda_ec installed, off to a good (?) start!"
+fail_msg="Failed to install $cuda_ec, woopsie..."
+$EB $cuda_ec --robot
+check_exit_code $? "${ok_msg}" "${fail_msg}"
+
+# install CUDA samples (requires EESSI support for CUDA)
+# TODO Run EESSI NVIDIA GPU support script here
```
ocaisa, line 416: We are going to need a dependent PR here, that performs the CUDA installation. We have a bit of a chicken-egg situation, as we need the installation to install the CUDA-Samples but we require the CUDA-Samples to check that things are working. This implies it'll have to be a forced installation without a verification check (which is not necessarily bad since it means we don't need a GPU).

eb_hooks.py
```diff
@@ -171,24 +220,80 @@ def wrf_preconfigure(self, *args, **kwargs):
...
+    if (
+        "CUDA" in [dep[0] for dep in iter(ec_dict["dependencies"])]
+        or ec_dict["toolchain"]["name"] in CUDA_ENABLED_TOOLCHAINS
+    ):
+        ec.log.info("[parse hook] Injecting gpu as Lmod arch property and envvar with CUDA version")
+        key = "modluafooter"
+        value = 'add_property("arch","gpu")'
+        cuda_version = 0
+        for dep in iter(ec_dict["dependencies"]):
+            # Make CUDA a build dependency only (rpathing saves us from link errors)
```
ocaisa, line 286: This approach saves us from explicitly loading CUDA to run a CUDA dependent package. This allows us to write an Lmod  hook that protects loading the CUDA module unless certain criteria are met (i.e., that the symlinks are unbroken).

eb_hooks.py
```diff
@@ -171,24 +220,80 @@ def wrf_preconfigure(self, *args, **kwargs):
...
+def inject_gpu_property(ec):
+    ec_dict = ec.asdict()
+    # Check if CUDA is in the dependencies, if so add the GPU Lmod tag
+    if (
+        "CUDA" in [dep[0] for dep in iter(ec_dict["dependencies"])]
+        or ec_dict["toolchain"]["name"] in CUDA_ENABLED_TOOLCHAINS
+    ):
+        ec.log.info("[parse hook] Injecting gpu as Lmod arch property and envvar with CUDA version")
+        key = "modluafooter"
+        value = 'add_property("arch","gpu")'
```
ocaisa, line 283: This property allows us to protect the loading of any GPU package via an Lmod hook (which can be overridden): unless the compat libraries are installed you can't load GPU modules

eb_hooks.py
```diff
@@ -171,24 +204,96 @@ def wrf_preconfigure(self, *args, **kwargs):
...
+            if not os.path.islink(os.path.join(root, filename)):
+                # check if the current file is part of the whitelist
+                basename = filename.split(".")[0]
+                if basename not in whitelist:
+                    # if it is not in the whitelist, delete the file and create a symlink to host_injections
+                    source = os.path.join(root, filename)
+                    target = source.replace("versions", "host_injections")
+                    os.remove(source)
+                    # Using os.symlink requires the existence of the target directory, so we use os.system
+                    os.system("ln %s %s" % (target, source))
```
ocaisa, line 255: We should be checking that these are succeeding

eb_hooks.py
```diff
@@ -171,24 +204,96 @@ def wrf_preconfigure(self, *args, **kwargs):
...
+            elif copy:
+                tmp_buffer.append(line)
+    # create whitelist without file extensions, they're not really needed and they only complicate things
+    whitelist = []
+    file_extensions = [".so", ".a", ".h", ".bc"]
+    for tmp in tmp_buffer:
+        for word in tmp.split():
+            if any(ext in word for ext in file_extensions):
+                whitelist.append(word.split(".")[0])
+    whitelist = list(set(whitelist))
```
ocaisa, line 241: Should do some sanity checking list to make sure at least some expected files exist

eb_hooks.py
```diff
@@ -171,24 +204,96 @@ def wrf_preconfigure(self, *args, **kwargs):
...
+            if line.strip() == "2.6. Attachment A":
+                copy = True
+                continue
+            elif line.strip() == "2.7. Attachment B":
+                copy = False
+                continue
+            elif copy:
+                tmp_buffer.append(line)
+    # create whitelist without file extensions, they're not really needed and they only complicate things
+    whitelist = []
```
ocaisa, line 235: we should always include `eula.txt`

ocaisa:
This is working but requires an additional script to unbreak the symlinks in the CUDA installation (which is being extracted from #172 )

ocaisa:
This requires https://github.com/easybuilders/easybuild-framework/pull/4119 _or_ you install the compat libraries so that the rpath check passes

install_cuda_host_injections.sh
```diff
@@ -0,0 +1,53 @@
...
+# This is only relevant for users, the shipped CUDA installation will
+# always be in versions instead of host_injections and have symlinks pointing
+# to host_injections for everything we're not allowed to ship
+# (existence of easybuild subdir implies a successful install)
+if [ -d ${cuda_install_dir}/software/CUDA/${install_cuda_version}/easybuild ]; then
+  echo "CUDA software found! No need to install CUDA again, proceed with testing."
+else
+  # The install is pretty fat, you need lots of space for download/unpack/install (~3*5GB), need to do a space check before we proceed
+  avail_space=$(df --output=avail ${cuda_install_dir}/ | tail -n 1 | awk '{print $1}')
+  if (( ${avail_space} < 16000000 )); then
```
ocaisa, line 28: Should allow this to be overridden

install_cuda_host_injections.sh
```diff
@@ -0,0 +1,53 @@
...
+  # (CUDA is a binary installation so no need to worry too much about the EasyBuild setup)
+  cuda_install_dir=${EESSI_SOFTWARE_PATH/versions/host_injections}
+fi
+
+# Only install CUDA if specified version is not found.
+# This is only relevant for users, the shipped CUDA installation will
+# always be in versions instead of host_injections and have symlinks pointing
+# to host_injections for everything we're not allowed to ship
+# (existence of easybuild subdir implies a successful install)
+if [ -d ${cuda_install_dir}/software/CUDA/${install_cuda_version}/easybuild ]; then
```
ocaisa, line 23: This dir also needs to be writable.

install_cuda_host_injections.sh
```diff
@@ -0,0 +1,53 @@
...
+  echo "CUDA software found! No need to install CUDA again, proceed with testing."
+else
+  # The install is pretty fat, you need lots of space for download/unpack/install (~3*5GB), need to do a space check before we proceed
+  avail_space=$(df --output=avail ${cuda_install_dir}/ | tail -n 1 | awk '{print $1}')
```
ocaisa, line 27: This assumes that `${cuda_install_dir}` exists, which it may not

boegel:
@ocaisa Conflicts to fix.

Shall we re-target this to the (new) `2023.04` branch?

ocaisa:

PR #213, opened 2022-12-20
move utils.sh to scripts/utils.sh + also update script to create tarballs to also cover scripts/ dir

huebner-m:
This is a suggestion for a new directory to be shipped with EESSI. We could use this directory to ship the scripts needed to add NVIDIA GPU support, see e.g. https://github.com/EESSI/software-layer/pull/212


ocaisa (dismissed):
This looks good to me, but would prefer if @boegel took a look to double-check

bedroge (dismissed):
Looks good to me too, we can also merge https://github.com/EESSI/filesystem-layer/pull/139 after this PR has been merged.

ocaisa:
@huebner-m Why did no CI run on this PR? This should have triggered something (tarballing the init/scripts is pretty easy and indirectly checks utils.sh)

boegel (changes requested):

boegel:
The `create_init_tarball.sh` script was not tested yet in CI, but it could/should have, so we should make sure the renamed `create_directory_tarballs.sh` is tested as well by `.github/workflows/tests_scripts.yml`?

.github/workflows/tests_scripts.yml
```diff
@@ -6,8 +6,9 @@ on:
       - build_container.sh
       - install_software_layer.sh
       - run_in_compat_layer_env.sh
-      - utils.sh
+      - scripts/utils.sh
       - update_lmod_cache.sh
+      - create_directory_tarballs.sh
```
boegel, line 11: This only makes sure that this workflow is *triggered* when changes are made to `create_directory_tarballs.sh` - we're not actually testing `create_directory_tarballs.sh` directly in this workflow... Could/should we?

huebner-m: The test should work now (it did locally). However, the CI will still fail until this PR is merged because the script downloads a tarball of the `main` branch to build the final tarballs and the new subdirectory doesn't exist there yet.

boegel (approved):
lgtm

PR #215, opened 2022-12-30
use separate script to determine EESSI pilot version and CVMFS repo

### LINKED ISSUE #207: Single means to define and to override repository and version
The repository (name) is currently set in

- `build_container.sh`
- `create_tarball.sh`
- `run_in_compat_layer_env.sh`
- (`README.md`)

The (default) version is currently set in

- `install_software_layer.sh`
- ~`create_init_tarball.sh`~ (only uses the version provided as command line argument and via EESSI_PILOT_VERSION)
- `init/minimal_eessi_env`
- (`README.md`)

For the bot it would be good if there would be just one single means (e.g., file) which sets the repository name and the default version. Also, there should be a single means to override these values.

This would allow the bot to easily build software stacks for different repositories and versions while still using the same commit of the software-layer repository.

trz42:
This PR fixes #207

- adds a new configuration file `init/eessi_defaults` which sets `EESSI_PILOT_VERSION` to a default or `$EESSI_PILOT_VERSION_OVERRIDE` and `EESSI_CVMFS_REPO` to a default or `$EESSI_CVMFS_REPO_OVERRIDE`
- existing init files and scripts that used hardcoded values are adjusted to `source` the new configuration file
- a comment in `EESSI-pilot-install-software.sh` is updated to note that not a hardcoded pilot version but the one set in `init/eessi_defaults` is used
- a simple test is added which checks if the default values in `init/eessi_defaults` correspond to values used in `README.md`

### DIFF
--- a/run_in_compat_layer_env.sh
+++ b/run_in_compat_layer_env.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
+
+base_dir=$(dirname $(realpath $0))
+source ${base_dir}/init/eessi_defaults
+
+BUILD_CONTAINER="docker://ghcr.io/eessi/build-node:debian10"
 if [ -z $EESSI_PILOT_VERSION ]; then
     echo "ERROR: \$EESSI_PILOT_VERSION must be set!" >&2
     exit 1
 fi
-EESSI_COMPAT_LAYER_DIR="/cvmfs/pilot.eessi-hpc.org/versions/${EESSI_PILOT_VERSION}/compat/linux/$(uname -m)"
+EESSI_COMPAT_LAYER_DIR="${EESSI_CVMFS_REPO}/versions/${EESSI_PILOT_VERSION}/compat/linux/$(uname -m)"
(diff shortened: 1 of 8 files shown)


run_in_compat_layer_env.sh
```diff
@@ -1,9 +1,14 @@
 #!/bin/bash
+
+base_dir=$(dirname $(realpath $0))
+source ${base_dir}/init/eessi_defaults
+
+BUILD_CONTAINER="docker://ghcr.io/eessi/build-node:debian10"
```
boegel, line 6: This is not used at all, please remove (probably a copy-paste error from `build_container.sh`

boegel (changes requested):

run_in_compat_layer_env.sh
boegel: fixed in 4d5cbce

boegel (approved):
lgtm

Issue #214, opened 2022-12-29
single script and container which serve both use and build scenarios


trz42:
Accessing the EESSI pilot software stack via a container differs when the software is just used (read-only access) or when software is built to be added (read & write access). 

For building the container available from `docker://ghcr.io/eessi/build-node:debian10` is used. For read only access the container available from `docker://ghcr.io/eessi/client-pilot:centos7` is used.

For just using the EESSI pilot (read-only access), the container used for building should be usable too. In that case one could provide a single script that serves both scenarios - either read-only access or read & write access. Similar to the `build_container.sh` script it could execute another script (or command) or provide an interactive shell.

Example runs of the script could be:

- `eessi_container.sh --access RO --mode SHELL`
- `eessi_container.sh --access RO --mode EXEC some_workload.sh`
- `eessi_container.sh --access RW --mode SHELL`
- `eessi_container.sh --access RW --mode EXEC install_package.sh`

Default settings could be `--access RO` and `--mode SHELL`.

Each invocation might have additional parameters such as `--new-tmp TMP_SPACE` or `--previous-tmp TMP_SPACE` which would enable a fresh run (`--new-tmp`) or reusing results from a previous run (`--previous-tmp`, particularly useful for incrementally building or debugging sessions).

The script could also enable to access several or specific CernVM-FS repositories, for example, adding parameters `--repository REPO` (e.g., `dev.eessi.io`), `--repository-version VERSION` (e.g., `2023.01`) and `--repository-config CFG_FILE` (e.g., providing all files necessary to access the repository - usually a public key, a `default.local` and a `<repo-domain>.conf`). Thus, any repository could be easily accessed. The implementation of that would be simplified by providing a solution to #207

boegel:
@trz42 This can be closed now that #216 is merged?

trz42:
Closing as suggested.

PR #216, opened 2023-01-04
Unified script to access EESSI (read-only and read & write)

trz42:
Addresses #214 -- provides a unified script to access the EESSI pilot (and any other EESSI-like repository or even the Compute Canada stack) via a container.

To access the EESSI pilot, simply run
```bash
$ ./eessi_container.sh
```
Default repository is set to the EESSI pilot (or more precisely to what is defined via `init/eessi_defaults`, see #215). Default access mode is `ro` (read-only). Default execution mode is `shell`.

To access the EESSI pilot with a writable overlay, simply run
```bash
$ ./eessi_container.sh --access rw
```

TODO (might go to a follow-up PR)
- [ ] describe means to access any repository (argument `-r | --repository`)
- [ ] implement missing features/options


eessi_container.sh
```diff
@@ -0,0 +1,458 @@
...
+      ;;
+    -g|--storage)
+      STORAGE="$2"
+      shift 2
+      ;;
+    -h|--help)
+      display_help
+      exit 0
+      ;;
+    -i|--info)
```
boegel, line 139: maybe rename to `--verbose`?

boegel (changes requested):

eessi_container.sh
boegel: fixed in [09feff7](https://github.com/EESSI/software-layer/pull/216/commits/09feff74d11d4cf14393cde33abd0672eb9f3bcf)

boegel (changes requested):

boegel (approved):

PR #219, opened 2023-01-14
use newer build container (Debian 11 w/ fuse-overlayfs 1.9)

boegel:
cfr. https://github.com/EESSI/filesystem-layer/pull/133


trz42 (approved):
Looks good. Derived RPM name can be downloaded. Installation not tested, but should be ok.

PR #220, opened 2023-01-14
update script to take into account $EESSI_SOFTWARE_SUBDIR_OVERRIDE if it's defined

boegel:
required to fix https://github.com/EESSI/eessi-bot-software-layer/issues/140


trz42 (approved):
Looks good to me. Interesting ways to pass env vars into containers and into a prefix environment.

PR #221, opened 2023-01-15
always exclude '.wh.*' files when creating tarball in overlay

boegel:
Fix for errors that pop up when creating the tarball because "whiteout" files are accidentally picked up, like:

```
tar: 2021.12/software/linux/x86_64/amd/zen2/modules/all/CaDiCaL/.wh..wh..opq: Cannot stat: No such file or directory
```

(starting seeing this in some cases when using updated build container, cfr. #219)

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -43,13 +43,13 @@ if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/.lmod ]; then
 fi
 if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules ]; then
     # module files
-    find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type f >> ${files_list}
+    find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type f | grep -v '/\.wh\.' >> ${files_list}
     # module symlinks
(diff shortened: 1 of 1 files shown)


create_tarball.sh
```diff
@@ -43,13 +43,13 @@ if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/.lmod ]; then
 fi
 if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules ]; then
     # module files
-    find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type f >> ${files_list}
+    find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type f | grep -v '/\.wh\.' >> ${files_list}
```
trz42, line 46: Wonder if it would make sense to collect a list of files which have been filtered out and add that to the metadata when uploading the tarball? This would allow the human overseeing the ingestion procedure to double-check if anything had been removed that shouldn't.

trz42 (approved):

create_tarball.sh
boegel: These `.wh*` files are temporary files created by `fuse-overlayfs`. Although we see them when compiling the list of files to include in the tarball, they're no longer there by the time we create the tarball.
Also, they're useless to us anyway...

PR #223, opened 2023-01-18
Feature archdetect rpi

hmeiland:
add raspberry pi support by adding feature to use CPU part for aarch64


hmeiland:
checks failed sue to HTTP rate limit, not code error...

ocaisa:
@hmeiland Can you sync this with upstream, you still have some older archspec CI

ocaisa:
@hmeiland You are targeting the wrong branch with this, it should be [2023.06-software.eessi.io](https://github.com/EESSI/software-layer/tree/2023.06-software.eessi.io)

ocaisa:
See https://github.com/EESSI/software-layer/actions/runs/8660373452/job/23748193970?pr=223 , correct output should be
as per https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/tests/archdetect/aarch64/neoverse_v1/AWS-awslinux-graviton3.all.output

@hmeiland Unfortunately, this is not working as intended, `archdetect` is no longer recognising `neoverse_n1` as an alternative to `neoverse_v1` before choosing `generic`

ocaisa:
Alternative to this is merged in #540 where I've ensured that an instruction from the 8.2 instruction  set is used to decide on the arch (rather than `asimd` which already appears in the 8.0 instruction set, and is why the Pi 4 was making a false match). If we wanted an explicit build for the 8.0 instruction set, we'd have to do a bit of work with `archdetect` (reintroduce `asimd` as another distinction)

PR #227, opened 2023-02-10
fix checks for read/write access using eessi_container.sh script

boegel:


boegel:
Maybe we should avoid using `alien` to convert the Apptainer RPM to `.deb`, and just use the prebuilt `.deb` that's provided now, see https://apptainer.org/docs/admin/latest/installation.html#install-debian-packages

trz42:
> Maybe we should avoid using `alien` to convert the Apptainer RPM to `.deb`, and just use the prebuilt `.deb` that's provided now, see https://apptainer.org/docs/admin/latest/installation.html#install-debian-packages

Didn't work. See https://github.com/trz42/software-layer/actions/runs/4236681193 😞

bedroge:
I had some suspicions that it could be related to the host OS/kernel, so maybe it's worth a shot to try Ubuntu 22.04 instead of 20.04?

trz42:
> I had some suspicions that it could be related to the host OS/kernel, so maybe it's worth a shot to try Ubuntu 22.04 instead of 20.04?

Can try that. My VM, where it works, has a slightly older kernel (5.4 vs 5.15).

laraPPr:
@boegel @bedroge @trz42  can you reopen this in https://github.com/EESSI/software-layer-scripts if you still want this feature.

Issue #217, opened 2023-01-12
GitHub API rate limits hit when using `EESSI-pilot-install-software.sh`


ocaisa:
I was developing for #212 and I hit the rate limits for GitHub after running that script only a couple of times. This is because there are many calls to `eb` which include `--from-pr` (and friends).

Using authenticated calls to the API would kick the can down the road for this (and we should be able to do this for the bot), but in general there is probably a better way to check if something is installed rather than calling the full `eb` command again (which is also quite expensive in terms of time). One approach could be to maintain a list of installed easyconfigs and check if an easyconfig is in the list or not before attempting the `eb` call.

There has already been some movement towards encapsulating the installation of an easyconfig in that script a bit better and we can see the shape appearing of an appropriate bash function:
```
EB=$1
target_ec=$2
ec_options=$3
ec_default_options="--robot"
list="/path/to/list/of/installed/easyconfigs/for/arch"

ec_installed=`grep -Fx "$target_ec" "$list" >/dev/null; echo $?`
case $ec_installed in
  0)
    # easyconfig installed
    echo ">> Target $target_ec already installed, skipping..."
    ;;
  1)
    # easyconfig not installed
    echo ">> Installing $target_ec..."
    ok_msg="$target_ec installed!"
    fail_msg="Failed to install $target_ec, woopsie..."
    $EB $cuda_samples_ec $ec_default_options $ec_options
    check_exit_code $? "${ok_msg}" "${fail_msg}"
    echo $target_ec >> $list
    ;;
  *)
    # some kind of error
    fatal_error "Checking for in $target_ec in $list returned error code $ec_installed"
    ;;
esac
```
Including some code for a rebuild case should also be straightforward enough.

The easystack format does not (right now) save us from the GitHub API calls, so that wouldn't rescue us here.

ocaisa:
This is more important for end users who may want to install/test something on top of EESSI in an overlay, especially once the stack grows in size (and more `--from-pr`s start to appear)

PR #228, opened 2023-02-14
Reimplement `mkdir -p` reporting where permissions break down

ocaisa:
This will allow us to log where creating directory structures under `host_injections` is breaking down.

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -31,3 +31,71 @@ function check_exit_code {
         fatal_error "${fail_msg}"
     fi
 }
+
+# Reimplement 'mkdir -p' with reporting on where permissions break down
+function create_directory_structure() {
+  # Ensure we are given a single path argument
+  if [ $# -ne 1 ]; then
+    echo "Function requires a single (relative or absolute) path argument" >&2
+    return 1
+  fi
+
+  # set a persistent variable that knows the full structure
+  # (i.e., retains the value upon recursive calls)
+  full_structure="${full_structure:="$1"}"
+
+  local directory_structure="$1"
+
+  # Check if directory exists and is writeable
+  if [ -d "${directory_structure}" ]; then
+    if [ "${directory_structure}" = "${full_structure}" ]; then
+        # release our (unneeded) global variable
+        unset full_structure
+    fi
+    if [ -w "${directory_structure}" ]; then
+      # Nothing to be done
+      return 0
+    else
+      echo "Directory ${directory_structure} exists but is not writeable" >&2
+      return 1
+    fi
+  fi
+
+  local directory_structure_parent=$(dirname "${directory_structure}")
+
+  # If the parent doesn't exist we need to create it
+  if [ ! -d "${directory_structure_parent}" ]; then
+    # Create the parent via a recursive call to this function
+    # (if this doesn't succeed we need to return the error code)
+    if ! create_directory_structure "${directory_structure_parent}"; then
+      if [ "${directory_structure}" = "${full_structure}" ]; then
+        # release our (unneeded) global variable
+        unset full_structure
+      fi
+      return 1
+    fi
+  fi
+
+  # Check the parent is writeable, and create the new subdir
+  if [ -w "${directory_structure_parent}" ]; then
+    if [ "${directory_structure}" = "${full_structure}" ]; then
+      # release our (unneeded) global variable
+      unset full_structure
+    fi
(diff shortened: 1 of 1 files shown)


ocaisa:
Rate limits are an issue with out PRs, see https://github.com/EESSI/software-layer/actions/runs/4175759364/jobs/7231122992#step:4:32
(a proposed solution is described in https://github.com/EESSI/software-layer/issues/217)

scripts/utils.sh
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
         fatal_error "${fail_msg}"
     fi
 }
+
+# Reimplement 'mkdir -p' with reporting on where permissions break down
+function create_directory_structure() {
+  # Ensure we are given a single path argument
+  if [ $# -ne 1 ]; then
+    echo "Function requires a single (relative or absolute) path argument" >&2
```
boegel, line 39: `Function` -> `create_directory_structure`

```suggestion
    echo "ERROR: create_directory_structure requires a single (relative or absolute) path argument" >&2

File: scripts/utils.sh [hunk 2]
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
...
+  if [ -d "${directory_structure}" ]; then
+    if [ "${directory_structure}" = "${full_structure}" ]; then
+        # release our (unneeded) global variable
+        unset full_structure
+    fi
+    if [ -w "${directory_structure}" ]; then
+      # Nothing to be done
+      return 0
+    else
+      echo "Directory ${directory_structure} exists but is not writeable" >&2
```
boegel, line 59:
```suggestion
      echo "ERROR: Directory ${directory_structure} exists but is not writeable" >&2

File: scripts/utils.sh [hunk 3]
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
...
+      return 1
```
boegel, line 60: Do we want different exit codes for different types of failures?

scripts/utils.sh
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
...
+    fi
+  fi
+
+  local directory_structure_parent=$(dirname "${directory_structure}")
+
+  # If the parent doesn't exist we need to create it
+  if [ ! -d "${directory_structure_parent}" ]; then
+    # Create the parent via a recursive call to this function
+    # (if this doesn't succeed we need to return the error code)
```
boegel, line 69: We're not returning the error code of the recursive call at all?

scripts/utils.sh
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
...
+    fi
+    if ! mkdir "${directory_structure}"; then
+      echo "'mkdir ${directory_structure}' failed for an unknown reason!" >&2
+      return 1
+    else
+      # Success!
+      return 0
+    fi
+  else
+    echo "Attempt to create ${full_structure} failed," \
```
boegel, line 93:
```suggestion
    echo "ERROR: Attempt to create ${full_structure} failed," \

File: scripts/utils.sh [hunk 6]
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
...
+  fi
+
+  # Check the parent is writeable, and create the new subdir
+  if [ -w "${directory_structure_parent}" ]; then
+    if [ "${directory_structure}" = "${full_structure}" ]; then
+      # release our (unneeded) global variable
+      unset full_structure
+    fi
+    if ! mkdir "${directory_structure}"; then
+      echo "'mkdir ${directory_structure}' failed for an unknown reason!" >&2
```
boegel, line 86:
```suggestion
      echo "'ERROR: mkdir ${directory_structure}' failed for an unknown reason!" >&2

**Review by boegel — CHANGES_REQUESTED** [review]

File: scripts/utils.sh [hunk 7]
```diff
@@ -31,3 +31,71 @@ function check_exit_code {
         fatal_error "${fail_msg}"
     fi
 }
+
+# Reimplement 'mkdir -p' with reporting on where permissions break down
```
boegel, line 35: @ocaisa Maybe it makes sense to elaborate on the use case for this function a bit?
Why is `mkdir -p` not good enough, is it mainly to get better error reporting?

Can we add a reference to an issue on something here with more info?

ocaisa:
Urgh, I'm going to close this. `mkdir -p` actually does report the location where directory creation fails. A simple wrapper is enough to also report the full path it was trying to create, so I will just add that to #212 instead

PR #232, opened 2023-02-18
enhancements/fixes for eessi_container.sh script

trz42:
This adds changes that are in use for the NESSI bot instances. Among the changes

- new arg `-l | --list-repos` to obtain a list of configured CernVM-FS repository identifiers
- args to specify HTTP(S) proxies
- `http(s)_proxy` env vars are used if set
- if `http_proxy` is set, the default.local is augmented with `CVMFS_HTTP_PROXY="${http_proxy}|PROXY_USING_IPv4"` (the `PROXY_USING_IPv4` uses the IPv4 address instead of the host name in `http_proxy`; it is determined by the script)
- ~for easier parsing of `--resume` arguments it prints `RESUME_FROM_DIR _DIR_` or `RESUME_FROM_TGZ _PATH_TO_TGZ_`~
- it stores repository information in the temporary storage to facilitate easier debugging of bot jobs


eessi_container.sh
```diff
@@ -43,17 +43,23 @@ MODE_UNKNOWN_EXITCODE=$((${ANY_ERROR_EXITCODE} << 5))
...
+export TMP_IN_CONTAINER=/tmp
+
 # repository cfg file, default name (default location: $PWD)
-#   can be overwritten by setting env var EESSI_REPOS_CFG_FILE_OVERRIDE
-export EESSI_REPOS_CFG_FILE="${EESSI_REPOS_CFG_FILE_OVERRIDE:=repos.cfg}"
+#   can be overwritten by setting env var EESSI_REPOS_CFG_DIR_OVERRIDE
+export EESSI_REPOS_CFG_FILE="${EESSI_REPOS_CFG_DIR_OVERRIDE:=${PWD}}/repos.cfg"
+# other repository cfg files in directory, default location: $PWD
+#   can be overwritten by setting env var EESSI_REPOS_CFG_DIR_OVERRIDE
+export EESSI_REPOS_CFG_DIR="${EESSI_REPOS_CFG_DIR_OVERRIDE:=${PWD}}"
```
boegel, line 62: suggestion:
```ini
export EESSI_REPOS_CFG_DIR="${EESSI_REPOS_CFG_DIR_OVERRIDE:=${PWD}}"
export EESSI_REPOS_CFG_FILE=${EESSI_REPOS_CFG_DIR}/repos.cfg
```

eessi_container.sh
```diff
@@ -263,6 +288,7 @@ else
   EESSI_HOST_STORAGE=$(mktemp -d --tmpdir eessi.XXXXXXXXXX)
   echo "Using ${EESSI_HOST_STORAGE} as tmp storage (add '--resume ${EESSI_HOST_STORAGE}' to resume where this session ended)."
 fi
+echo "RESUME_FROM_DIR ${EESSI_HOST_STORAGE}"
```
boegel, line 291: No need for this imho, we can just grep line above?

eessi_container.sh
```diff
@@ -279,18 +305,31 @@ fi
...
 # configure Singularity
-export SINGULARITY_CACHEDIR=${EESSI_TMPDIR}/singularity_cache
-mkdir -p ${SINGULARITY_CACHEDIR}
+if [[ -z ${SINGULARITY_CACHEDIR} ]]; then
+    export SINGULARITY_CACHEDIR=${EESSI_TMPDIR}/singularity_cache
+    mkdir -p ${SINGULARITY_CACHEDIR}
+fi
 [[ ${VERBOSE} -eq 1 ]] && echo "SINGULARITY_CACHEDIR=${SINGULARITY_CACHEDIR}"
 
+# pull & convert image and reset CONTAINER
```
boegel, line 322: clarify this comment a bit?

explain that we want to ensure that same container was used as last time?

eessi_container.sh
```diff
@@ -279,18 +305,31 @@ fi
...
+CONTAINER_URL_FMT=".*://(.*)"
+if [[ ${CONTAINER} == ${CONTAINER_URL_FMT} ]]; then
```
boegel, line 324: should be using `=~` here?

eessi_container.sh
```diff
@@ -331,6 +372,9 @@ else
   #    standard EESSI repositories)
   cfg_load ${EESSI_REPOS_CFG_FILE}
 
+  # copy repos.cfg to job directory --> makes it easier to inspect the job
+  cp ${EESSI_REPOS_CFG_FILE} ${EESSI_TMPDIR}/repos_cfg/.
```
boegel, line 376:
```
cp -a ${EESSI_REPOS_CFG_FILE} ${EESSI_TMPDIR}/repos_cfg/
```

eessi_container.sh
```diff
@@ -361,29 +405,61 @@ else
...
+    [[ ${VERBOSE} -eq 1 ]] && echo "http_proxy='${http_proxy}'"
+    PROXY_HOST=$(get_host_from_url ${http_proxy})
+    [[ ${VERBOSE} -eq 1 ]] && echo "PROXY_HOST='${PROXY_HOST}'"
+    PROXY_PORT=$(get_port_from_url ${http_proxy})
+    [[ ${VERBOSE} -eq 1 ]] && echo "PROXY_PORT='${PROXY_PORT}'"
+    HTTP_PROXY_IPV4=$(get_ipv4_address ${PROXY_HOST})
+    [[ ${VERBOSE} -eq 1 ]] && echo "HTTP_PROXY_IPV4='${HTTP_PROXY_IPV4}'"
+    echo "CVMFS_HTTP_PROXY=\"${http_proxy}|http://${HTTP_PROXY_IPV4}:${PROXY_PORT}\"" \
+       >> ${EESSI_TMPDIR}/repos_cfg/default.local
+    cat ${EESSI_TMPDIR}/repos_cfg/default.local
```
boegel, line 456: if VERBOSE, also add `echo "contents of .../default.local`

eessi_container.sh
```diff
@@ -456,6 +537,7 @@ if [[ ! -z ${SAVE} ]]; then
   fi
   tar cf ${TGZ} -C ${EESSI_TMPDIR} .
   echo "Saved contents of '${EESSI_TMPDIR}' to '${TGZ}' (to resume, add '--resume ${TGZ}')"
+  echo "RESUME_FROM_TGZ ${TGZ}"
```
boegel, line 540: no need for this, we can grep for `--resume` in line above

.github/workflows/test_eessi_container_script.yml
```diff
@@ -39,6 +41,28 @@ jobs:
...
+
+              # test use of --list-repos with custom repos.cfg
+              elif [[ ${{matrix.SCRIPT_TEST}} == 'listrepos_custom' ]]; then
+                outfile=out_listrepos.txt
+                outfile2=out_listrepos_2.txt
+                mkdir -p ${PWD}/cfg
+                echo "[EESSI/2021.12]" > cfg/repos.cfg
+                echo "repo_version = 2021.12" >> cfg/repos.cfg
+                echo "[EESSI/2023.02]" >> cfg/repos.cfg
+                echo "repo_version = 2023.02" >> cfg/repos.cfg
```
boegel, line 58: maybe use `20XY.AB` instead?

boegel (changes requested):

eessi_container.sh
trz42: Because we use wildcards only, `==` should work too. However, using `=~` might be safer anyway. Will change it.

eessi_container.sh
trz42: Done that plus significantly improved it. Should also handle much better when we are resuming from a previous session. And also handles the case that CONTAINER was already pointing to an image file.

eessi_container.sh
trz42: Removed.

eessi_container.sh
trz42: Done.

eessi_container.sh
trz42: Removed.

eessi_container.sh
trz42: Changed as suggested.

eessi_container.sh
trz42: Added and only printing when VERBOSE is set.

.github/workflows/test_eessi_container_script.yml
trz42: Changed.

trz42:
Implemented first round of requested changes in commit https://github.com/EESSI/software-layer/pull/232/commits/47cc1c6be44896675949567d88f5ae85a998b64f

eessi_container.sh
```diff
@@ -279,18 +303,62 @@ fi
...
+#     path to an image file)
+CONTAINER_IMG=
+CONTAINER_URL_FMT=".*://(.*)"
+if [[ ${CONTAINER} =~ ${CONTAINER_URL_FMT} ]]; then
+    # replace : and - with _ in match (everything after ://) and append .sif
+    CONTAINER_IMG=${BASH_REMATCH[1]//[:-]/_}.sif
+    # pull container to ${EESSI_TMPDIR} if it is not there yet (i.e. when
+    # resuming from a previous session)
+    if [[ ! -x ${EESSI_TMPDIR}/${CONTAINER_IMG} ]]; then
+        singularity pull ${EESSI_TMPDIR}/${CONTAINER_IMG} ${CONTAINER}
```
boegel, line 345: There will be situations in which you do want to re-pull the container image rather than using the existing one, for example if a fix was deployed for the container image.

So we should have a separate option for this, to enforce re-pulling the image? (future work, not in this PR)

eessi_container.sh
```diff
@@ -279,18 +303,62 @@ fi
...
+    fi
```
boegel, line 346: else here with verbose message: "reusing existing container image ..." ?

eessi_container.sh
boegel: Make sure that quiet option is also used here unless `--verbose` was used?

eessi_container.sh
boegel, line 346: `else` here with verbose message: "reusing existing container image ..." ?

boegel (changes requested):

boegel (approved):

PR #237, opened 2023-03-07
fix check for missing installations

### LINKED ISSUE #236: `check_missing_installations.sh` fails to capture missing installations correctly
Seems that `check-missing-installations.sh` in combination with `EESSI-pilot-install-software.sh` fails to capture missing installations correctly:
```bash
[foo ~]$ cat fake-stack.yml
easyconfigs:
  - EasyBuild-4.6.2.eb
  - xxd-8.2.4220-GCCcore-10.2.0.eb
  - Nextflow-22.10.1.eb
[foo ~]$ ml --show-hidden av EasyBuild/4.6.2

--------------------------------------------------------------------------------------------------------- /cluster/modulefiles/all ----------------------------------------------------------------------------------------------------------
   EasyBuild/4.6.2 (H)

  Where:
   H:  Hidden Module

Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".


[foo ~]$ ml --show-hidden av xxd/8.2.4220-GCCcore-10.2.0

--------------------------------------------------------------------------------------------------------- /cluster/modulefiles/all ----------------------------------------------------------------------------------------------------------
   xxd/8.2.4220-GCCcore-10.2.0

Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".


[foo ~]$ ml --show-hidden av Nextflow/22.10.1
No modules found!
Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching any of the "keys".


[foo ~]$ eb --easystack fake-stack.yml --missing --experimental
== Temporary log file in case of crash /tmp/eb-7g6d1dfs/easybuild-_5z6vudw.log
== found valid index for /cluster/software/EasyBuild/4.7.0/easybuild/easyconfigs, so using it...

No missing modules!

== found valid index for /cluster/software/EasyBuild/4.7.0/easybuild/easyconfigs, so using it...
== found valid index for /cluster/software/EasyBuild/4.7.0/easybuild/easyconfigs, so using it...

No missing modules!

== found valid index for /cluster/software/EasyBuild/4.7.0/easybuild/easyconfigs, so using it...

2 out of 3 required modules missing:

* Java/11.0.16 (Java-11.0.16.eb)
* Nextflow/22.10.1 (Nextflow-22.10.1.eb)

== Temporary log file(s) /tmp/eb-7g6d1dfs/easybuild-_5z6vudw.log* have been removed.
== Temporary directory /tmp/eb-7g6d1dfs has been removed.
```
For each line (easyconfig file) in the easystack file the test is run (probably fine), which then prints several `No missing modules!` lines (also fine), but `check_missing_installations.sh` just greps for this line in the output of the command `eb --easystack fake-stack.yml --missing --experimental` and therefore does not catch the missing installation.

trz42:
fixes #236 

Added comments to explain the needed change.

Should we add a test for this?

### DIFF
--- a/check_missing_installations.sh
+++ b/check_missing_installations.sh
@@ -38,9 +38,6 @@ ${EB:-eb} --from-pr 16531 --easystack eessi-${EESSI_PILOT_VERSION}.yml --experim
 
 grep " required modules missing:" ${eb_missing_out} > /dev/null
 
-# we need to process the result (from finding `No missing modules` to NOT finding
-# ` required modules missing:` and no other error happened)
-#
 # if grep returns 1 (` required modules missing:` was NOT found), we set
 # MODULES_MISSING to 0, otherwise (it was found or another error) we set it to 1
 [[ $? -eq 1 ]] && MODULES_MISSING=0 || MODULES_MISSING=1
--- a/check_missing_installations.sh
+++ b/check_missing_installations.sh
@@ -27,5 +27,21 @@ eb_missing_out=$LOCAL_TMPDIR/eb_missing.out
 # we need to use --from-pr to pull in some easyconfigs that are not available in EasyBuild version being used
 # PR #16531: Nextflow-22.10.1.eb
 ${EB:-eb} --from-pr 16531 --easystack eessi-${EESSI_PILOT_VERSION}.yml --experimental --missing | tee ${eb_missing_out}
-grep "No missing modules" ${eb_missing_out} > /dev/null
-check_exit_code $? "${ok_msg}" "${fail_msg}"
+
+# the above assesses the installed software for each easyconfig provided in
+# the easystack file and then print messages such as
+# `No missing modules!`
+# or
+# `2 out of 3 required modules missing:`
+# depending on the result of the assessment. Hence, we need to check if the
+# output does not contain any line with ` required modules missing:`
+
+grep " required modules missing:" ${eb_missing_out} > /dev/null
+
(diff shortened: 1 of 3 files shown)


trz42:
Converted to a **draft** because it is only necessary if we use EB v4.7.0 or newer.

trz42:
Changed script works for EB v4.5.1 (current latest EB version in EESSI pilot).

Added a test that captures a missing package.

The tests may fail if EB v4.7.0 or newer is used AND the format of the easystack file is not updated.

check_missing_installations.sh
```diff
@@ -27,5 +27,21 @@ eb_missing_out=$LOCAL_TMPDIR/eb_missing.out
...
+# `No missing modules!`
+# or
+# `2 out of 3 required modules missing:`
+# depending on the result of the assessment. Hence, we need to check if the
+# output does not contain any line with ` required modules missing:`
+
+grep " required modules missing:" ${eb_missing_out} > /dev/null
+
+# we need to process the result (from finding `No missing modules` to NOT finding
+# ` required modules missing:` and no other error happened)
```
boegel, line 42: @trz42 This comment looks like a TODO description, can be removed?

boegel (changes requested):

boegel (approved):
lgtm

PR #238, opened 2023-03-17
restore $PATH and $PYTHONPATH only after last run of pip installed eb

trz42:
When starting a fresh software subdir, the `eb` command in https://github.com/EESSI/software-layer/blob/25943f55ddfa3e5e397335a6be7654426d2f108e/EESSI-pilot-install-software.sh#L172 was not found because the PATH was restored too early. See https://github.com/EESSI/software-layer/blob/25943f55ddfa3e5e397335a6be7654426d2f108e/EESSI-pilot-install-software.sh#L169-L170

This PR moves these lines to after the `if` block.

It's been tested with ~ 2 PRs for NESSI starting fresh software subdirs.

The lines originated from #224

### DIFF
--- a/EESSI-pilot-install-software.sh
+++ b/EESSI-pilot-install-software.sh
@@ -165,10 +165,6 @@ else
     eb --install-latest-eb-release &> ${eb_install_out}
     check_exit_code $? "${ok_msg}" "${fail_msg}"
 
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

Issue #225, opened 2023-02-01
Prevent .pyc files from being added to tarballs


bedroge:
Not sure when this started happening, but for the tarballs of PR #195 (see e.g. https://github.com/EESSI/staging/pull/129) I noticed that they were changing an existing Python installation. It turns out that it's because of a new `.pyc` file that is generated in the Python installation directory:

```
$ ls -l 2021.12/software/linux/aarch64/generic/software/Python/3.9.5-GCCcore-10.3.0/easybuild/python/__pycache__/
total 12
-rw-r--r-- 1 bob bob 1493 Feb  1 02:38 sitecustomize.cpython-310.pyc
-rw-r--r-- 1 bob bob 1469 Jan 19 13:08 sitecustomize.cpython-39.opt-1.pyc
-rw-r--r-- 1 bob bob 1469 Jan  8  2022 sitecustomize.cpython-39.pyc
```

We have run into something similar before, and added a fix for this to the installation script that should write the `.pyc` files to a temporary directory:
https://github.com/EESSI/software-layer/blob/main/EESSI-pilot-install-software.sh#L81
But for some reason that doesn't seem to work here.

trz42:
We have seen this happening when building `FlexiBLAS-3.0.4-GCC-10.3.0.eb` for NESSI/2022.11, too. The existing `sitecustomize.cpython-310.pyc` is overwritten. Possibly this happens when tests are run. `FlexiBLAS-3.0.4-GCC-10.3.0.eb` has a build dependency on Python/3.9.5
```python
    ('Python', '3.9.5'),  # required for running the tests
```

This happens despite `PYTHONPYCACHEPREFIX` being set in `EESSI-pilot-install-software.sh`. It might be that the test run for LAPACK (see https://gitlab.mpi-magdeburg.mpg.de/software/flexiblas-release/-/blob/master/test/lapack-3.10.0/CMakeLists.txt#L27) does not take the setting into account or it is simply not present when the test is called.

[...]

PR #239, opened 2023-03-17
only tar up software directories for which modules were generated

trz42:
This lets `create_tarball.sh` only include software directories for which module files have been created.

It implements the first option in https://github.com/EESSI/software-layer/issues/225#issuecomment-1473437691

It has been tested interactively (resuming from a previous build job) and also in a bot job (https://github.com/NorESSI/software-layer/pull/57).

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -38,6 +38,7 @@ cd ${overlay_upper_dir}/versions/
 echo ">> Collecting list of files/directories to include in tarball via ${PWD}..."
 
 files_list=${tmpdir}/files.list.txt
+module_files_list=${tmpdir}/module_files.list.txt
 
 if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/.lmod ]; then
     # include Lmod cache and configuration file (lmodrc.lua),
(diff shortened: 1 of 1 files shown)


create_tarball.sh
```diff
@@ -49,12 +50,26 @@ if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules ]; then
     find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type f | grep -v '/\.wh\.' >> ${files_list}
     # module symlinks
     find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules -type l | grep -v '/\.wh\.' >> ${files_list}
+    # module files and symlinks
+    find ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules/all -type f -o -type l \
+        | grep -v '/\.wh\.' | sed -e 's/.lua$//' | awk -F'/' '{printf "%s/%s\n", $(NF-1), $NF}' | sort | uniq \
```
boegel, line 55: avoid hardcoding standard EasyBuild module naming scheme by using `sed 's@.*/modules/all/@@g'`

use `sort -u` instead of `sort | uniq`

create_tarball.sh
```diff
@@ -49,12 +50,26 @@ if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/modules ]; then
...
+        >> ${module_files_list}
 fi
-if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/software ]; then
-    # installation directories
-    ls -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/software/*/* | grep -v '/\.wh\.' >> ${files_list}
+if [ -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/software -a -r ${module_files_list} ]; then
+    # installation directories but only those for which module files were created
+    for package_version in $(cat ${module_files_list}); do
+        echo "handling ${package_version}"
+        ls -d ${pilot_version}/software/${os}/${cpu_arch_subdir}/software/${package_version} \
```
boegel, line 62: Add a comment here that the assumption here is that the module name and name of software installation directory match.
That will be the case as long as we use the standard EasyBuild module naming scheme, but will not be the case if we ever switch to another module naming scheme (like `HierarchicalMNS`)

boegel (changes requested):

create_tarball.sh
trz42: Changed via https://github.com/EESSI/software-layer/pull/239/commits/4d132273569f6d40e503bdd48b5b3aca88f7c907

boegel (approved):
lgtm

PR #244, opened 2023-04-24
improve `check_missing_installations.sh`

### LINKED ISSUE #243: script to check for missing installations reports false positive
We hit two cases where `check_missing_installations.sh` fails to report missing installations:

1. The easystack file does not exist. `eb` reports `ERROR: Failed to read eessi-2023.04.yml: [Errno 2] No such file or directory: 'eessi-2023.04.yml'`
2. The easystack file contains errors, eg, a misspelled option such as `include-easyblock-from-pr` (correct spelling: `include-easyblocks-from-pr`). `eb` reports `eb: error: no such option: --include-easyblock-from-pr`

In both cases, the `eb` command reports an error but the error is not caught by the analysis of the output. It is not searched for, but also the error message is written to `stderr` which is not redirected to the output file.

`stderr` should be redirected such that it is also captured. Then, the exit code of the `eb` command could/should be analysed and the whole script should only continue (that is analysing the output of `eb`) if `eb` succeeded.

trz42:
Fixes #243 

- redirects `stderr` into output file
- captures exit code of `eb` (first command in pipe, captured via `${PIPESTATUS[0]}`)
- analyses exit code of `eb`
- stores exit code of `grep` in `exit_code`

Can be tested as follows
1. Set pilot version via `export EESSI_PILOT_VERSION=YYYY.MM`
2. Test case that easystack file does not exist:
   - Set version such that no `eessi-${EESSI_PILOT_VERSION}.yml` exists in the current directory.
   - Then simply run `check_missing_installations.sh`. It should report an error and ask to check a log file.
3. Test case that easystack file contains spelling mistakes:
   - Create simple easystack file named `eessi-${EESSI_PILOT_VERSION}.yml` in the current directory with content such as
     ```yaml
     easyconfigs:
       - GCC-9.3.0.eb:
           options:
             option-that-does-not-exist
     ```
   - Then simply run `check_missing_installations.sh`. It should report an error and ask to check a log file.

### DIFF
--- a/check_missing_installations.sh
+++ b/check_missing_installations.sh
@@ -3,6 +3,7 @@
 # Script to check for missing installations in EESSI pilot software stack (version 2021.12)
 #
 # author: Kenneth Hoste (@boegel)
+# author: Thomas Roeblitz (@trz42)
 #
 # license: GPLv2
 #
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm, tested on top of `2021.12`

PR #245, opened 2023-04-27
Make sure installations are read only by enabling --read-only-installdir EasyBuild configure option

ocaisa:
This should help us avoid the issue we saw with `.pyc` files


ocaisa:
We are still hitting rate limits in our CI: https://github.com/EESSI/software-layer/actions/runs/4819671118/jobs/8583176010?pr=245#step:5:54

boegel (approved):

PR #255, opened 2023-05-31
flesh out load_easybuild_module.sh script from EESSI-pilot-install-software.sh

boegel:


load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+
+if [[ $? -eq 0 ]]; then
+    echo_green ">> Module for EasyBuild v${EB_VERSION} found!"
+else
+    echo_yellow ">> No module yet for EasyBuild v${EB_VERSION}, installing it..."
+
+    EB_TMPDIR=${TMPDIR}/ebtmp
+    echo ">> Temporary installation (in ${EB_TMPDIR})..."
+    pip_install_out=${TMPDIR}/pip_install.out
+    pip3 install --prefix ${EB_TMPDIR} easybuild &> ${pip_install_out}
```
trz42, line 48: Does this (usually) install the latest release?

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+    ORIG_PATH=${PATH}
+    ORIG_PYTHONPATH=${PYTHONPATH}
+
+    echo ">> Final installation in ${EASYBUILD_INSTALLPATH}..."
+    export PATH=${EB_TMPDIR}/bin:${PATH}
+    export PYTHONPATH=$(ls -d ${EB_TMPDIR}/lib/python*/site-packages):${PYTHONPATH}
+    eb_install_out=${TMPDIR}/eb_install.out
+    ok_msg="Latest EasyBuild release installed, let's go!"
+    fail_msg="Installing latest EasyBuild release failed, that's not good... (output: ${eb_install_out})"
+    ${EB} --install-latest-eb-release 2>&1 | tee ${eb_install_out}
```
trz42, line 60: Is it necessary to install a module for the latest release?

Maybe we don't want that version (may not hurt, but might be confusing). If the `eb` installed with `pip` is the latest release we have two latest installations, but only use the `pip`-installed `eb` for installing the latest release and the required version.

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+    ${EB} --search ${eb_ec} | grep ${eb_ec} > /dev/null
+    if [[ $? -eq 0 ]]; then
+        echo "Easyconfig found for EasyBuild v${EB_VERSION}, so installing it..."
+        ok_msg="EasyBuild v${EB_VERSION} installed, alright!"
+        fail_msg="Installing EasyBuild v${EB_VERSION}, yikes! (output: ${eb_install_out})"
+        #${EB} EasyBuild-${EB_VERSION}.eb >> ${eb_install_out} 2>&1
+        ${EB} EasyBuild-${EB_VERSION}.eb 2>&1 | tee -a ${eb_install_out}
+        check_exit_code $? "${ok_msg}" "${fail_msg}"
+    else
+        echo "No easyconfig found for EasyBuild v${EB_VERSION}"
```
trz42, line 73: If we want to install version 4.7.2 (the currently latest release), it won't be found by the latest release, right? But because we installed it already via `--install-latest-eb-release`, it is already there.

Maybe it is sufficient to clarify that message a bit, e.g., "No easyconfig ..., it might nevertheless already installed as the latest release. Check subsequent messages regarding the module EasyBuild/${EB_VERSION}."

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+    fi
+
+    # restore origin $PATH and $PYTHONPATH values, and clean up environment variables that are no longer needed
+    export PATH=${ORIG_PATH}
+    export PYTHONPATH=${ORIG_PYTHONPATH}
+    unset EB_TMPDIR ORIG_PATH ORIG_PYTHONPATH
+
+    module avail easybuild/${EB_VERSION} &> ${ml_av_easybuild_out}
+    if [[ $? -eq 0 ]]; then
+        echo_green ">> EasyBuild module installed!"
```
trz42, line 83: Maybe add '/${EB_VERSION}' to be consistent with other messages.

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+    else
+        fatal_error "EasyBuild/${EB_VERSION} module failed to install?! (output of 'pip install' in ${pip_install_out}, output of 'eb' in ${eb_install_out}, output of 'ml av easybuild' in ${ml_av_easybuild_out})"
```
trz42, line 85: I guess `module avail` (what is run) is exactly the same as `ml av`? If not might be a little less confusing to write

"output of 'module avail ...'"

.github/workflows/tests_scripts.yml
```diff
@@ -33,6 +37,48 @@ jobs:
...
+          echo '' >> ${test_script}
+          echo "source /tmp/load_easybuild_module.sh ${EB_VERSION}" >> ${test_script}
+          echo 'module list' >> ${test_script}
+          echo 'eb --version' >> ${test_script}
+
+          chmod u+x ${test_script}
+
+          # run wrapper script + capture & check output
+          out="${TMPDIR}/eb-${EB_VERSION}.out"
+          ./build_container.sh run /tmp/$USER/EESSI /tmp/run_in_compat_layer_env.sh ${test_script} 2>&1 | tee ${out}
```
trz42, line 77: Should use `eessi_container.sh` 😛

trz42 (changes requested):
Nice work! Some minor questions/suggestions.

.github/workflows/tests_scripts.yml
boegel: Yeah, good point, fixed in abb931c (was trickier than I anticipated!)

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+
+set -o pipefail
+
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <EasyBuild version>" >&2
+    exit 1
+fi
+
+# don't use $EB_VERSION, since that enables always running 'eb --version'
+EB_VERSION=${1}
```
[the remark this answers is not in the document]
boegel: Comment should say not to use `$EASYBUILD_VERSION`, fixing that...

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+
+# make sure that environment variables that we expect to be set are indeed set
+if [ -z "${TMPDIR}" ]; then 
+    echo "\$TMPDIR is not set" >&2
+    exit 2
+fi
+
+if [ -z "${EB}" ]; then
```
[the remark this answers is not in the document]
boegel: OK, will clarify with a comment

load_easybuild_module.sh
```diff
@@ -0,0 +1,108 @@
...
+    ${EB} --install-latest-eb-release 2>&1 | tee ${eb_install_out}
+    check_exit_code $? "${ok_msg}" "${fail_msg}"
+
+    eb_ec=EasyBuild-${EB_VERSION}.eb
+    ${EB} --search ${eb_ec} | grep ${eb_ec} > /dev/null
+    if [[ $? -eq 0 ]]; then
+        echo "Easyconfig found for EasyBuild v${EB_VERSION}, so installing it..."
+        ok_msg="EasyBuild v${EB_VERSION} installed, alright!"
+        fail_msg="Installing EasyBuild v${EB_VERSION}, yikes! (output: ${eb_install_out})"
+        #${EB} EasyBuild-${EB_VERSION}.eb >> ${eb_install_out} 2>&1
```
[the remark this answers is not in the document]
boegel: Of course, will do

load_easybuild_module.sh
boegel: Yes, if no version is specified (like `pip install ... easybuild==4.7.2`), then the latest release available in PyPI is automatically installed

load_easybuild_module.sh
boegel: It's not strictly necessary to install the latest release, but keep in mind there's 2 scenarios here:

1) If the requested EasyBuild version is an older version, say `4.5.0`, since installing the current latest version (`4.7.2`) as a module is overkill, since `EasyBuild/4.5.0` can also be installed with the temporary installed that was obtained with `pip`.
2) If the requested EasyBuild version is the latest release, then the temporary EasyBuild installation with `pip` does not include an easyconfig file for that version of EasyBuild (chicken-egg problem: the release has to be available before an easyconfig file can be created for it).

Scenario 2 is why we use `eb --install-latest-eb-release`, since that download the easyconfig file for the latest EasyBuild release from [the `develop` branch of the central easyconfigs repository](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/e/EasyBuild) (having an easyconfig file available there for the latest EasyBuild release is a part of the procedure for releasing EasyBuild).

So while in scenario 1 the module for the latest EasyBuild release obtained via `eb --install-latest-eb-release` is overkill, it's necessary in scenario 2 (and then the `eb --search` is futile, the installation using an easyconfig file that's included with EasyBuild can't be done).

We can consider changing this approach, but the benefit would be very small, and I think that should be done in a follow-up PR. The whole point of this PR is to put the installation procedure of EasyBuild in a stand-alone script, so it can be used to kickstart a new EESSI pilot version.

load_easybuild_module.sh
boegel: It would be better to do an additional `module avail` check after `--install-latest-eb-release`, I'll look into implementing that.

load_easybuild_module.sh
boegel: Makes sense, will do

load_easybuild_module.sh
boegel: `ml av` and `module avail` are the same, indeed, but I'll use `module avail` consistently to avoid confusion

load_easybuild_module.sh
trz42: Thanks a lot for the explanation! Fully agree that we could change the approach in a follow-up PR. So no need to change it here.

If we want to change it, we could do something like this:
1. `pip install` latest `eb` release.
2. check if the required version ($EB_VERSION) can be installed with the `pip`-installed `eb`, e.g., run `eb --search EasyBuild-${EB_VERSION}.eb` or similar. if so, install it.
3. if not, check if the `pip`-installed `eb` has the version that is required, if so run `eb --install-latest-eb-release ...`

load_easybuild_module.sh
```diff
@@ -0,0 +1,116 @@
...
+        if [[ $? -eq 0 ]]; then
+            echo "Easyconfig ${eb_ec} found for EasyBuild v${EB_VERSION}, so installing it..."
+            ok_msg="EasyBuild v${EB_VERSION} installed, alright!"
+            fail_msg="Installing EasyBuild v${EB_VERSION}, yikes! (output: ${eb_install_out})"
+            ${EB} EasyBuild-${EB_VERSION}.eb 2>&1 | tee -a ${eb_install_out}
+            check_exit_code $? "${ok_msg}" "${fail_msg}"
+        else
+            fatal_error "No easyconfig found for EasyBuild v${EB_VERSION}"
+        fi
+    fi
```
trz42, line 82: Nice change!

trz42 (approved):

trz42 (dismissed):
Well done. We don't have to change the `eb` installation procedure, and if we want to, we can do it in a later PR.

Last suggestion. Maybe add LICENSE and author to the comment section at the start. Could also be done later.

PR #241, opened 2023-04-19
first version of `bot/check-result.sh` script

trz42:
It's been tested with NESSI bot instances (see [NorESSI/software-layer PR125](https://github.com/NorESSI/software-layer/pull/125)). It requires [EESSI/eessi-bot-software-layer PR174](https://github.com/EESSI/eessi-bot-software-layer/pull/174).

Because the tests for NESSI included other changes than just [EESSI/eessi-bot-software-layer PR174](https://github.com/EESSI/eessi-bot-software-layer/pull/174) it would be good to repeat a more limited test.

The script contains extensive sections with examples for HTML blobs for how the produced result to be added to PR comment looks. They might be moved elsewhere (for documentation purposes) or removed.

The script is also quite long. Mostly because it really intends to support the bot to just use anything produced by the script. Probably this could be improved.

Retargeted to EESSI/2023.06.

PR #260, opened 2023-06-14
add EasyBuild 4.7.2 + foss/2021a to EESSI pilot 2023.06

boegel:
~~WIP because we need to have the compat layer for 2023.06 in place first (see https://github.com/EESSI/compatibility-layer/pull/188)~~

This has been tested on top of a test build of the 2023.06 compat layer (which includes OpenSSL 1.1.1 and GCC 10.4.0)  though, it works.

It includes a number of changes:

* CI workflows in `.github/` are updated to use 2023.06 version of EESSI pilot;
  * requires that `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/init` is in place, so requires #265 + deployment of `init` tarball for `2023.06`;
* `EESSI-pilot-install-software.sh` is updated so main body is a simple double for loop over EasyBuild versions and easystack files;
* `check_missing_installations.sh` script is updated to take path to easystack file to use for checking;
* additional EasyBuild hook for OpenBLAS was added to `eb_hooks.py` (+ some minimal refactoring);
* easystack file was added for installing `foss/2021a`;


boegel:
bot: build arch:x86_64/amd/zen2 repo:eessi-2023.06-software
bot: build arch:aarch64/graviton3 repo:eessi-2023.06-software

boegel:
Job manager had crashed again, which is why it didn't pick up on job `5115` finishing at 12:08 UTC.
Job `5116` got stuck somehow, wasn't making progress (last part of output in Slurm build log was sanity check for `Python`).

I'll re-trigger the bot to build for all CPU targets, let's see how that goes...

boegel:
bot: show_config

boegel:
Bot has been reconfigured with configuration file from #262 (cf85694390f4c44d6461e386f8ad052a6d89cc00), so it uses `aarch64/neoverse_n1` and `aarch64/neoverse_v1` instead of `aarch64/graviton*`.

boegel:
bot: build repo:eessi-2023.06-software

EESSI-pilot-install-software.sh
```diff
@@ -137,235 +137,29 @@ else
...
-$EB SciPy-bundle-2021.05-foss-2021a.eb --robot
-check_exit_code $? "${ok_msg}" "${fail_msg}"
+    for gen in '2021a'; do
+
+        es="eessi-${EESSI_PILOT_VERSION}-eb-${eb_version}-${gen}.yml"
+
+        if [ -f ${es} ]; then
+            echo_green "Feeding easystack file ${es} to EasyBuild..."
+
+            ${EB} --easystack ${TOPDIR}/${es} --missing && ${EB} --easystack ${TOPDIR}/${es} --robot
```
trz42, line 154: The first part

```bash
${EB} --easystack ${TOPDIR}/${es} --missing
```

lets the downstream checking of the result of a build job **always** report a failure because it produces lines such as

```
2 out of 2 required modules missing
```
and the checking of the result just scans for lines containing ` required modules missing`

It's not immediately clear what benefits this additional `eb ... --missing` command has. If it is essential, its output should be written to another file or every downstream action that relies on the message being there or not need to be adjusted.

trz42 (changes requested):
Just a small finding with a recent build for NESSI.

EESSI-pilot-install-software.sh
boegel: Oh, I totally overlooked that...

I was only including this as an informative message of what needs to be installed to complete the request, but since that interferes with the missing installations check done at the end, I'll just remove it.

boegel:
bot: build arch:aarch64/neoverse_v1 repo:eessi-2023.06-software

boegel:
Hmm, job `5170` got stuck somehow, was still running after 17+h, so I've killed it.

EESSI-pilot-install-software.sh
boegel: fixed in 4cc8f41

boegel:
Bot crashed halfway submitting jobs, because I was fiddling with the `app.cfg` (to make sure the updates from #262 were in place, so I had to make another request for the remaining CPU targets... Works like a charm 🥳

boegel:
bot: build repo:eessi-2023.06-software

boegel:
Tests are passing now since:

* `foss/2021a` was built and deployed to `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/software` by the bot (via this PR, using `bot/check-result.sh` that was collapsed in here from #241)
* `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/init` directory is in place (cfr. #265)

ocaisa (approved):

PR #262, opened 2023-06-15
update bot configuration for EESSI 2023.06

boegel:


trz42:
I think this should go into a different repository. Maybe one that captures information about the setup of our infrastructure.

trz42:
> It used to be in a bot repo, but since the configuration is specific to the repo for which it's being used, we moved it here.

Partially true (if the bot would only build for a specific repository).
 
> If we move it somewhere else, don't we re-introduce that problem, and create a disconnect between the config and the repo for which it is used?

Possibly. Eventually one might want to split the configuration into settings which a specific to a bot instance (and store that in some `infrastructure` repository) and settings which are specific to the target repository (and store that in the target repository).

> That said, there's also stuff missing here (the `repos.cfg`)...

Those are probably target repo specific settings and could be moved there easily.

trz42 (approved):
Instead of blocking this, we can leave it (`bot-eessi-aws-citc.cfg`) in the software-layer repository.

Possibly one could add the `repos` directory which includes CVMFS config files and settings for different target repositories. Could also be done in a follow up PR.

PR #263, opened 2023-06-16
Allow `archdetect` to print all possible cpu paths

ocaisa:
Also fix incorrect generic path


ocaisa:
The idea with this PR is that if `archdetect` provides a full list of possible architecture matches. When used with a particular `version` of EESSI, you could cycle through the list until you find a path that actually exists.


It would allow us to keep `archdetect` independent of the `version` and universally update it.

PR #264, opened 2023-06-16
Allow `archdetect` to print all possible cpu paths + fix paths for `aarch64/*` (drop `/arm/` subdir)

ocaisa:
Also fix incorrect `generic` path.

The idea with this PR is that if `archdetect` provides a full list of possible architecture matches. When used with a particular `version` of EESSI, you could cycle through the list until you find a path that actually exists.

It would allow us to keep `archdetect` independent of the `version` and universally update it.


init/eessi_archdetect.sh
```diff
@@ -138,6 +145,6 @@ shift "$(($OPTIND -1))"
 ARGUMENT=${1:-none}
 
 case "$ARGUMENT" in
-    "cpupath") cpupath; exit;;
-    *) echo "$USAGE"; log "ERROR" "Missing <action> argument";;
+    "cpupath") cpupath $all; exit;;
```
boegel, line 148: @ocaisa Rather than passing an argument to `cpupath`, I would make `-a` change the default value of a `$CPUPATCH_RESULT` from `best` to `all`, which matches better with what's done with `$LOG_LEVEL`?

init/eessi_archdetect.sh
```diff
@@ -114,22 +115,28 @@ cpupath(){
...
-USAGE="Usage: eessi_archdetect.sh [-h][-d] <action>"
+USAGE="Usage: eessi_archdetect.sh [-h][-d][-a] <action>"
 
-while getopts 'hdv' OPTION; do
+while getopts 'hdva' OPTION; do
     case "$OPTION" in
         h) echo "$USAGE"; exit 0;;
         d) LOG_LEVEL="DEBUG";;
         v) echo "eessi_archdetect.sh v$VERSION"; exit 0;;
+        a) all="all";;
```
boegel, line 139: Ideally we also have a constant like `$CPUARCH_RESULT_ALL` (where `"cpuarch_result_all"` is the value) than we can use for both setting and checking, to avoid bugs due to silly typos (I guess we should also try adding `set -e` so use of unset environment variables are treated like errors)

boegel (changes requested):

init/eessi_archdetect.sh
ocaisa: Not quite sure what you mean here, but I think it is addressed in https://github.com/EESSI/software-layer/pull/264/commits/7db0616a0c78f40864507d77085817af6eb7028a

init/eessi_archdetect.sh
ocaisa: Should be addressed in https://github.com/EESSI/software-layer/pull/264/commits/7db0616a0c78f40864507d77085817af6eb7028a

boegel (approved):

PR #265, opened 2023-06-16
bump EESSI pilot version to 2023.06 + update create_directory_tarballs.sh script to use '2023.06' branch of EESSI/software-layer repo

boegel:


ocaisa (dismissed):
LGTM

boegel:
Hitting a chicken-egg situation: CI can only pass here if `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/*` is in place (which required deploying from #260), and merging this PR is required to get `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/` in place which is required for CI to pass in #260).

So we should probably first let the bot deploy the built tarballs from #260, re-trigger CI here, then deploy `2023.06/init`, which should be sufficient to make CI green in #260...

trz42 (approved):
Looks good.

PR #267, opened 2023-06-20
clean up local tmp if it has been saved to a tarball

trz42:
The `bot/build.sh` runs `eessi_container.sh` twice and each run results in a tarball containing the temporary folder. If a job runs on a machine where this temporary folder is not cleaned up automatically, disk usage can pile up leading to failing build jobs.

This PR checks if a temporary folder is accessible (`${STORAGE}`). If so it removes it if there is any tarball created for it.

Improves situation that resulted in creating https://github.com/EESSI/eessi-bot-software-layer/issues/186

### DIFF
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -207,7 +207,7 @@ if [[ -d ${STORAGE} ]]; then
     # double-check that at least one tarball of the temporary storage was created
     if check_tmp_tarball ${TARBALL_TMP_BUILD_STEP_DIR} || check_tmp_tarball ${TARBALL_TMP_TARBALL_STEP_DIR}; then
         echo "Removing temporary storage under '${STORAGE}'"
-        # rm -rf ${STORAGE}
+        rm -rf ${STORAGE}
     else
         echo "Did not find any tarball containing the temporary storage for neither the"
         echo "build nor the tar step. Hence, not removing the storage at '${STORAGE}'."
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -211,6 +211,7 @@ if [[ -d ${STORAGE} ]]; then
     else
         echo "Did not find any tarball containing the temporary storage for neither the"
         echo "build nor the tar step. Hence, not removing the storage at '${STORAGE}'."
+    fi
 else
     echo "Local disk storage at '${STORAGE}' not accessible, so wasn't cleaned up."
 fi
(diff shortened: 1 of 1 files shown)


bot/build.sh
```diff
@@ -196,4 +196,24 @@ echo "                     -- ./create_tarball.sh ${TMP_IN_CONTAINER} ${EESSI_PI
...
+    ls ${dir} | grep -E 'tgz$|\.gz$'
+    return $?
+}
+
+# clean storage used on local disk
+if [[ -d ${STORAGE} ]]; then
+    # double-check that at least one tarball of the temporary storage was created
+    if check_tmp_tarball ${TARBALL_TMP_BUILD_STEP_DIR} || check_tmp_tarball ${TARBALL_TMP_TARBALL_STEP_DIR}; then
+        echo "Removing temporary storage under '${STORAGE}'"
+        # rm -rf ${STORAGE}
```
boegel, line 210: There's little point in doing this if the actual `rm` command is commented out? :)

trz42: What 🤯

trz42:
bot: show_config

trz42:
bot: show_help

trz42:
bot: show_config

trz42:
bot: show_config

bot/build.sh
```diff
@@ -196,4 +196,24 @@ echo "                     -- ./create_tarball.sh ${TMP_IN_CONTAINER} ${EESSI_PI
...
+    ls ${dir} | grep -E 'tgz$|\.gz$'
+    return $?
+}
+
+# clean storage used on local disk
+if [[ -d ${STORAGE} ]]; then
+    # double-check that at least one tarball of the temporary storage was created
+    if check_tmp_tarball ${TARBALL_TMP_BUILD_STEP_DIR} || check_tmp_tarball ${TARBALL_TMP_TARBALL_STEP_DIR}; then
+        echo "Removing temporary storage under '${STORAGE}'"
+        rm -rf ${STORAGE}
```
bedroge, line 210: If I understand things correctly, `$STORAGE` refers to the base directory where all jobs create their unique subdirectory? In that case, shouldn't we only remove `${JOB_STORAGE}` here, as there may be other (running) jobs on the same machine?

trz42: I think, that's right.

Surprisingly, this `${JOB_STORAGE}` is just created (a subdirectory under `${STORAGE}`) but it is never used.

We should do two things
- replace uses of `${STORAGE}` by `${JOB_STORAGE}` when running `eessi_container.sh` and test properly if that still works (maybe a separate PR would be beneficial)
- replace uses of `${STORAGE}` by `${JOB_STORAGE}` in this PR here such that only job specific job temporary storage is removed

laraPPr:
@trz42 will you return to this pr? What is still require? It also seem that these changes should be maid in the repo and not moved to https://github.com/EESSI/software-layer-scripts but please correct me I'm wrong.

laraPPr:
Should also retarget this pr if you want to keep it

laraPPr:
@trz42 Can you retarget this pr?

ocaisa:
Closing, @trz42 an reopen under the other repo if needed

PR #268, opened 2023-06-20
fix bot config: add required `job_result_unknown_fmt` + use 16 cores for jobs

boegel:
Without this, the job manager crashes - required because of changes in https://github.com/EESSI/eessi-bot-software-layer/pull/174


trz42 (approved):
Looks good. Maybe you want to add more instances or someone wonders why a job is nor starting. There are fewer nodes for the larger instances configured.

If we want to support multiple jobs per node, we probably need to add `--mem=` and also make `local_tmp = /tmp/$USER/EESSI` job-specific. The latter could be achieved by adding `/$SLURM_JOB_ID`.

boegel:
We can easily scale up extra `*.4xlarge` nodes on the fly in the CitC cluster. I'll make sure there's 2 of each `*.4xlarge` type

PR #270, opened 2023-06-22
switch to colored emojis in bot/check-result.sh

boegel:
:x: instead of :heavy_multiplication_x:
✅ instead of ✔️


trz42 (approved):
Looks good.

PR #272, opened 2023-06-22
test load_easybuild_module.sh with a single EasyBuild version

boegel:
testing the `load_easybuild_module.sh` can take quite a bit of time, so better limit it to a single EasyBuild version (testing with multiple version doesn't really buy us much)


bedroge (approved):
lgtm

PR #278, opened 2023-06-23
change upload_policy to 'latest'

boegel:
Following a discussion in our group meeting this afternoon, I *think* we should change the upload policy to `latest`, but I'm not quite sure - have to check the code to fully grasp the difference.

Would love some feedback from @trz42 on this...


trz42 (approved):
Makes sense.

PR #299, opened 2023-07-06
{2023.06}[foss/2021a] RStudio-Server V1.4.1717-Java-11-R V4.1.0

TopRichard:


TopRichard:
bot: build repo:eessi-2023.06-software arch:x86_64/generic

ocaisa:
Not sure this one is a good idea, https://github.com/easybuilders/easybuild-easyconfigs/issues/13951
I had a similar PR open that had a lot more going on, https://github.com/easybuilders/easybuild-easyconfigs/pull/15524

boegel:
@TopRichard For `Java`, you'll probably need to use https://github.com/easybuilders/easybuild-easyblocks/pull/2557 via `--include-easyblocks-from-pr`

TopRichard:
bot: build repo:eessi-2023.06-software arch:x86_64/generic

bedroge:
Just checked the log of this build, and it failed during the build of `Xvfb/1.20.11-GCCcore-10.3.0`:

```
== postprocessing...
  >> running command:
        [started at: 2023-07-06 08:29:25]
        [working dir: /tmp/bot/easybuild/build/Xvfb/1.20.11/GCCcore-10.3.0/xorg-server-1.20.11]
        [output logged in /tmp/eb-akzyf_n9/eb-1_93q_0q/eb-lzt2gn2z/eb-9xo5zmng/eb-u5dyuu1k/eb-0vncxrhv/eb-l_ls80gm/eb-i4m2bxtz/eb-oq4jykos/eb-4ni3tfkx/easybuild-run_cmd-aqpwi3rn.log]
        cp -a xvfb-run /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/generic/software/Xvfb/1.20.11-GCCcore-10.3.0/bin/ && chmod u+x  /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/generic/software/Xvfb/1.20.11-GCCcore-10.3.0/bin/xvfb-run
  >> command completed: exit 1, ran in < 1s
== ... (took < 1 sec)
== FAILED: Installation ended unsuccessfully (build directory: /tmp/bot/easybuild/build/Xvfb/1.20.11/GCCcore-10.3.0): build failed (first 300 chars): cmd "cp -a xvfb-run /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/generic/software/Xvfb/1.20.11-GCCcore-10.3.0/bin/ && chmod u+x  /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/generic/software/Xvfb/1.20.11-GCCcore-10.3.0/bin/xvfb-run" exited with exit code 1 a (took 5 mins 3 secs)
== Results of the build can be found in the log file(s) /tmp/eb-akzyf_n9/eb-1_93q_0q/eb-lzt2gn2z/eb-9xo5zmng/eb-u5dyuu1k/eb-0vncxrhv/eb-l_ls80gm/eb-i4m2bxtz/eb-oq4jykos/eb-4ni3tfkx/easybuild-Xvfb-1.20.11-20230706.082422.yCixV.log
```

It's not really clear to me why that `cp` or `chmod` command failed. Do note the weird nested structure in the path of that log file, which is caused by https://github.com/easybuilders/easybuild-framework/issues/4291. Not sure of that's causing the issue here, though.

bedroge:
And digging a bit more... initially I suspected that it may be related to the version of `fuse-overlayfs`, which may be partly true. The old version (0.3) that we used before, does seem to work fine for files without write permission.

However, the main cause is probably that the file `xvfb-run` is part of the Easybuild installation (it's actually a patch for `Xvfb`, see https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/x/Xvfb/xvfb-run and https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/x/Xvfb/Xvfb-1.20.11-GCCcore-10.3.0.eb#L86), and somehow the file got these weird permission for our installation. It doesn't have these permissions in the Easybuild source tarball, and neither do I see them for any other versions. So I suspect something weird happened for our Easybuild installation, either in the build or in the tarball creation (it's not the ingestion, as the permissions are really like this in the original tarball).

I'm going to try and rebuild EasyBuild 4.7.2 and see if the same thing happens again.

ocaisa:
TBH I have also found that read-only installation led to a problem with an R extension installation that didn't expect a template file  to be read-only (and of course it wouldn't normally be).

bedroge:
I'm not sure if it makes sense to change this upstream, by chaing the `postinstallcmds` in the easyconfig from:
```
postinstallcmds = ["cp -a xvfb-run %(installdir)s/bin/ && chmod u+x  %(installdir)s/bin/xvfb-run"]
```
to something like:
```
postinstallcmds = ["chmod u+wx xvfb-run && cp -a xvfb-run %(installdir)s/bin/"]
```

but that does solve the issue. This can also be easily achieved by using a hook, so perhaps that's the quickest solution for now anyway. I've just tried the following, and that worked fine:
```
def parse_hook_xvfb_cp_permissions(ec, eprefix):
    """Add user write permission to xvfb-run before copying it in the postinstallcmds."""
    if ec.name == 'Xvfb':
        ec['postinstallcmds'] = ["chmod u+wx xvfb-run && cp -a xvfb-run %(installdir)s/bin/"]
        print_msg("Using custom postinstallcmds for %s: %s", ec.name, ec['postinstallcmds'])
    else:
        raise EasyBuildError("Xvfb-specific hook triggered for non-Xvfb easyconfig?!")

# and add it to the dict at the bottom:

PARSE_HOOKS = {
    ...
    'Xvfb': parse_hook_xvfb_cp_permissions,
}


```

TopRichard:
bot: build repo:eessi-2023.06-software arch:x86_64/generic

bedroge:
```
checking libcurl version ... 8.1.2
checking curl/curl.h usability... yes
checking curl/curl.h presence... yes
checking for curl/curl.h... yes
checking if libcurl is version 7 and >= 7.28.0... no
configure: error: libcurl >= 7.28.0 library and headers are required with support for https
```
Looks like the `curl` from our compat layer is too new...

eb_hooks.py
```diff
@@ -199,6 +199,15 @@ def parse_hook_ucx_eprefix(ec, eprefix):
         raise EasyBuildError("UCX-specific hook triggered for non-UCX easyconfig?!")
 
 
+def parse_hook_xvfb_cp_permissions(ec, eprefix):
+    """Add user write permission to xvfb-run before copying it in the postinstallcmds."""
+    if ec.name == 'Xvfb':
+        ec['postinstallcmds'] = ["chmod u+wx xvfb-run && cp -a xvfb-run %(installdir)s/bin/"]
```
boegel, line 205: @TopRichard Can you clarify why this is needed?

If there's a problem here, I'm not sure why we didn't see it outside of EESSI...

TopRichard: @boegel Xvfb is a dependency, and so it was set to avoid a failure.

boegel: I got that part, but I don't see why we would see this problem with Xvfb outside of EESSI.

I'm seeing it too in #328 where I'm trying to get `R` installed (just to break down this PR a bit), so I'll take a closer look.

bedroge: @boegel I debugged that issue some time ago, see https://github.com/EESSI/software-layer/pull/299#issuecomment-1629170830 and the messages below that one. TLDR (if I remember correctly): it's a combination of our writable overlay + setting the installation dir as read-only (and `xvfb-run` is shipped as "patch" in the EB installation dir).

laraPPr: I'm also running into the same issue with https://github.com/EESSI/software-layer/pull/334

laraPPr: @boegel Could the issue with Xvfb be related to the version of XZ? That is the only difference I can discern in EESSI.

ocaisa: It's (I believe) somehow related to the fact we now have write-only installations and the writable overlay. We should probably create an upstream PR to reverse the order of statements in the `Xvfb` `postinstallcmds`, which is what the hook update here is doing (or we just move forward with the hook update here).

laraPPr:
With the following pr, https://github.com/easybuilders/easybuild-easyconfigs/pull/18834/files. The installation of Xvfb works.

PR #302, opened 2023-07-09
copy build log to build logs dir if build failed

boegel:
requires https://github.com/EESSI/eessi-bot-software-layer/pull/195 + bot to be configured with `build_logs_dir` set


boegel:
tested via https://github.com/boegel/software-layer/pull/26, works as designed (and doesn't cause trouble if `build_logs_dir` is not specified)

PR #305, opened 2023-07-12
Set permissions on copied build log for failed builds

bedroge:
@laraPPr just found out that the build logs were not accessible, and this was due to the file having incorrect permissions:
```
-rw-------. 1 bot users 16103454 12 jul 09:21 /mnt/shared/bot-build-logs/jobs/5806/easybuild-y0xokk13.log
```

I've added a `chmod 0644` which makes the file readable for everyone. We could also do a `0640`, since all `citc` users should be in the `users` group, so both should have the same effect.


casparvl (approved):
Lgtm, will be useful for other CITC users to be able to access this :)

PR #306, opened 2023-07-14
[2023.06]{foss/2021b} SciPy-bundle v2021.10

laraPPr:


laraPPr:
bot: build arch:aarch64/neoverse_v1 repo:eessi-2023.06-software

laraPPr:
> @laraPPr do you have access to the AWS cluster? It would probably be faster to try and do this build yourself, rather than let the bot do it. It'll allow you to go through 'regular' EasyBuild debugging procedures (e.g. get the build environment with `--dump-env-script` and then try to rerun the test suite manually). If that reproduces your issue, it's much easier to debug, as you can probably start modifying things in the test suite, make it more verbose, increase tolerance, etc, to see where it fails.

I do not think I have enough access to the AWS cluster to build it myself. I've also never done  'regular' EasyBuild debugging procedures but I am willing to try.

laraPPr:
> I have seen `numpy` failures before that were related to the version of `OpenBLAS` underneath. I think we also see more test failures for OpenBLAS on ARM, right? I think we ignore those, since they are 'only' tolerance failures. But then any test on top (like a numpy test, which calls OpenBLAS) will likely also fail.

Yes there was a hook set for the installation of OpenBLAS on aarch64 that relaxed the allowed number of failing numerical LAPACK tests.

boegel:
In EESSI pilot 2021.12, we installed `SciPy-bundle-2020.03-foss-2020a-Python-3.8.2.eb` with `--skip-test-step` for `aarch64` to dance around failing tests, see https://github.com/EESSI/software-layer/blob/2f6709f019d9b4a0d7852883842d06da46b961fc/EESSI-pilot-install-software.sh#L220C25-L220C41
We did *not* have to do this for `SciPy-bundle-2021.05-foss-2021a.eb` however.

We have run into issues like this before in EasyBuild, see https://github.com/easybuilders/easybuild-easyconfigs/issues/11959 

I think we should try and asses how "serious" these test failures are, but most likely we'll need to move forward with ignoring the failing tests, at least for now, since fixing may require significant effort...
It's not like we guarantee for other software (where we may not be running tests at all) that it's 100% bugfree, so the effort we do has to be within reason imho: there's no way we can guarantee that all software we include in EESSI doesn't have any bugs at all, we have to basically provide it as-is (with some exceptions where patching is obvious/easy).

boegel:
bot: build repo:eessi-2023.06-software

boegel:
@laraPPr To ignore the failing `numpy` tests for now, maybe we can try simply removing the corresponding test modules (`fft/tests/test_pocketfft.py` and `polynomial/tests/test_classes.py`) right before running the tests, via `pretestopts` for example (controlled via a hook).
Maybe we'll need to run `python -c 'import fft.tests; print(fft.tests.__file__)'` to figure out the right place.

laraPPr:
bot: build arch:aarch64/neoverse_v1 repo:eessi-2023.06-software

eb_hooks.py
```diff
@@ -258,6 +258,10 @@ def pre_configure_hook_wrf_aarch64(self, *args, **kwargs):
     else:
         raise EasyBuildError("WRF-specific hook triggered for non-WRF easyconfig?!")
 
+def pre_test_hook_ignore_failing_tests_SciPybundle(self, *args, **kwargs):
+    if self.name == 'SciPy-bundle' and '/neoverse_v1/' in self.installdir:
```
boegel, line 262: @laraPPr looks good, but I would add some context here by including a comment.

Maybe the condition should also be a bit more specific for now, since installing `SciPy-bundle/2021.05-foss-2021a` worked fine (because we weren't so strict on the numpy/scipy tests yet then).

We want to restrict this to only ignore the failing tests on `neoverse_v1`, but the check for `/neoverse_v1` in `self.installdir` is a bit of a hack, I would check the value of `${EESSI_SOFTWARE_SUBDIR}` instead (which is used by `configure_easybuild` to set `$EASYBUILD_INSTALLPATH`).

So, something like this (untested, but should work):

```python
def pre_test_hook_ignore_failing_tests_SciPybundle(self, *args, **kwargs):
    # ignore failing test suite for numpy & co when installing SciPy-bundle on Arm neoverse_v1,
    # see <link to EESSI/software-layer issue with more goes here>
    cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
    if self.name == 'SciPy-bundle' and self.version == '2021.10' and cpu_target == 'aarch64/neoverse_v1':
```

Note that `get_eessi_envvar` will fail if `${EESSI_SOFTWARE_SUBDIR}` is not set, which is a good thing (in case these hooks are used in a different context).

boegel (changes requested):
@laraPPr You will also need to define `pre_test_hook` that picks up on `PRE_TEST_HOOKS`, similar to the `pre_configure_hook` that's already there.

laraPPr:
bot: build arch:aarch64/neoverse_v1 repo:eessi-2023.06-software

laraPPr:
bot: build arch:aarch64/neoverse_v1 repo:eessi-2023.06-software

laraPPr:
> @laraPPr Hook looks good to me now, let's make sure builds are done for all CPU targets so we can deploy this?
> 
> We should also open an internal issue with details on the failing tests, and start looking into reporting the problems upstream. Let's do at least the internal issue before merging this PR (to avoid forgetting about it).

Should I than also look at other test that were skipped before they failed?

boegel:
> > @laraPPr Hook looks good to me now, let's make sure builds are done for all CPU targets so we can deploy this?
> > We should also open an internal issue with details on the failing tests, and start looking into reporting the problems upstream. Let's do at least the internal issue before merging this PR (to avoid forgetting about it).
> 
> Should I than also look at other tests that might have failed and were skipped?

The issue should cover all failing tests in this SciPy-bundle, yes, so perhaps also `scipy` (if there are failing tests there).

You should be able to spot if there's additional failing test suite easily by looking for "`Test failure ignored`" in the EasyBuild log that is included with the installation.

boegel:
All staging PRs merged, so deployment is being done right now, should be fully in place in about an hour or so.

boegel:
@laraPPr Sorry, my mistake, the way to check for failing test steps is to check for lines that start with "``", since that's the message emitted by the `echo` that the hook injects.
I only see this for the numpy test suite in `/cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/aarch64/neoverse_v1/software/SciPy-bundle/2021.10-foss-2021b/easybuild/easybuild-SciPy-bundle-2021.10-20230822.105656.log.bz2`:

```
=========================== short test summary info ============================
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_identity - AssertionError:
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifft[None] - AssertionErr...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifft[backward] - Assertio...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifft[ortho] - AssertionEr...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifft[forward] - Assertion...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifft2 - AssertionError:
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_ifftn - AssertionError:
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_irfft2 - AssertionError:
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_irfftn - AssertionError:
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_all_1d_norm_preserving - ...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_dtypes[float16] - Asserti...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_dtypes[float32] - Asserti...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_dtypes[float64] - Asserti...
FAILED fft/tests/test_pocketfft.py::TestFFT1D::test_dtypes[float128] - Assert...
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[Polynomial]
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[Legendre]
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[Chebyshev]
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[Laguerre]
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[Hermite] - ...
FAILED polynomial/tests/test_classes.py::test_bad_conditioned_fit[HermiteE]
= 20 failed, 14429 passed, 81 skipped, 1253 deselected, 22 xfailed, 79 warnings in 171.82s (0:02:51) =
ignoring failing tests
```

laraPPr:
created issue for the numy test failures with SciPy-bundle:

[#318](https://github.com/EESSI/software-layer/issues/318)

boegel (approved):

PR #315, opened 2023-08-09
use shorter 'pilot' job name in workflow to check for missing installations

boegel:
This will help to make the overview of jobs in the workflow more readable, since there's no way to make the column wider:

Screenshot 2023-08-09 at 17 29 55


casparvl (approved):
Lgtm!

PR #310, opened 2023-07-18
{2023.06} foss/2022a

casparvl:
17 out of 37 required modules missing:

* FFTW/3.3.10-GCC-11.3.0 (FFTW-3.3.10-GCC-11.3.0.eb)
* UnZip/6.0-GCCcore-11.3.0 (UnZip-6.0-GCCcore-11.3.0.eb)
* cURL/7.83.0-GCCcore-11.3.0 (cURL-7.83.0-GCCcore-11.3.0.eb)
* libarchive/3.6.1-GCCcore-11.3.0 (libarchive-3.6.1-GCCcore-11.3.0.eb)
* CMake/3.23.1-GCCcore-11.3.0 (CMake-3.23.1-GCCcore-11.3.0.eb)
* libffi/3.4.2-GCCcore-11.3.0 (libffi-3.4.2-GCCcore-11.3.0.eb)
* make/4.3-GCCcore-11.3.0 (make-4.3-GCCcore-11.3.0.eb)
* Tcl/8.6.12-GCCcore-11.3.0 (Tcl-8.6.12-GCCcore-11.3.0.eb)
* SQLite/3.38.3-GCCcore-11.3.0 (SQLite-3.38.3-GCCcore-11.3.0.eb)
* Python/3.10.4-GCCcore-11.3.0-bare (Python-3.10.4-GCCcore-11.3.0-bare.eb)
* BLIS/0.9.0-GCC-11.3.0 (BLIS-0.9.0-GCC-11.3.0.eb)
* OpenBLAS/0.3.20-GCC-11.3.0 (OpenBLAS-0.3.20-GCC-11.3.0.eb)
* FlexiBLAS/3.2.0-GCC-11.3.0 (FlexiBLAS-3.2.0-GCC-11.3.0.eb)
* gompi/2022a (gompi-2022a.eb)
* FFTW.MPI/3.3.10-gompi-2022a (FFTW.MPI-3.3.10-gompi-2022a.eb)
* ScaLAPACK/2.2.0-gompi-2022a-fb (ScaLAPACK-2.2.0-gompi-2022a-fb.eb)
* foss/2022a (foss-2022a.eb)

PR #317, opened 2023-08-22
script to more easily inspect a build job

trz42:
Provides the script `bot/inspect.sh` which should be called with the full path to a tarball containing the state of a build job's temporary storage.

Before calling the script, launch an interactive job, e.g.,

```bash
srun --time=8:00:00 -C shape=c4.2xlarge --pty /bin/bash
```

Then do the following to enter the build job's environment

```bash
./bot/inspect.sh --resume /mnt/shared/home/bot/eessi-bot-software-layer/jobs/2023.07/pr_304/5793/previous_tmp/build_step/eessi-2023.06-software-1689005270.tgz
```

This follows all the setup done for a build job until the step where `eb --easystack ...` is called. A message such as

```
Ready for inspection of build job:
 - job directory is /eessi_bot_job ($HOME), check for slurm-*.out file
 - temporary data of job available at /tmp
 - Note, prefix /cvmfs/pilot.nessi.no/versions/2023.06 is writable
 - EasyBuild v4.7.2 is available
```

is printed when it's ready to be used.

Testing & feedback welcome. Would like to keep it as WIP for a few days to allow for polishing and improvements.


casparvl:
I've played with this a bit, but wanted to do so in my own homedir. I think it would be good to see what the minimal 'unit' is we need to give to someone to make this work. From my attempt in https://github.com/EESSI/software-layer/pull/310#issuecomment-1678947443 I'm assuming it needs the tarball, and the config dir.

The issue is that in this script, you make the assumption that the complete build tree looks the same. E.g.

```
job_dir=$(dirname $(dirname $(dirname ${resume_tgz})))
```
It would be nice if that is just configurable. Than we (people who have access to the bot account) could simply give others (without access) only the necessary things: config dir & tarball. And from there, they could resume.

Though I haven't yet had the time to read through the code in detail, I expect the only thing this would need is an argument to point to the location of the `cfg` dir. 

But maybe it's not so easy...?

casparvl:
Ok, let me log what I do here, just to have written down somewhere how I debug https://github.com/EESSI/software-layer/pull/310 as an example:

- Check out your feature branch from the repo, and in that checkout run `./bot/inspect.sh --resume /mnt/shared/home/bot/eessi-bot-software-layer/jobs/2023.08/pr_310/6355/previous_tmp/build_step/eessi-2023.06-software-1691495231.tgz`
- I checked the EasyBuild logs at `/tmp/eb-vnem_h5_/easybuild-1ni1a4ed.log`. This requires a bit of searching, but they are always in `/tmp/eb-*`, so should be doable to find them.
- Dump the environment script for the failing software `eb FFTW.MPI-3.3.10-gompi-2022a.eb --dump-env-script`, as it is failing in the test suite. Similarly, if I wanted to mimic the build, I could have just run  `eb FFTW.MPI-3.3.10-gompi-2022a.eb` to mimic the build, or `eb eessi-2023.06-eb-4.7.2-2022a.yml` (though this may take a bit longer as it needs to figure out what was installed already).
- Check which test step command failed, e.g. from the bottom we search for `Starting test step` and find the command that is being run:
```
running cmd  export OMPI_MCA_rmaps_base_oversubscribe=true &&   make check  (in /tmp/bot/easybuild/build/FFTWMPI/3.3.10/gompi-2022a/fftw-3.3.10)
```
- `cd /tmp/bot/easybuild/build/FFTWMPI/3.3.10/gompi-2022a/fftw-3.3.10`, `source FFTW.MPI-3.3.10-gompi-2022a.env` and run the test command `OMPI_MCA_rmaps_base_oversubscribe=true &&   make check`

And, of course, everything runs like a charm now...

bot/inspect.sh
```diff
@@ -0,0 +1,430 @@
...
+# ignore current $MODULEPATH entirely
+module unuse $MODULEPATH
+module use $EASYBUILD_INSTALLPATH/modules/all
+if [[ -z ${MODULEPATH} ]]; then
+    fatal_error "Failed to set up \$MODULEPATH?!"
+else
+    echo_green ">> MODULEPATH set up: ${MODULEPATH}"
+fi
+
+eb_version='4.7.2'
```
boegel, line 393: Ideally this is an option to the script, since we'll soon be using a newer EasyBuild version for `2023.06`.

Maybe we shouldn't be loading an EasyBuild module at all here (at least by default).

casparvl (commented):
As mentioned, it might be useful if one can also drive this script in a batch job. For that to be useful, we need to be able to run commands in the container. I added:

```
  echo "  -c | --command COMMAND -  command to execute inside the container, in the prefix environment"
...
    -c|--command)
      export run_in_prefix="${2}"
      shift 2
      ;;
...
./eessi_container.sh "${CMDLINE_ARGS[@]}" \
                     -- ${EESSI_COMPAT_LAYER_DIR}/startprefix <<< ${run_in_prefix}
```
which works, but I haven't tested how it behaves if no `--command` is passed. Might have to do some more elegant if-else there to avoid passing in `${run_in_prefix}` if it isn't defined.

bot/inspect.sh
trz42: Agree. Should not be set to a specific version because different ones may be used by the same job.

I think it would be best to skip this, and change the comments with instructions to load EasyBuild.

casparvl (approved):
I've used this a couple of times by now. Works like a charm for me. Lgtm!

PR #341, opened 2023-09-26
set `shared_fs_path` for bot on AWS CitC Slurm cluster

boegel:
using `shared` subdirectory in home directory of bot on AWS CitC cluster so build script can use `easybuild/sources` in there as EasyBuild sourcepath which is shared across all build jobs


bedroge (approved):
lgtm

PR #345, opened 2023-09-28
only increase limit for numerical test failures for OpenBLAS for `aarch64/neoverse_v1`

boegel:
The hook that relaxes the max. number of numerical test failures for OpenBLAS was a bit too forgiving, since for both `aarch64/generic` and `aarch64/neoverse_n1`, since there we're only seeing 75 numerical test failures for OpenBLAS 0.3.18 and 0.3.20, and only 55 failures for OpenBLAS 0.3.21, so well below the default limit of 150.

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -166,7 +166,7 @@ def parse_hook_openblas_relax_lapack_tests_num_errors(ec, eprefix):
     cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
     if ec.name == 'OpenBLAS':
         # relax maximum number of failed numerical LAPACK tests for aarch64/neoverse_v1 CPU target
-        # since the default setting of 150 that works well on other aarch64 target and x86_64 is a bit too strict
+        # since the default setting of 150 that works well on other aarch64 targets and x86_64 is a bit too strict
         # See https://github.com/EESSI/software-layer/issues/314
         cfg_option = 'max_failing_lapack_tests_num_errors'
         if cpu_target == CPU_TARGET_NEOVERSE_V1:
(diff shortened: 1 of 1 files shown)


eb_hooks.py
```diff
@@ -160,13 +162,14 @@ def parse_hook_fontconfig_add_fonts(ec, eprefix):
 
 
 def parse_hook_openblas_relax_lapack_tests_num_errors(ec, eprefix):
-    """Relax number of failing numerical LAPACK tests on Arm 64-bit systems."""
+    """Relax number of failing numerical LAPACK tests on for aarch64/neoverse_v1 CPU target."""
```
bedroge, line 165:
```suggestion
    """Relax number of failing numerical LAPACK tests for aarch64/neoverse_v1 CPU target."""
```

eb_hooks.py
```diff
@@ -160,13 +162,14 @@ def parse_hook_fontconfig_add_fonts(ec, eprefix):
...
+    cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
     if ec.name == 'OpenBLAS':
-        # relax maximum number of failed numerical LAPACK tests on Arm 64-bit systems,
-        # since the default setting of 150 that works well on x86_64 is a bit too strict
+        # relax maximum number of failed numerical LAPACK tests for aarch64/neoverse_v1 CPU target
+        # since the default setting of 150 that works well on other aarch64 target and x86_64 is a bit too strict
```
bedroge, line 169:
```suggestion
        # since the default setting of 150 that works well on other aarch64 targets and x86_64 is a bit too strict
```

bedroge (changes requested):

bedroge (approved):

PR #351, opened 2023-10-03
determine easystack files to process via PR patch file

boegel:
We're starting to hit GitHub rate limits more often because we're using `--from-pr` and `--include-easyblocks-from-pr` in various easystack files now.

Currently, the `EESSI-pilot-install-software.sh` script that is run by `bot/build.sh` script simply iterates over all existing easystack files, regardless of whether or not they have been changed, and thus triggers a whole bunch of useless EasyBuild sessions that hammer the GitHub API.

These changes use the `<PR>.diff` files that is left behind in the working directory that is prepared by the bot to only process the easystack files that are actually changed in a PR, which significantly limits the EasyBuild sessions being started.

I think it would be better if the bot collects the info on which files have been changed/added/removed by a PR, but for now this "dirty" approach would already help a lot, and can be replaced with a cleaner approach later if the bot feeds in a list of changed/removed/added files via the `cfg/job.cfg` file it prepares for each build job.

tested via https://github.com/boegel/software-layer/pull/28, works as intended

### DIFF
--- a/EESSI-pilot-install-software.sh
+++ b/EESSI-pilot-install-software.sh
@@ -186,11 +186,8 @@ else
     echo_green ">> MODULEPATH set up: ${MODULEPATH}"
 fi
 
-# assume there's only one diff file that corresponds to the PR patch file
-pr_diff=$(ls [0-9]*.diff | head -1)
-
-# use PR patch file to determine in which easystack files stuff was added
-for easystack_file in $(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^eessi.*yml$'); do
+# use 'git diff' to determine which easystack files were changed
+for easystack_file in $(git diff --name-only | grep '^eessi.*yml$'); do
 
     echo -e "Processing easystack file ${easystack_file}...\n\n"
 
--- a/EESSI-pilot-install-software.sh
+++ b/EESSI-pilot-install-software.sh
@@ -41,7 +41,7 @@ function copy_build_log() {
         echo "- working directory: ${PWD}" >> ${build_log_path}
         echo "- Slurm job ID: ${SLURM_OUT}" >> ${build_log_path}
         echo "- EasyBuild version: ${eb_version}" >> ${build_log_path}
-        echo "- easystack file: ${es}" >> ${build_log_path}
+        echo "- easystack file: ${easystack_file}" >> ${build_log_path}
 
         echo "EasyBuild log file ${build_log} copied to ${build_log_path} (with context appended)"
     fi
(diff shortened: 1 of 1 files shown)


EESSI-pilot-install-software.sh
```diff
@@ -186,36 +186,42 @@ else
     echo_green ">> MODULEPATH set up: ${MODULEPATH}"
 fi
 
-for eb_version in '4.7.2' '4.8.0' '4.8.1'; do
+# assume there's only one diff file that corresponds to the PR patch file
+pr_diff=$(ls [0-9]*.diff | head -1)
+
+# use PR patch file to determine in which easystack files stuff was added
+for easystack_file in $(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^eessi.*yml$'); do
```
trz42, line 193: That's ok. However, if we can assume that `git` is available we could also just do

```bash
for easystack_file in $(git diff --name-only | grep '^eessi.*yml$'); do
```

EESSI-pilot-install-software.sh
```diff
@@ -186,36 +186,42 @@ else
...
+
+    echo -e "Processing easystack file ${easystack_file}...\n\n"
+
+    # determine version of EasyBuild module to load based on EasyBuild version included in name of easystack file
+    eb_version=$(echo ${easystack_file} | sed 's/.*eb-\([0-9.]*\).*/\1/g')
```
trz42, line 198: Alternative way might be

```bash
eb_version=$(echo ${easystack_file} | cut -f 4 -d '-')
```

trz42 (approved):
Looks good. Two minor suggestions. Feel free to ignore them.

EESSI-pilot-install-software.sh
boegel: Wouldn't `git diff` also list files that were removed?
With `+++` I'm making sure we only get files where stuff was added (so the file will exist).

May not matter much in practice, as long as we don't remove easystack files (I think we may do that when rebuilding for `eessi.io`, but we could do a cleanup PR first to collapse easystack files to latest EasyBuild release, and then follow up with "build" PRs for `eessi.io`).

So I'll switch to this, it's a lot cleaner (and hopefully it's only temporary, since I would really prefer getting this info from the bot instead).

EESSI-pilot-install-software.sh
boegel: That's a bit less strict, at least with `eb-` we have some guarantee that which follows is an EasyBuild version...

So I prefer sticking to the `sed`.

Eventually we probably want to specify the EasyBuild version to use *in* the easystack file itself (but then EasyBuild needs to allow for that, I think).

EESSI-pilot-install-software.sh
boegel: fixed in [4657384](https://github.com/EESSI/software-layer/pull/351/commits/46573849c7030c9ec9c2d6b9baa7cc92ec48b7ab)

EESSI-pilot-install-software.sh
trz42: ack.

trz42 (approved):
Looks good.

PR #354, opened 2023-10-04
{2023.06} get changed/new files from PR.diff file

trz42:
Reverts method to determine which easystack files have been changed back to original commit in #351


boegel (approved):

boegel:
I've started looking into a better way to do this on the bot side, but haven't gotten very far yet, so this is good enough in the short term...

PR #359, opened 2023-10-06
simplify CI workflow to check for missing installations, just loop over easystack files per CPU target

boegel:
This bring down the number of jobs to check the easystack files against what's currently installed in EESSI down from 104 currently (13 easystack files times 8 CPU targets) to just 8 (one per CPU target).


bedroge (approved):
Looks good to me.

PR #366, opened 2023-10-15
[WIP] add scripts to run test suite after build job has finished

trz42:
First outline of how testing could be added to build-and-deploy procedure. See also accompanying PR for the bot (https://github.com/EESSI/eessi-bot-software-layer/pull/222).

This PR adds three scripts:

- `bot/test.sh` prepares the environment for running the tests inside the build container
  - because it runs on the same node (in fact, the same job) that was building software packages, it tries to re-use the existing temporary storage (e.g., cached CernVM-FS data, and particularly the fuse-overlayfs upper directory): directly or from a tarball of the build step
  - alternatively, it might unpack or bind mount built software from the tarball containing the software installation
  - the script is largely similar to `bot/build.sh` and `bot/inspect.sh` (for the latter see #317)
- `run_tests.sh` is just a small script that runs `test_suite.sh` in a compat layer environment (very similar to `install_software_layer.sh`)
- `test_suite.sh` runs some tests
  - at the moment it runs `eb --sanity-check-only` for all easyconfigs that were "added" to an easystack file (by design, it will also process easyconfigs whose line has changed)
  - only changed easystack files are processed
  - the script is very similar to `EESSI-pilot-install-software.sh`

TODO:
- [ ] test that it works
- [ ] add script `bot/check-test.sh` that processes the result of the test(s) and produces and update to be added to the PR comment (job status table); this is likely somewhat similar to `bot/check-build.sh`

### DIFF
--- a/bot/test.sh
+++ b/bot/test.sh
@@ -206,9 +206,9 @@ fi
 
 # prepare arguments to test_suite.sh (specific to test step)
 declare -a TEST_SUITE_ARGS=()
-# if [[ ${EESSI_SOFTWARE_SUBDIR_OVERRIDE} =~ .*/generic$ ]]; then
-#     TEST_SUITE_ARGS+=("--generic")
-# fi
+if [[ ${EESSI_SOFTWARE_SUBDIR_OVERRIDE} =~ .*/generic$ ]]; then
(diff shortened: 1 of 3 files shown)


boegel:
note: got merged via larger PR #467 that includes these commits

PR #367, opened 2023-10-18
Allow sending additional args to singularity in build script

ocaisa:
This does two things, it allows you to send additional options to `singularity` (like `--nv`), and it adds an `/opt/eessi` directory to the container which the default location for `host_injections` installations.


trz42:
Should we add these parameters in some way also to `eessi_container.sh`?

The bot uses `eessi_container.sh` for building. Also, other scripts such as `bot/inspect.sh` use it.

ocaisa:
Urgh, this is confusing, what's `build_container.sh` for if not for building?

ocaisa:
Closing this in favour of the changes in #368

PR #368, opened 2023-10-18
Build CUDA under `host_injections` and make EESSI aware of host CUDA drivers

ocaisa:


ocaisa:
This can be tested by starting the build container from the PR and running, e.g., 
```
source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash 
./gpu_support/nvidia/install_cuda_host_injections.sh 12.1.1
```

ocaisa:
But I need to transition the PR to (also?) use `eessi_container.sh`

ocaisa:
This PR can be tested without GPUs, use
```
./eessi_container.sh -m shell --nvidia install -v
```
to start the build container with the necessary settings, then you can install CUDA (under `host_injections`) with
```
source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash 
./gpu_support/nvidia/install_cuda_host_injections.sh 12.1.1
```

@trz42 There are some sizable changes to the build container in the PR so you may wish to take it for a test drive.

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
casparvl, line 1: Should add some comment here on what this script is meant for. E.g.

```
This script can be used to install CUDA in the host_injections directory. This provides the parts of the CUDA installation that can not be redistributed as part of EESSI due to license limitations. While GPU-based software from EESSI will run without these, installation of additional CUDA software requires the CUDA installations in host_injections to be present.

The host_injections directory is a variant symlink that points to `/opt/eessi`, unless otherwise defined in the local CVMFS configuration (see https://cvmfs.readthedocs.io/en/stable/cpt-repo.html#variant-symlinks). For the installation to be succesful, this directory needs to be writeable to the user executing this script.
```

scripts/utils.sh
```diff
@@ -32,11 +32,81 @@ function check_exit_code {
...
+  # Now do the actual evaluation
+  return_code=$ANY_ERROR_EXITCODE
+  result=$(echo "$1" "$2" | awk '{if ($1 > $2) print "true"}')
+  if [ "$result" = true ] ; then
+    return_code=0
+  fi
+  return $return_code
+}
+
+function check_in_prefix_shell() {
```
casparvl, line 67: Seems unused in this PR?

scripts/utils.sh
```diff
@@ -32,11 +32,81 @@ function check_exit_code {
...
 
+function check_eessi_initialised() {
+  if [[ -z "${EESSI_SOFTWARE_PATH}" ]]; then
+    fatal_error "EESSI has not been initialised!"
+  else
+    return 0
+  fi
+}
+
+function float_greater_than() {
```
casparvl, line 43: Seems unused in this PR?

eessi_container.sh
```diff
@@ -394,12 +432,30 @@ fi
...
+if [[ ${SETUP_NVIDIA} -eq 1 ]]; then
+    if [[ "${NVIDIA_MODE}" == "run" || "${NVIDIA_MODE}" == "all" ]]; then
+        # Give singularity the appropriate flag
+        ADDITIONAL_SINGULARITY_FLAGS="--nv ${ADDITIONAL_SINGULARITY_FLAGS}"
+        [[ ${VERBOSE} -eq 1 ]] && echo "ADDITIONAL_SINGULARITY_FLAGS=${ADDITIONAL_SINGULARITY_FLAGS}"
+    fi
+    if [[ "${NVIDIA_MODE}" == "install" || "${NVIDIA_MODE}" == "all" ]]; then
+        # Add additional bind mounts to allow CUDA to install within a container
+        EESSI_VAR_LOG=${EESSI_TMPDIR}/var-log
+        EESSI_USR_LOCAL_CUDA=${EESSI_TMPDIR}/usr-local-cuda
```
casparvl, line 451: Why are these needed? Is this just because the installation of CUDA needs to paths to be available and/or writeable? maybe good to include in the comment above.

casparvl (changes requested):

eessi_container.sh
ocaisa: The true answer is only something NVIDIA can now, these paths were decided based on trial and error

casparvl: I figured you might have gotten a `permission denied` or something on these paths before, and that's why you decided to bind-mound them - if that's the case it'd be good to include that. But if you don't really know why they are needed, there's not much we can do :)

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,92 @@
...
+    error="Need at least ${required_space_in_tmpdir} disk space under ${tmpdir}.\n"
+    error="${error}Set the environment variable CUDA_TEMP_DIR to a location with adequate space to pass this check."
+    error="${error}You can alternatively set EASYBUILD_BUILDPATH and/or EASYBUILD_SOURCEPATH "
+    error="${error}to reduce this requirement. Exiting now..."
+    fatal_error "${error}"
+  fi
+
+  if [[ -z "${EBROOTEASYBUILD}" ]]; then
+    echo_yellow "Loading EasyBuild module to do actual install"
+    module load EasyBuild
```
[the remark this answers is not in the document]
ocaisa: I've added error catching in https://github.com/EESSI/software-layer/pull/368/commits/090351864cbc5b97f35498b278d16041a5b97b1e and made a few more updates to how we check for `eb`

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,92 @@
...
+  fi
+
+  # we need the --rebuild option and a (random) dir for the module since we are
```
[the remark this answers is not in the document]
ocaisa: Added in https://github.com/EESSI/software-layer/pull/368/commits/090351864cbc5b97f35498b278d16041a5b97b1e

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,92 @@
...
+
+# Initialise our bash functions
+TOPDIR=$(dirname $(realpath $BASH_SOURCE))
+source "$TOPDIR"/../../scripts/utils.sh
+
+# Make sure EESSI is initialised
+check_eessi_initialised
+
+if [[ $# -eq 0 ]] ; then
+    fatal_error "You must provide the CUDA version as an argument, e.g.:\n $0 11.3.1"
```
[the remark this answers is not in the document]
ocaisa: I've added changes in https://github.com/EESSI/software-layer/pull/368/commits/090351864cbc5b97f35498b278d16041a5b97b1e that should help people use the right version here (the one that mirrors what they have under EESSI) and does a good bit of related error checking

scripts/utils.sh
ocaisa: I've left it in place as it is small and useful in later PRs

eessi_container.sh
ocaisa: I've added a clarification comment in https://github.com/EESSI/software-layer/pull/368/commits/090351864cbc5b97f35498b278d16041a5b97b1e

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,92 @@
...
+  fi
+
+  # we need the --rebuild option and a (random) dir for the module since we are
+  # fixing the broken links of the EESSI-shipped installation
+  extra_args="--rebuild --installpath-modules=${tmpdir}"
+
+  # We don't want hooks used in this install, we need a vanilla CUDA installation
+  touch "$tmpdir"/none.py
+  # shellcheck disable=SC2086  # Intended splitting of extra_args
+  eb --prefix="$tmpdir" ${extra_args} --hooks="$tmpdir"/none.py --installpath="${cuda_install_parent}"/ CUDA-"${install_cuda_version}".eb
```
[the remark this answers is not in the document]
ocaisa: I didn't try this either but the current approach isn't that dirty in my opinion (and I know it works!)

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,198 @@
...
+  # since it will provide the symlinked targets for the parts of the CUDA
+  # installation in the `.../versions/...` prefix
+  # We install the module in our `tmpdir` since we do not need the modulefile,
+  # we only care about providing the targets for the symlinks.
+  extra_args="--rebuild --installpath-modules=${tmpdir}"
+
+  # We don't want hooks used in this install, we need a vanilla CUDA installation
+  touch "$tmpdir"/none.py
+  # shellcheck disable=SC2086  # Intended splitting of extra_args
+  eb --prefix="$tmpdir" ${extra_args} --hooks="$tmpdir"/none.py --installpath="${cuda_install_parent}"/ "${cuda_easyconfig}"
```
ocaisa, line 189: @boegel Why don't they have to accept the terms of the EULA here? I think we probably need to enforce that they explicitly do this

ocaisa:
@casparvl I've addressed your review as much as I can and I've added better command line options, including a critical acceptance of the CUDA EULA. This should be read for showtime now

gpu_support/nvidia/install_cuda_host_injections.sh
ocaisa: Implemented in https://github.com/EESSI/software-layer/pull/368/commits/12719eccaabf57112332cc4a4958d0d03e382ef4

ocaisa:
I think I'll fold the script to the linking to the host drivers in here as well...working on it now

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,213 @@
...
+        return 0  # Return success (0) if it's a semantic version
+    else
+        return 1  # Return failure (1) if it's not a semantic version
+    fi
+}
+if ! is_semantic_version "$install_cuda_version"; then
+  show_help
+  error="\nYou must provide a semantic version for CUDA (e.g., 12.1.1) via the appropriate\n"
+  error="${error}command line option. This script is intended for use with EESSI so the 'correct'\n"
+  error="${error}version to provide is probably the one that is available under\n"
```
casparvl, line 92: Tiny detail: it might not be one. I guess it's a completely valid use case to have a CUDA installed in `host_injections` for _each_ cuda in `$EESSI_SOFTWARE_PATH/software/CUDA`. Not sure how I would phrase that here more clearly though... I'm also ok with leaving it like this.

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,213 @@
...
+  error="${error}$EESSI_SOFTWARE_PATH/software/CUDA\n"
```
casparvl, line 93: Another small detail: you use `$EESSI_SOFTWARE_PATH` here in the error message, before checking it exists (that is currently on line 105 and 107). You could move up the check for `$EESSI_SOFTWARE_PATH` before this semantic-version check.

gpu_support/nvidia/install_cuda_host_injections.sh
```diff
@@ -0,0 +1,213 @@
...
+if [ "$eula_accepted" -ne 1 ]; then
+  show_help
+  error="\nYou _must_ accept the CUDA EULA via the appropriate command line option.\n"
+  fatal_error "${error}"
+fi
+
+# Make sure EESSI is initialised
+check_eessi_initialised
+
+if [[ -z "${EESSI_SOFTWARE_PATH}" ]]; then
```
casparvl, line 107: Isn't that what `check_eessi_initialised` already checks? I believe that even exits upon failure, so you wouldn't ever make it to this point if `EESSI_SOFTWARE_PATH`  wasn't defined, right?

gpu_support/nvidia/install_cuda_host_injections.sh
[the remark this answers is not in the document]
casparvl: I tested it, it doesn't work with `--hooks=None`, so good thing you kept this :)

gpu_support/nvidia/install_cuda_host_injections.sh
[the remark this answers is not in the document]
casparvl: Lgtm

gpu_support/nvidia/link_nvidia_host_libraries.sh
```diff
@@ -0,0 +1,133 @@
...
+# We cannot use a version of ldconfig that's being shipped under CVMFS
+exclude_prefix="/cvmfs"
+
+found_paths=()
+# Always attempt to use /sbin/ldconfig
+if [ -x "/sbin/$command_name" ]; then
+    found_paths+=("/sbin/$command_name")
+fi
+IFS=':' read -ra path_dirs <<< "$PATH"
+for dir in "${path_dirs[@]}"; do
```
casparvl, line 21: As first part of this loop, do:
```
    if [ "$dir" = "/sbin" ]; then
        continue  # we've already checked for $command_name in /sbin, don't need to do it twice
    fi
```
Otherwise, if `/sbin` is also on `PATH`, `/sbin/ldconfig` gets added twice to `found_paths`:
```
Found ldconfig in the following locations:
- /sbin/ldconfig
- /sbin/ldconfig
```

gpu_support/nvidia/link_nvidia_host_libraries.sh
```diff
@@ -0,0 +1,133 @@
...
+  # Inject driver and CUDA versions into dir
+  echo $host_driver_version > driver_version.txt
+  echo $host_cuda_version > cuda_version.txt
+  drivers_linked=1
+
+  # Remove the temporary directory when done
+  rm -r "$temp_dir"
+fi
+
+# Make latest symlink for NVIDIA drivers
```
casparvl, line 106: I'm actually wondering: what is the point of the `latest` symlink here? You're doing a check that if the drivers don't match the linked drivers, you do it again. So there is only ever going to be _one_ copy of the driver symlinks here, right?

gpu_support/nvidia/link_nvidia_host_libraries.sh
```diff
@@ -0,0 +1,133 @@
...
+    ln -s host latest
+fi
+
+# Make sure the libraries can be found by the EESSI linker
+host_injection_linker_dir=${EESSI_EPREFIX/versions/host_injections}
+if [ -L "$host_injection_linker_dir/lib" ]; then
+  target_path=$(readlink -f "$host_injection_linker_dir/lib")
+  if [ "$target_path" != "$$host_injections_nvidia_dir/latest" ]; then
+    cd $host_injection_linker_dir
+    ln -sf $host_injections_nvidia_dir/latest lib
```
casparvl, line 125: I'm wondering if this is the way to go. It'll create a link like:
```
/cvmfs/pilot.eessi-hpc.org/host_injections/2023.06/compat/linux/x86_64/lib -> /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/x86_64/latest
```
But what if we want to inject other libraries to be picked up later? If `/cvmfs/pilot.eessi-hpc.org/host_injections/2023.06/compat/linux/x86_64/lib` is a symlink, it can only point one way.

It brings me to the broader question: why do we even install this in `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/x86_64/latest` first? Why not directly create symlinks in `/cvmfs/pilot.eessi-hpc.org/host_injections/2023.06/compat/linux/x86_64/lib`? You probably had a good reason for it I assume, but it's not immediately clear to me :)

casparvl (changes requested):

gpu_support/nvidia/link_nvidia_host_libraries.sh
ocaisa: Not once we support compatibility libraries

ocaisa: In that scenario we would have a script that figures out which version of the CUDA compat libraries that we ship work with the driver and then have `latest` point to that

gpu_support/nvidia/link_nvidia_host_libraries.sh
ocaisa: You're right, this is not the way to go, but it is what is supported right now by the EESSI linker. What I would like to do in the next compat layer is have a few additional trusted directories:
* a generic one as is right now
* one for nvidia
* one for AMD

It's better not to put symlinks in the trusted dir directly so that when we have a global update of the drivers, we only have to make a single update to the symlinks (rather than for every version of EESSI)

gpu_support/nvidia/install_cuda_host_injections.sh
ocaisa: Rephrased in https://github.com/EESSI/software-layer/pull/368/commits/480d35643c7740c48f5508303e1cccadf04aae76

gpu_support/nvidia/install_cuda_host_injections.sh
ocaisa: You're right, simplified in https://github.com/EESSI/software-layer/pull/368/commits/480d35643c7740c48f5508303e1cccadf04aae76

gpu_support/nvidia/link_nvidia_host_libraries.sh
ocaisa: Adopted in https://github.com/EESSI/software-layer/pull/368/commits/480d35643c7740c48f5508303e1cccadf04aae76

gpu_support/nvidia/link_nvidia_host_libraries.sh
casparvl: Ah right, I didn't think about the compat libs here. Makes sense

PR #381, opened 2023-11-09
Install CUDA and CUDA-Samples via the bot

ocaisa:
Requires #368


ocaisa:
Unfortunately this is tripping over the issue with removing software from the overlay:
```
  File "/home/rocky/software-layer/eb_hooks.py", line 380, in post_package_cuda
    os.remove(source)
PermissionError: [Errno 13] Permission denied: '/cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/intel/skylake_avx512/software/CUDA/12.1.1/version.json'
```

eessi-2023.06-eb-4.8.2-2023a.yml
```diff
@@ -0,0 +1,2 @@
+easyconfigs:
+  - CUDA-Samples-12.1-GCC-12.3.0-CUDA-12.1.1.eb
```
boegel, line 2: We need to use the easyconfig from https://github.com/easybuilders/easybuild-easyconfigs/pull/19189 here

boegel (changes requested):

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+        local arch = os.getenv("EESSI_CPU_FAMILY") or ""
+        local cudaVersionFile = "/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/" .. arch .. "/latest/version.txt"
+        local cudaDriverExists = exists(cudaVersionFile)
+        local singularityCudaExists = exists("/.singularity.d/libs/libcuda.so")
+        if not (cudaDriverExists or singularityCudaExists)  then
+            io.stderr:write("You requested to load ",simpleName,"\\n")
+            io.stderr:write("which relies on the CUDA runtime environment and its compatibility libraries.\\n")
+            io.stderr:write("In order to be able to use the module, please follow the instructions in the\\n")
+            io.stderr:write("gpu_support folder. Installing the needed compatibility libraries can be as easy as:\\n")
+            io.stderr:write("./add_nvidia_gpu_support.sh\\n")
```
boegel, line 93: This should be reworded a bit, there is no script `add_nvidia_gpu_support.sh` yet, and the compat libraries are only needed if the GPU driver is too old.

boegel (changes requested):

eessi-2023.06-eb-4.8.2-2023a.yml
boegel:
```suggestion
  - CUDA-Samples-12.1-GCC-12.3.0-CUDA-12.1.1.eb:
       # use easyconfig that only install subset of CUDA samples,
       # to circumvent problem with nvcc linking to glibc of host OS;
       # see https://github.com/easybuilders/easybuild-easyconfigs/pull/19189
       options:
         from-pr: 19189

File: EESSI-pilot-install-software.sh [hunk 3]
```diff
@@ -229,7 +229,7 @@ done
 
 echo ">> Creating/updating Lmod cache..."
 export LMOD_RC="${EASYBUILD_INSTALLPATH}/.lmod/lmodrc.lua"
-if [ ! -f $LMOD_RC ]; then
+if [ ! -f $LMOD_RC ] || 'create_lmodrc.py' == $(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^create_lmodrc.py$' | egrep -v 'known-issues|missing'); then
```
boegel, line 232: Syntax is not correct here, it's missing `[ ... ]` after the `||`

Also, to make this a bit more readable, I suggest this approach:

```suggestion
lmodrc_changed=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^create_lmodrc.py$' > /dev/null; echo $?)
if [ ! -f $LMOD_RC ] || [ ${lmodrc_changed} == '0' ]; then
```

Not too happy with the `== '0'` bit to indicate that `create_lmodrc.py` was changed, but it'll do the job...

boegel (changes requested):

create_lmodrc.py
boegel: I would reword this to point to https://eessi.io/docs/gpu for now, see also https://github.com/EESSI/docs/pull/128

ocaisa:
To actually test the final build you will need to have the drivers in place. The below should/may work and at least give you an idea of what is required (PR in the works):
```
# Change directory to a location under host_injections
mkdir -p /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/host
cd /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/host

# Gather libraries on the host (_must_ be host ldconfig)
ldconfig -p | awk '{print $NF}' > libs.txt
# Allow for the fact that we may be in a container
ls /.singularity.d/libs/* >> libs.txt

# Link the relevant libraries
curl -O https://raw.githubusercontent.com/apptainer/apptainer/main/etc/nvliblist.conf
grep '.so$' nvliblist.conf | xargs -i grep {} libs.txt | xargs -i ln -s {}
# Inject CUDA version into dir
nvidia-smi --query-gpu=driver_version --format=csv,noheader | tail -n1 > version.txt

# Make latest symlink for NVIDIA drivers
cd ..
ln -s host latest

# Make sure the libraries can be found by the EESSI linker
source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash
host_injection_linker_dir=${EESSI_EPREFIX/versions/host_injections}
mkdir -p $host_injection_linker_dir
cd $host_injection_linker_dir
ln -s /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest libs 
```

boegel:
```bash
function find_host_ldconfig () {
    if [ ! -z ${EESSI_HOST_LDCONFIG} ]; then
        echo ${EESSI_HOST_LDCONFIG}
    else
        if [ -f /sbin/ldconfig ]; then
            echo /u/bin/ldconfig
        elif [ -f /usr/sbin/ldconfig ]; then
            echo /usr/sbin/ldconfig
        else
             echo "This is weird, you should set ${EESSI_HOST_LDCONFIG} (and a support issue)" >&2
             exit 1
        fi
    fi
}

```

on system without CUDA:

```bash
$ nvidia_smi_out=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader 2> nvidia-smi.errors)
$ echo $?
127
$ cat nvidia-smi.errors
-bash: nvidia-smi: command not found
```

casparvl:
I'm in the EESSI container and ran everything in https://github.com/EESSI/software-layer/pull/381#issuecomment-1831761949 except the `nvidia-smi`. That still gives me:
```
 $ nvidia-smi
NVIDIA-SMI couldn't find libnvidia-ml.so library in your system. Please make sure that the NVIDIA Display Driver is properly installed and present in your system.
Please also try adding directory that contains libnvidia-ml.so to your system PATH.
```

I now have:
```
[EESSI pilot 2023.06] $ ls -al /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest
lrwxrwxrwx 1 casparl casparl 4 Nov 29 15:06 /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest -> host
[EESSI pilot 2023.06] $ ls -al /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/
total 39
drwxr-x--- 2 casparl casparl  4096 Nov 29 15:06 .
drwxr-x--- 3 casparl casparl  4096 Nov 29 15:06 ..
lrwxrwxrwx 1 casparl casparl    30 Nov 29 15:06 libEGL.so -> /.singularity.d/libs/libEGL.so
lrwxrwxrwx 1 casparl casparl    32 Nov 29 15:06 libEGL.so.1 -> /.singularity.d/libs/libEGL.so.1
lrwxrwxrwx 1 casparl casparl    39 Nov 29 15:06 libEGL_nvidia.so.0 -> /.singularity.d/libs/libEGL_nvidia.so.0
lrwxrwxrwx 1 casparl casparl    29 Nov 29 15:06 libGL.so -> /.singularity.d/libs/libGL.so
lrwxrwxrwx 1 casparl casparl    31 Nov 29 15:06 libGL.so.1 -> /.singularity.d/libs/libGL.so.1
lrwxrwxrwx 1 casparl casparl    36 Nov 29 15:06 libGLESv1_CM.so -> /.singularity.d/libs/libGLESv1_CM.so
lrwxrwxrwx 1 casparl casparl    38 Nov 29 15:06 libGLESv1_CM.so.1 -> /.singularity.d/libs/libGLESv1_CM.so.1
lrwxrwxrwx 1 casparl casparl    45 Nov 29 15:06 libGLESv1_CM_nvidia.so.1 -> /.singularity.d/libs/libGLESv1_CM_nvidia.so.1
lrwxrwxrwx 1 casparl casparl    33 Nov 29 15:06 libGLESv2.so -> /.singularity.d/libs/libGLESv2.so
lrwxrwxrwx 1 casparl casparl    35 Nov 29 15:06 libGLESv2.so.2 -> /.singularity.d/libs/libGLESv2.so.2
lrwxrwxrwx 1 casparl casparl    42 Nov 29 15:06 libGLESv2_nvidia.so.2 -> /.singularity.d/libs/libGLESv2_nvidia.so.2
lrwxrwxrwx 1 casparl casparl    30 Nov 29 15:06 libGLX.so -> /.singularity.d/libs/libGLX.so
lrwxrwxrwx 1 casparl casparl    32 Nov 29 15:06 libGLX.so.0 -> /.singularity.d/libs/libGLX.so.0
lrwxrwxrwx 1 casparl casparl    39 Nov 29 15:06 libGLX_nvidia.so.0 -> /.singularity.d/libs/libGLX_nvidia.so.0
lrwxrwxrwx 1 casparl casparl    37 Nov 29 15:06 libGLdispatch.so -> /.singularity.d/libs/libGLdispatch.so
lrwxrwxrwx 1 casparl casparl    39 Nov 29 15:06 libGLdispatch.so.0 -> /.singularity.d/libs/libGLdispatch.so.0
lrwxrwxrwx 1 casparl casparl    35 Nov 29 15:06 libOpenCL.so.1 -> /.singularity.d/libs/libOpenCL.so.1
lrwxrwxrwx 1 casparl casparl    33 Nov 29 15:06 libOpenGL.so -> /.singularity.d/libs/libOpenGL.so
lrwxrwxrwx 1 casparl casparl    35 Nov 29 15:06 libOpenGL.so.0 -> /.singularity.d/libs/libOpenGL.so.0
lrwxrwxrwx 1 casparl casparl    31 Nov 29 15:06 libcuda.so -> /.singularity.d/libs/libcuda.so
lrwxrwxrwx 1 casparl casparl    33 Nov 29 15:06 libcuda.so.1 -> /.singularity.d/libs/libcuda.so.1
lrwxrwxrwx 1 casparl casparl    34 Nov 29 15:06 libnvcuvid.so -> /.singularity.d/libs/libnvcuvid.so
lrwxrwxrwx 1 casparl casparl    36 Nov 29 15:06 libnvcuvid.so.1 -> /.singularity.d/libs/libnvcuvid.so.1
lrwxrwxrwx 1 casparl casparl    37 Nov 29 15:06 libnvidia-cfg.so -> /.singularity.d/libs/libnvidia-cfg.so
lrwxrwxrwx 1 casparl casparl    39 Nov 29 15:06 libnvidia-cfg.so.1 -> /.singularity.d/libs/libnvidia-cfg.so.1
lrwxrwxrwx 1 casparl casparl    47 Nov 29 15:06 libnvidia-egl-wayland.so.1 -> /.singularity.d/libs/libnvidia-egl-wayland.so.1
lrwxrwxrwx 1 casparl casparl    52 Nov 29 15:06 libnvidia-eglcore.so.535.104.12 -> /.singularity.d/libs/libnvidia-eglcore.so.535.104.12
lrwxrwxrwx 1 casparl casparl    40 Nov 29 15:06 libnvidia-encode.so -> /.singularity.d/libs/libnvidia-encode.so
lrwxrwxrwx 1 casparl casparl    42 Nov 29 15:06 libnvidia-encode.so.1 -> /.singularity.d/libs/libnvidia-encode.so.1
lrwxrwxrwx 1 casparl casparl    37 Nov 29 15:06 libnvidia-fbc.so -> /.singularity.d/libs/libnvidia-fbc.so
lrwxrwxrwx 1 casparl casparl    39 Nov 29 15:06 libnvidia-fbc.so.1 -> /.singularity.d/libs/libnvidia-fbc.so.1
lrwxrwxrwx 1 casparl casparl    51 Nov 29 15:06 libnvidia-glcore.so.535.104.12 -> /.singularity.d/libs/libnvidia-glcore.so.535.104.12
lrwxrwxrwx 1 casparl casparl    49 Nov 29 15:06 libnvidia-glsi.so.535.104.12 -> /.singularity.d/libs/libnvidia-glsi.so.535.104.12
lrwxrwxrwx 1 casparl casparl    54 Nov 29 15:06 libnvidia-glvkspirv.so.535.104.12 -> /.singularity.d/libs/libnvidia-glvkspirv.so.535.104.12
lrwxrwxrwx 1 casparl casparl    49 Nov 29 15:06 libnvidia-gtk3.so.535.104.12 -> /.singularity.d/libs/libnvidia-gtk3.so.535.104.12
lrwxrwxrwx 1 casparl casparl    36 Nov 29 15:06 libnvidia-ml.so -> /.singularity.d/libs/libnvidia-ml.so
lrwxrwxrwx 1 casparl casparl    38 Nov 29 15:06 libnvidia-ml.so.1 -> /.singularity.d/libs/libnvidia-ml.so.1
lrwxrwxrwx 1 casparl casparl    42 Nov 29 15:06 libnvidia-opencl.so.1 -> /.singularity.d/libs/libnvidia-opencl.so.1
lrwxrwxrwx 1 casparl casparl    47 Nov 29 15:06 libnvidia-opticalflow.so.1 -> /.singularity.d/libs/libnvidia-opticalflow.so.1
lrwxrwxrwx 1 casparl casparl    48 Nov 29 15:06 libnvidia-ptxjitcompiler.so -> /.singularity.d/libs/libnvidia-ptxjitcompiler.so
lrwxrwxrwx 1 casparl casparl    50 Nov 29 15:06 libnvidia-ptxjitcompiler.so.1 -> /.singularity.d/libs/libnvidia-ptxjitcompiler.so.1
lrwxrwxrwx 1 casparl casparl    51 Nov 29 15:06 libnvidia-rtcore.so.535.104.12 -> /.singularity.d/libs/libnvidia-rtcore.so.535.104.12
lrwxrwxrwx 1 casparl casparl    48 Nov 29 15:06 libnvidia-tls.so.535.104.12 -> /.singularity.d/libs/libnvidia-tls.so.535.104.12
-rw-r----- 1 casparl casparl 10984 Nov 29 15:06 libs.txt
-rw-r----- 1 casparl casparl  1383 Nov 29 15:06 nvliblist.conf
```

casparvl:
Note that
```
$ LD_LIBRARY_PATH=/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest/:$LD_LIBRARY_PATH nvidia-smi
Wed Nov 29 15:15:16 2023
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.12             Driver Version: 535.104.12   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-40GB          On  | 00000000:31:00.0 Off |                  Off |
| N/A   29C    P0              50W / 400W |      4MiB / 40960MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-40GB          On  | 00000000:32:00.0 Off |                  Off |
| N/A   29C    P0              50W / 400W |      4MiB / 40960MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A100-SXM4-40GB          On  | 00000000:CA:00.0 Off |                  Off |
| N/A   29C    P0              49W / 400W |      4MiB / 40960MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   3  NVIDIA A100-SXM4-40GB          On  | 00000000:E3:00.0 Off |                  Off |
| N/A   28C    P0              49W / 400W |      4MiB / 40960MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```
Works fine. I'm a bit confused: how is our runtime linker supposed to pick up on things from `/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest`? Shouldn't we _do_ something to make it look there?

ocaisa:
I updated the last line of the script in my previous comment, it should have been `lib` and not `libs`

casparvl:
Yeah, I used your code from after your last edit, so that isn't the issue here:
```
$ ls -al $host_injection_linker_dir/lib
lrwxrwxrwx 1 casparl casparl 56 Nov 29 15:07 /cvmfs/pilot.eessi-hpc.org/host_injections/2023.06/compat/linux/x86_64/lib -> /cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/latest
```

> $host_injection_linker_dir/lib is a trusted directory for our linker (so any libraries found in there are added to the defaults)

Where / how is that arranged? (then I can check if for me somehow that is _not_ correct)

ocaisa:
If you want to use `nvidia-smi` with our linker, you need to make a copy and set it's linker:
```
cp /usr/bin/nvidia-smi .
patchelf --set-interpreter $EESSI_EPREFIX/lib64/ld-linux-x86-64.so.2 ./nvidia-smi
./nvidia-smi
```
You don't have to do that though, you can just load the CUDA-Samples module and try to run `deviceQuery`

casparvl:
> If you want to use `nvidia-smi` with our linker, you need to make a copy and set it's linker:
> 
> ```
> cp /usr/bin/nvidia-smi .
> patchelf --set-interpreter $EESSI_EPREFIX/lib64/ld-linux-x86-64.so.2 ./nvidia-smi
> ./nvidia-smi
> ```
> 
> You don't have to do that though, you can just load the CUDA-Samples module and try to run `deviceQuery`

Yes, I came to this conclusion as well. Unfortunately, my writeable overlay has disappeared (again), and with it all the contents. So no more `deviceQuery` available. This issue is really really annoying XD I'll try to resume the container, see if that brings it back without having to rebuild...

casparvl:
True, and much easier / foolproof. I meant: just make sure to include that support for running it in the container if you make the PR for that :)

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+    local file = open(path, "rb") -- r read mode and b binary mode
+    if not file then return nil end
+    local content = file:read "*a" -- *a or *all reads the whole file
+    file:close()
+    return content
+end
+
+-- from https://stackoverflow.com/a/40195356
+--- Check if a file or directory exists in this path
+function exists(file)
```
casparvl, line 34: This function didn't work for checking the `libcuda.so` in `/.singularity.d`. I modified it to:
```
function exists(file)
    local ok, err, code = os.rename(file, file)
    if not ok then
        io.stderr:write(string.format("Error opening file %s. Error msg: %s. Error code: %d\n", file, err, code))
        if code == 13 then
            -- Permission denied, but it exists
            return true
        end
    end
    return ok, err
end
```
to figure out what was going on and got:
```
Error opening file /.singularity.d/libs/libcuda.so. Error msg: /.singularity.d/libs/libcuda.so: Read-only file system. Error code: 30
```
So, there are more exit codes that should _probably_ pass. 

The most robust solution might be better to find another way to check if a file exists, but I couldn't find any that doesnt either 1. try to open the file in read mode, 2. tries to rename (as you do) or 3. requires a non-standard LUA packages (`lsf`  which can use `lsf.attributes` apparently to check for file existence, or `luaposix` which can do `stat` calls).

Since all of those are annoying, maybe the best thing to do for now is just add this error code. At least it'll tackle the specific case of this being a read-only dir.

casparvl (changes requested):

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+    if not ok then
+        if code == 13 then
+            -- Permission denied, but it exists
+            return true
+        end
+    end
+    return ok, err
+end
+
+local function visible_hook(modT)
```
casparvl, line 45: I think we should describe what this hook is meant to do. As far as I can tell, the description should be something like:
"This hook checks if the directory $EESSI_CVMFS_REPO/host_injections/nvidia exists. If set up correctly, this directory contains the symlinks to the CUDA drivers on the host machine. This setup is required to make CUDA software from the EESSI stack find the drivers, they cannot run otherwise. If this directory does not exist, the symlinks are not setup, and using CUDA-enabled EESSI modules is not supported on the system. Hence, we hide those modules with this hook, to prevent users from loading modules that cannot work on their machine."

casparvl (changes requested):

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+            io.stderr:write("While the module file exists, the actual software is not shipped with EESSI.\\n")
+            io.stderr:write("In order to be able to use the CUDA module, please follow the instructions in the\\n")
+            io.stderr:write("gpu_support folder. Adding the CUDA software can be as easy as:\\n")
+            io.stderr:write("export INSTALL_CUDA=true && ./add_nvidia_gpu_support.sh\\n")
+            frameStk:__clear()
+        end
+    end
+    -- when loading CUDA enabled modules check if the necessary matching compatibility libraries are installed
+    -- otherwise, refuse to load the requested module and print error message
+    local haveGpu = mt:haveProperty(simpleName,"arch","gpu")
```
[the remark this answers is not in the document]
casparvl: Aaah, I thought this was just something full text that we choose. Ok, makes sense then, we'll deal with it when we get there...

create_lmodrc.py
ocaisa: Given we've added logic to help people know when they can and can't use a module, this bit is useless. Kicking it out

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+        local singularityCudaExists = exists("/.singularity.d/libs/libcuda.so")
+        if not (cudaDriverExists or singularityCudaExists)  then
+            io.stderr:write("You requested to load ",simpleName,"\\n")
+            io.stderr:write("which relies on the CUDA runtime environment and its compatibility libraries.\\n")
+            io.stderr:write("In order to be able to use the module, please follow the instructions in the\\n")
+            io.stderr:write("gpu_support folder. Installing the needed compatibility libraries can be as easy as:\\n")
+            io.stderr:write("./add_nvidia_gpu_support.sh\\n")
+            frameStk:__clear()
+        else
+            if cudaDriverExists then
```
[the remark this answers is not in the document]
ocaisa: Done

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+            modT.isVisible = false
+        end
+    end
+end
+
+local function cuda_enabled_load_hook(t)
+    local frameStk  = require("FrameStk"):singleton()
+    local mt        = frameStk:mt()
+    local simpleName = string.match(t.modFullName, "(.-)/")
+    local eprefix = os.getenv('EESSI_PREFIX') .. "/init/gpu_support"
```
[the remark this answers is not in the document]
ocaisa: Gone

create_lmodrc.py
ocaisa: Pointing to that doc page for now

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+            io.stderr:write("export INSTALL_CUDA=true && ./add_nvidia_gpu_support.sh\\n")
+            frameStk:__clear()
+        end
+    end
+    -- when loading CUDA enabled modules check if the necessary matching compatibility libraries are installed
+    -- otherwise, refuse to load the requested module and print error message
+    local haveGpu = mt:haveProperty(simpleName,"arch","gpu")
+    if haveGpu then
+        local arch = os.getenv("EESSI_CPU_FAMILY") or ""
+        local cudaVersionFile = "/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/" .. arch .. "/latest/version.txt"
```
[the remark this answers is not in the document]
ocaisa: Resolved now

create_lmodrc.py
ocaisa: Fixed by using native Lmod capabilities

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+        end
+    end
+end
+
+local function cuda_enabled_load_hook(t)
+    local frameStk  = require("FrameStk"):singleton()
+    local mt        = frameStk:mt()
+    local simpleName = string.match(t.modFullName, "(.-)/")
+    local eprefix = os.getenv('EESSI_PREFIX') .. "/init/gpu_support"
+    -- if we try to load CUDA itself, check if the software exists in host_injections
```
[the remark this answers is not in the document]
ocaisa: Done

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+        local hostInjections = string.gsub(os.getenv('EESSI_SOFTWARE_PATH') or "", 'versions', 'host_injections')
+        -- build final path where the CUDA software should be installed
+        local cudaEasyBuildDir = hostInjections .. "/software/" .. t.modFullName .. "/easybuild"
+        local cudaDirExists = exists(cudaEasyBuildDir)
+        if not cudaDirExists then
+            io.stderr:write("You requested to load ",simpleName,"\\n")
+            io.stderr:write("While the module file exists, the actual software is not shipped with EESSI.\\n")
+            io.stderr:write("In order to be able to use the CUDA module, please follow the instructions in the\\n")
+            io.stderr:write("gpu_support folder. Adding the CUDA software can be as easy as:\\n")
+            io.stderr:write("export INSTALL_CUDA=true && ./add_nvidia_gpu_support.sh\\n")
```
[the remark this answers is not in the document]
ocaisa: Pointer to docs is done for now

create_lmodrc.py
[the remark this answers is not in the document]
ocaisa: Done

create_lmodrc.py
```diff
@@ -17,6 +17,113 @@
...
+                    elseif minor == minor_req then
+                        if patch < patch_req then
+                            compat_libs_need_update = true
+                        end
+                    end
+                end
+                if compat_libs_need_update == true then
+                    io.stderr:write("You requested to load CUDA version ",cudaVersion)
+                    io.stderr:write("but the module you want to load requires CUDA version ",cudaVersion_req,".\\n")
+                    io.stderr:write("Please update your CUDA compatibility libraries in order to use ",simpleName,".\\n")
```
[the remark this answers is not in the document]
ocaisa: done

create_lmodrc.py
```diff
@@ -17,6 +17,81 @@
...
+        end
+    end
+    -- when loading CUDA enabled modules check if the necessary driver libraries are accessible to the EESSI linker,
+    -- otherwise, refuse to load the requested module and print error message
+    local haveGpu = mt:haveProperty(simpleName,"arch","gpu")
+    if haveGpu then
+        local arch = os.getenv("EESSI_CPU_FAMILY") or ""
+        local cudaVersionFile = "/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/" .. arch .. "/latest/cuda_version.txt"
+        local cudaDriverFile = "/cvmfs/pilot.eessi-hpc.org/host_injections/nvidia/" .. arch .. "/latest/libcuda.so"
+        local cudaDriverExists = isFile(cudaDriverFile)
```
ocaisa, line 59: This seems not to be sufficient, if I start the container without a GPU the module still succeeds in loading. I suspect that `isFile` accepts that the link exists and ignores that it happens to be dead.

ocaisa: Yes, that seems to be the case https://github.com/TACC/Lmod/blob/55d353b3a3a4649cbaaef18b2f00cdc017d7cc92/proj_mgmt/updateVersion#L126

casparvl: Hm, that's annoying. Should we revert back to your function and just make additional error codes pass? Any fix from Lmod will take a while, we defintily don't want to wait for that...

ocaisa: I think it is ok to leave this as is, the case is a real corner one: you have the link available but you are on a node that doesn't have a GPU...no surprise that that won't work

ocaisa: If the user does things "properly" that should not happen

casparvl: Well, I can see that corner case appearing if HPC sites point their `host_injections` variant symlink to a shared directory for convenience, and just put symlinks there to where the drivers are on their nodes that _do_ contain GPUs. But agreed, let's just go ahead for now. It's not like it breaks something that wasn't already broken: if you expect GPU software to work on a non-GPU node, you have different issues xD

ocaisa: The only negative is that the error when you run the GPU code makes it look like all you have is a driver compatibility problem

ocaisa: Oh ye if little faith! Lmod has been updated in the develop branch to address this.

ocaisa:
This actually needs to be built by the bot, but won't work without first using the script to install CUDA under `host_injections`

ocaisa:

PR #382, opened 2023-11-14
adding waLBerla 6.1 with foss/2021a to EESSI pilot 2023.06

xinan1911:


boegel:
@xinan1911 If you close a PR for some reason, please mention the reason in a comment when closing (and if there's a follow-up PR, provide a pointer to that), so nobody is left wondering why this was closed.

xinan1911:
@boegel Thanks for the tip. Next time I will leave a comment if I close the PR. 
This PR is closed because the branch where I created this PR is not suitable as it was already used for a different PR.

PR #383, opened 2023-11-15
update script for building for software.eessi.io (version 2023.06)

boegel:
* change to `/cvmfs/software.eessi.io` as default target repo
* get rid of "pilot" (rename `EESSI_PILOT_VERSION` to `EESSI_VERSION`, etc.)

marked as WIP because i) untested, ii) maybe we want to introduce a new `2023.06-eessi.io` branch, or fork off `2023.06-pilot` first to retain possibility for building for EESSI pilot repo

edit: as requested, I will break up this PR into smaller ones that are easier to review & merge:
* #387
* #388
* #389
* #390
* #391
* #392
* #393
* #398

Extra PRs beyond what's changed in this one:

* #397
* #399

Issue #322, opened 2023-08-31
Non-optimal CPU detection of zen3 using archspec


laraPPr:
When setting up the EESSI environment on the `zen3` nodes on `aws/citc` archspec detects `zen3` instead of `zen3`.

```
[laraPPr@fair-mastodon-c6a-4xlarge-0001 ~]$ archspec cpu

zen2

```

laraPPr:
There is one missing CPU feature on `c6a.2xlarge` and `c6a.4xlarge`: pku

There is also an [open issue](https://github.com/archspec/archspec-json/issues/38) in archspec on this.

boegel:
As mentioned in #322, this in combination with #322, is a sufficient reason to switch to our homegrown `archdetect` alternative as default mechanism imho.

laraPPr:
EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash

- c6a.2xlarge
```
[laraPPr@fair-mastodon-c6a-2xlarge-0001 ~]$ EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash

Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2023.06!

2023-09-01 11:51:42 [INFO] cpupath: best match for host CPU: x86_64/amd/zen3

archdetect says x86_64/amd/zen3

Using x86_64/amd/zen3 as software subdirectory.

Using /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/amd/zen3/modules/all as the directory to be added to MODULEPATH.

Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/amd/zen3/.lmod/lmodrc.lua

Initializing Lmod...

Prepending /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/x86_64/amd/zen3/modules/all to $MODULEPATH...

Environment set up to use EESSI pilot software stack, have fun!

[EESSI pilot 2023.06] $ 
```
- c6a.4xlarge
```
[laraPPr@fair-mastodon-c6a-4xlarge-0001 ~]$ EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash

[...]

Issue #320, opened 2023-08-29
Non-optimal CPU detection of neoverse_v1


laraPPr:
When setting up the EESSI environment on the `neoverse_v1` nodes on `aws/citc` archspec detects `neoverse_n1` instead of `neoverse_v1`.
```
@fair-mastodon-c7g-2xlarge-0002 ~]$ source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash

Found EESSI pilot repo @ /cvmfs/pilot.eessi-hpc.org/versions/2023.06!

archspec says aarch64/neoverse_n1

Using aarch64/neoverse_n1 as software subdirectory.

Using /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/aarch64/neoverse_n1/modules/all as the directory to be added to MODULEPATH.

Found Lmod configuration file at /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/aarch64/neoverse_n1/.lmod/lmodrc.lua

Initializing Lmod...

Prepending /cvmfs/pilot.eessi-hpc.org/versions/2023.06/software/linux/aarch64/neoverse_n1/modules/all to $MODULEPATH...

Environment set up to use EESSI pilot software stack, have fun!
```

laraPPr:
There are two missing CPU features on c7g-2xlarge: `paca` and `pacg`
You can check the list [here](https://github.com/archspec/archspec-json/blob/master/cpu/microarchitectures.json)

boegel:
Together with #322, this is enough motivation to switch to using our own minimal `archdetect` implementation rather than relying on `archspec` for EESSI pilot 2023.06, I think...

ocaisa:
@boegel There's an error in archspec that is fixed as part of https://github.com/EESSI/software-layer/pull/264

boegel:
@ocaisa You mean `archdetect`, right?

boegel:
@laraPPr Can you check whether `archdetect` correctly detects both `neoverse_v1` and `zen3` (cfr. #322), using:

```
EESSI_USE_ARCHDETECT=1 source /cvmfs/pilot.eessi-hpc.org/versions/2023.06/init/bash
```

laraPPr:
Results on `zen3`: https://github.com/EESSI/software-layer/issues/322#issuecomment-1702640074
- c7g.2xlarge
```
[...]

PR #387, opened 2023-11-24
update init scripts + README for `software.eessi.io` (+ drop 'pilot' terminology)

boegel:
Tests in CI will still fail, because they require that `/cvmfs/software.eessi.io/versions/2023.06/init/*` is in place...

### DIFF
--- a/init/eessi_defaults
+++ b/init/eessi_defaults
@@ -8,5 +8,8 @@
 # license: GPLv2
 #
 
-export EESSI_CVMFS_REPO="${EESSI_CVMFS_REPO_OVERRIDE:=/cvmfs/pilot.eessi-hpc.org}"
-export EESSI_PILOT_VERSION="${EESSI_PILOT_VERSION_OVERRIDE:=2023.06}"
+export EESSI_CVMFS_REPO="${EESSI_CVMFS_REPO_OVERRIDE:=/cvmfs/software.eessi.io}"
+export EESSI_VERSION="${EESSI_VERSION_OVERRIDE:=2023.06}"
+# use archdetect by default, unless otherwise specified
+export EESSI_USE_ARCHDETECT="${EESSI_USE_ARCHDETECT:=1}"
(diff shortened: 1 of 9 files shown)


boegel:
@bedroge We should ignore CI failure here, and get this merged first, since we need this to deploy `/cvmfs/software.eessi.io/versions/2023.06/init`.

Without having that in place, we'll run into CI failures anyhow, because some of the tests rely on having `/cvmfs/software.eessi.io/versions/2023.06/init` in place.

We need to break the chicken-egg situation here, or we'll just end up with the same changes as in #383 (with CI still failing).

init/eessi_defaults
```diff
@@ -8,5 +8,8 @@
 # license: GPLv2
 #
 
-export EESSI_CVMFS_REPO="${EESSI_CVMFS_REPO_OVERRIDE:=/cvmfs/pilot.eessi-hpc.org}"
-export EESSI_PILOT_VERSION="${EESSI_PILOT_VERSION_OVERRIDE:=2023.06}"
+export EESSI_CVMFS_REPO="${EESSI_CVMFS_REPO_OVERRIDE:=/cvmfs/software.eessi.io}"
+export EESSI_VERSION="${EESSI_VERSION_OVERRIDE:=2023.06}"
+# use archdetect by default, unless otherwise specified
+export EESSI_USE_ARCHDETECT="${EESSI_USE_ARCHDETECT:=1}"
```
boegel, line 14: I just want to point out here that this flips the default from using `archspec` (like we do in `pilot.eessi-hpc.org`) to our pure bash `archdetect` alternative.

Not only is that significantly faster (no overhead of starting a Python interpreter), but it will also help to bypass some shortcomings in `archspec` (see #320 + #322).

bedroge (approved):
Lgtm

init/eessi_defaults
bedroge: :+1:

PR #391, opened 2023-11-25
get rid of 'pilot' terminology in scripts

boegel:
requires (for CI to pass):

* #390


boegel:
CI fails with:

```
failed to discover HTTP proxy servers (23 - proxy auto-discovery failed)
cannot load store (null) at /cvmfs_ro/software.eessi.io
fuse-overlayfs: cannot read lower dirs: Transport endpoint is not connected
ERROR: /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64 does not exist!
```

Maybe I overlooked something in #390? 🤔

bedroge (approved):

PR #400, opened 2023-11-26
add JSON file + Python script to keep track of software licenses

boegel:
Draft, since I would like to add another step in the CI workflow that collects a list of software names installed in `/cvmfs/software.eessi.io`, and checks whether all software licenses are known for all of these.


dagonzalezfo (approved):
As initial step for including licence management for EESSI distributed software.

PR #410, opened 2023-12-01
Build Cuda under host injections. Replicates changes that were initially targetted at 2023.06-pilot repo

casparvl:
Replicates https://github.com/EESSI/software-layer/pull/368 for the 2023.06-pilot repo

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -187,6 +187,22 @@ fi
 # assume there's only one diff file that corresponds to the PR patch file
 pr_diff=$(ls [0-9]*.diff | head -1)
 
+# install any additional required scripts
+# order is important: these are needed to install a full CUDA SDK in host_injections
+install_scripts_changed=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^install_scripts.sh$' > /dev/null; echo $?)
+if [ ${install_scripts_changed} == '0' ]; then
+    # for now, this just reinstalls all scripts. Note the most elegant, but works
+    ${TOPDIR}/install_scripts.sh --prefix ${EESSI_CVMFS_REPO}
+fi
+
+# Install full CUDA SDK in host_injections
+# Hardcode this for now, see if it works
+# TODO: We should make a nice yaml and loop over all CUDA versions in that yaml to figure out what to install
+${EESSI_CVMFS_REPO}/gpu_support/nvidia/install_cuda_host_injections.sh 12.1.1
+
+# Install drivers in host_injections
+${EESSI_CVMFS_REPO}/gpu_support/nvidia/link_nvidia_host_libraries.sh
+
 # use PR patch file to determine in which easystack files stuff was added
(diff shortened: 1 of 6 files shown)


casparvl:
I believe all the changes for software.eessi.io have been made in this PR, but still need to test...

ocaisa (approved):

ocaisa:
This needs to actually be deployed, but the deployment is coming in #434 so I think it's ok to merge as is

EESSI-install-software.sh
```diff
@@ -187,6 +187,22 @@ fi
 # assume there's only one diff file that corresponds to the PR patch file
 pr_diff=$(ls [0-9]*.diff | head -1)
 
+# install any additional required scripts
+# order is important: these are needed to install a full CUDA SDK in host_injections
+install_scripts_changed=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^install_scripts.sh$' > /dev/null; echo $?)
```
ocaisa, line 192: @casparvl I messed up a little by merging this as now the file checked here will not be updated in the other PR, and as a result the scripts will not get installed. It doesn't matter too much as I think the use of `$pr_diff` is overkill here. I think we should always run the installer script and instead of using `cp` use `cp -u`, that could be implemented as part of #434

Also, this was a little wrong anyway, as it doesn't check if the scripts _installed_ by the installer have changed

EESSI-install-software.sh
```diff
@@ -187,6 +187,22 @@ fi
...
+if [ ${install_scripts_changed} == '0' ]; then
+    # for now, this just reinstalls all scripts. Note the most elegant, but works
+    ${TOPDIR}/install_scripts.sh --prefix ${EESSI_CVMFS_REPO}
+fi
+
+# Install full CUDA SDK in host_injections
+# Hardcode this for now, see if it works
+# TODO: We should make a nice yaml and loop over all CUDA versions in that yaml to figure out what to install
+${EESSI_CVMFS_REPO}/gpu_support/nvidia/install_cuda_host_injections.sh 12.1.1
```
ocaisa, line 201: Wow, I really messed up here...the name of the script and arguments are incorrect!

install_scripts.sh
```diff
@@ -0,0 +1,62 @@
...
+# Subdirs for GPU support
+NVIDIA_GPU_SUPPORT_DIR_SOURCE=${TOPDIR}/gpu_support/nvidia/  # Source dir
+NVIDIA_GPU_SUPPORT_DIR_TARGET=${INSTALL_PREFIX}/gpu_support/nvidia/  # Target dir
+
+# Create target dir
+mkdir -p ${NVIDIA_GPU_SUPPORT_DIR_TARGET}
+
+# Copy files from this directory into the prefix
+# To be on the safe side, we dont do recursive copies, but we are explicitely copying each individual file we want to add
+for file in install_cuda_host_injections.sh link_nvidia_host_injections.sh; do
```
ocaisa, line 60: Script name is wrong here

PR #411, opened 2023-12-02
Prepend EESSI version to PS1 instead of overriding PS1

pescobar:
While testing EESSI I noticed PS1 is overridden which I don't find very convenient. 

With current settings I get this $PS1 when I enable EESSI:

```
source /cvmfs/software.eessi.io/versions/2023.06/init/bash
[EESSI 2023.06] $
```

With this change I prepend EESSI version to my existing $PS1

```
 source /tmp/bash
[EESSI 2023.06][escobar@sei68 ~]$
```

### DIFF
--- a/init/bash
+++ b/init/bash
@@ -13,7 +13,7 @@ source $(dirname "$BASH_SOURCE")/eessi_environment_variables
 # only continue if setting EESSI environment variables worked fine
 if [ $? -eq 0 ]; then
 
-    export PS1="[EESSI $EESSI_VERSION]$PS1"
+    export PS1="{EESSI $EESSI_VERSION}$PS1"
 
     # add location of commands provided by compat layer to $PATH;
     # see https://github.com/EESSI/software-layer/issues/52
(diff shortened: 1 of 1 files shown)


init/bash
```diff
@@ -13,7 +13,7 @@ source $(dirname "$BASH_SOURCE")/eessi_environment_variables
 # only continue if setting EESSI environment variables worked fine
 if [ $? -eq 0 ]; then
 
-    export PS1="[EESSI $EESSI_VERSION] $ "
+    export PS1="[EESSI $EESSI_VERSION]$PS1"
```
boegel, line 16: Maybe `{EESSI 2023.06} ` would even be better (with a space in between) since `[...]` is already quite common in `$PS1`, so it stands a bit out more?

Compare:

```
[EESSI 2023.06][escobar@sei68 ~]$
```

with

```
{EESSI 2023.06} [escobar@sei68 ~]$
```

boegel (changes requested):

init/bash
pescobar: I have no strong opinion. Whatever that doesn't override my PS1 works for me.

I have no computer access now. I will update the PR with your suggestion later today.

trz42 (approved):
Looks good to me. Thanks a lot for this contribution @pescobar !

boegel:
tarball `eessi-2023.06-init-1703152153.tar.gz` created and uploaded to S3 bucket to deploy these changes...

PR #415, opened 2023-12-05
Fix permission denied when `echo`ing to `/dev/stdout`

### LINKED ISSUE #413: -bash: /dev/stdout: Permission denied when enabling EESSI
while testing EESSI I had this issue when using `su -` to test as a regular use

```
$> source /cvmfs/software.eessi.io/versions/2023.06/init/bash
-bash: /dev/stdout: Permission denied
-bash: /dev/stdout: Permission denied
-bash: /dev/stdout: Permission denied
-bash: /dev/stdout: Permission denied
-bash: /dev/stdout: Permission denied
```

The issue is described here https://unix.stackexchange.com/a/38580

I think the issue is triggered here 

```
$> head -7 /cvmfs/software.eessi.io/versions/2023.06/init/bash
# Allow for a silent mode
if [[ -v EESSI_SILENT ]]; then
  # EESSI_SILENT set
  output=/dev/null
else
  output=/dev/stdout
fi
```

I can workaround the issue connecting as a regular user instead of using `su -` or setting `export EESSI_SILENT=true` but I am reporting it as suggested by @boegel

trz42:
In some cases, `echo`ing to `/dev/stdout` triggers a permission denied. Particularly, this happens when sourcing the init script (`init/bash`) or the script to set key EESSI environment variables (`init/eessi_environment_variables`). See #413

This PR works around the issue by implementing a fix suggested in https://unix.stackexchange.com/a/38580

fixes #413

### DIFF
--- a/init/bash
+++ b/init/bash
@@ -1,10 +1,17 @@
-# Allow for a silent mode
-if [[ -v EESSI_SILENT ]]; then
-  # EESSI_SILENT set
-  output=/dev/null
-else
-  output=/dev/stdout
-fi
+function show_msg {
+  msg=$1
+  if [[ ! -v EESSI_SILENT ]]; then
+    echo "$msg"
+  fi
+}
+
+## Allow for a silent mode
+#if [[ -v EESSI_SILENT ]]; then
+#  # EESSI_SILENT set
+#  output=/dev/null
+#else
+#  output=/dev/stdout
+#fi
 
 # The following method should be safe, but might break if file is a symlink
 # (could switch to $(dirname "$(readlink -f "$BASH_SOURCE")") in that case)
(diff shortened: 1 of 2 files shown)


ocaisa (approved):
Seems to work fine:
```
ocaisa@LAPTOP-O6HF2IKC:~$ cat bash.func
function show_msg {
  # only echo msg if EESSI_SILENT is unset
  msg=$1
  if [[ ! -v EESSI_SILENT ]]; then
    echo "$msg"
  fi
}
ocaisa@LAPTOP-O6HF2IKC:~$ source bash.func
ocaisa@LAPTOP-O6HF2IKC:~$ show_msg test
test
ocaisa@LAPTOP-O6HF2IKC:~$ EESSI_SILENT=1 show_msg test
ocaisa@LAPTOP-O6HF2IKC:~$
```

ocaisa:
@boegel I'm not sure how this would actually get deployed so I'm reluctant to hit merge

bedroge:
> @boegel I'm not sure how this would actually get deployed so I'm reluctant to hit merge

We basically need to redo the procedure for ingesting the init scripts: that means merging this PR, creating a new tarball using https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/create_directory_tarballs.sh, and uploading the tarball to the S3 bucket using https://github.com/EESSI/eessi-bot-software-layer/blob/develop/scripts/eessi-upload-to-staging.

boegel:
tarball `eessi-2023.06-init-1703152153.tar.gz` created and uploaded to S3 bucket to deploy these changes...

PR #418, opened 2023-12-06
{2023.06}[2023a] Rivet 3.1.9

APN-Pucky:


trz42:
Thanks for the PR @APN-Pucky . I retargeted the PR to `2023.06-software.eessi.io`. In that branch/version of the software stack we currently have foss/2023a (GCC/12.3.0) available. Soon, there should be foss/2022b (GCC/12.2.0) too.

Could you try to use either of these?
- You would need to use a different easyconfig file (if it is not available yet via EasyBuild v4.8.2, you would need to open a pull request for a new/updated easyconfig file to the [easybuilders/easyconfigs](https://github.com/easybuilders/easybuild-easyconfigs) repository first, apologies for the extra work)
- You would have to move the easyconfig from `eessi-2023.06-eb-4.7.2-2022a.yml` to either `easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.8.2-2023a.yml` or `easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.8.2-2022b.yml`

Please let us know if you struggle with any of these changes / additional work (PR for easyconfig).

APN-Pucky:
Since this will also require updates of Rivets dependencies should they all in one PR to easyconfigs or single per-package requests?

trz42:
> Since this will also require updates of Rivets dependencies should they all in one PR to easyconfigs or single per-package requests?

They should go in one PR I think. Otherwise you would have to list the dependencies in the easystack file and add the option `from-pr:` for each easyconfig you have opened a PR.

boegel:
@APN-Pucky Are you planning to follow up on this?

APN-Pucky:
> @APN-Pucky Are you planning to follow up on this?

Yes, but I thought when I am already at it I'd update all the dependencies in the 2023a for easybuild too.

APN-Pucky:
The easyconfigs MR is merged now. Can the build bot please be triggered?

APN-Pucky:
What does that mean?

casparvl:
It means there was _some_ error during the build. I'll have to go into the system that builds things to find the error. 

Note however that another way to debug is to rerun the build yourself, manually, according to the instructions on https://www.eessi.io/docs/adding_software/debugging_failed_builds/ . This _should_ mimic exactly what our bot does, and thus should result in the same build failure. Ideally, that lets you debug the issue on your own system, which is much easier :)

For now, I'll go into our build system and check if the error is maybe trivial... I'll paste it here in a minute

casparvl:
```
checking for zlib.h... no
configure: error: either specify a valid zlib installation with --with-zlib=DIR or disable zlib usage with --without-zlib
```
That's... weird to say the least, since I'm assuming that `zlib` is listed _somewhere_ as a dependency (directly or indirectly) for `Rivet`...

casparvl:
Note that some issues may pop up in our builds that you might not have seen before since we build in a very minimal container. Thus, where some EasyConfigs might succesfully build because they silently pick up on system dependencies, they would fail in this minimal container. It's actually a very good test to figure out if all dependencies have been listed correctly.

In YODA's case, they haven't. This configure tells me YODA (at least in the configuration as provided by [this EasyConfig](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/y/YODA/YODA-1.9.9-GCC-12.3.0.eb)) depends on `zlib`, but that is not listed in the `dependencies`. So, step 1 would be to add that.

However, in this particular case, it won't resolve the issue: in EESSI, we filter `zlib` as a dependency, as we want to be able to do security updates to `zlib`. Thus, this package _should_ pick up on the `zlib` from our [compatibility layer](https://www.eessi.io/docs/compatibility_layer/), which we can apply security patches to if needed. However, clearly, the configure fails to find it. My bet is, it fails to find it, because it is in a non-standard location (e.g. `/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/include/zlib.h` for x86_64 architectures). I need to think how to best resolve this. I'm actually quite surprised we haven't hit this with other packages - zlib is a very common dependency. Maybe this configure somehow uses a different mechanism of finding zlib then most others...? I'll check with some other people in EESSI for ideas.

Note, unrelated to it's installation in EESSI: it's still a good idea to add `zlib` as dependency to the `YODA` EasyConfig. It'll guarantee that it uses the `zlib` provided by EasyBuild when people install `YODA` with EasyBuild, instead of a system-version.

APN-Pucky:
Thanks, I'll update the YODA easyconfigs.

APN-Pucky:
> Maybe this configure somehow uses a different mechanism of finding zlib then most others...?

I updated the zlib dependency here: https://github.com/easybuilders/easybuild-easyconfigs/pull/19679
How does the zlib dependency replacement work? Above MR also adds a previously unused `--with-zlib=$EBROOTZLIB` configopt. If EESSI just replaces that variable, this might fix it.

casparvl:
> How does the zlib dependency replacement work? Above MR also adds a previously unused --with-zlib=$EBROOTZLIB configopt.

We just do `--filter-deps=zlib,<some more>`, so we use that native EasyBuild option. It _does_ mean that actually `--with-zlib=$EBROOTZLIB` will not work in that context, since `EBROOTZLIB` will be unset if `zlib` is filtered by EasyBuild...

I'm still curious how the configure was _not_ picking up on the `zlib` from the compatibility layer though. Did you try to reproduce the issue interactively with the container, using the instructions I linked? I'd be interested to know what the `config.log` shows for the `zlib.h` test. That sometimes contains more specific information on what `configure` _tried_ to do to find `zlib` and might give us a hint on why it failed...

casparvl:
From the `config.log`:

```
configure:18085: checking for zlib.h
configure:18085: gcc -c -O2 -ftree-vectorize -march=native -fno-math-errno -fPIC -I/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.3-GCCcore-12.3.0/include -I/usr/include conftest.c >&5
In file included from conftest.c:31:
/usr/include/stdio.h:781:10: fatal error: bits/sys_errlist.h: No such file or directory
  781 | #include <bits/sys_errlist.h>
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

APN-Pucky:
> /usr/include/stdio.h

The problem here is `-I/usr/include`, right? Just `autoreconf` probably won't help?

casparvl:
As for the solution... I'd really be interested what triggers to configure to add this `-I/usr/include` and to get rid of it... The fact that you don't want to pick up anything from the host (e.g. from `/usr/include`) is more general in EasyBuild. It's just that our container-build setup exposed the issue...

casparvl:
Uhg, it has this very ugly section in the configure...

```
## Optional zlib support for gzip-compressed data streams/files
#
# Handle user hints
#
{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking if zlib is wanted" >&5
printf %s "checking if zlib is wanted... " >&6; }
zlib_places="/usr/local /usr /opt/local /sw"

# Check whether --with-zlib was given.
if test ${with_zlib+y}
then :
  withval=$with_zlib; if test "$withval" != no ; then
  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
printf "%s\n" "yes" >&6; }
  if test -d "$withval"
  then
    zlib_places="$withval $zlib_places"
  else
    { printf "%s\n" "$as_me:${as_lineno-$LINENO}: WARNING: Sorry, $withval does not exist, checking usual places" >&5
printf "%s\n" "$as_me: WARNING: Sorry, $withval does not exist, checking usual places" >&2;}
  fi
else
  zlib_places=
  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: no" >&5
printf "%s\n" "no" >&6; }
fi
else $as_nop
  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
printf "%s\n" "yes" >&6; }
fi


#
# Locate zlib, if wanted
#
if test -n "${zlib_places}"
then
        # check the user supplied or any other more or less 'standard' place:
        #   Most UNIX systems      : /usr/local and /usr
        #   MacPorts / Fink on OSX : /opt/local respectively /sw
        for ZLIB_HOME in ${zlib_places} ; do
          if test -f "${ZLIB_HOME}/include/zlib.h"; then break; fi
          ZLIB_HOME=""
        done

  ZLIB_OLD_LDFLAGS=$LDFLAGS
  ZLIB_OLD_CPPFLAGS=$CPPFLAGS
  if test -n "${ZLIB_HOME}"; then
        LDFLAGS="$LDFLAGS -L${ZLIB_HOME}/lib"
        CPPFLAGS="$CPPFLAGS -I${ZLIB_HOME}/include"
  fi
```
Clearly, that `for ZLIB_HOME` loop is looping over system directories, and then locating the system zlib (instead of the one provided by EasyBuild _or_ the one from the compatiblity layer - as we would want in EESSI).

casparvl:
It originally comes from here btw:
```
cat m4/ax_check_zlib.m4 | grep zlib_places
zlib_places="/usr/local /usr /opt/local /sw"
    zlib_places="$withval $zlib_places"
  zlib_places=
if test -n "${zlib_places}"
        for ZLIB_HOME in ${zlib_places} ; do
```

casparvl:
Apparently, it is the officially documented behaviour for `ax_check_zlib`, even if we might not like it from an EasyBuild perspective... https://www.gnu.org/software/autoconf-archive/ax_check_zlib.html 

I see two potential approaches:
1. Change the `ax_check_zlib.m4` so that it finds the right zlib. For standard EB, that would mean: the zlib listed as dependency. For anyone having zlib in the `filter-deps` (like EESSI), it would mean using `%(sysroot)s/include`. 
2. Pass `--with-zlib` and as argument, use `$EBROOTZLIB/include` (if zlib is not in `filter-deps`) or use `%(sysroot)s/include` if it is.

Now, since this involves logic (namely, checking an EasyBuild option, doing one thing in one case, and another thing in another case), it has to be done at an EasyBlock level, regardless of whether we go for option 1 or 2. Personally, I think 2 would be the cleanest approach.

It shouldn't be rocket science to implement it, but still a bit of work. Before you go through that, I'd like to at least get one more opinion from one of the EESSI support people... I'll reach out to them on Slack, see if I can get some input on this :)

boegel:
I'm not sure it's true that this requires logic and hence an easyblock. How about this construct (use `%(sysroot)s` template to define a fallback value in case `$EBROOTZLIB` is not defined:

```python
configopts = "--with-zlib=${EBROOTZLIB:-%(sysroot)s}/include "
```

APN-Pucky:
That's smart. The fallback mode seems to work in my build on top of EESSI and in the debug container. I updated the PR reference in this PR.

casparvl:
Great idea from @boegel and much simpler. I knew I was asking for a second pair of eyes for a good reason ;-)

I'll kick off another bot build.

Just in case we wonder about this in the future: I saw that your `--with-zlib` was slightly different from what Kenneth suggested, namely

```
--with-zlib=${EBROOTZLIB:-%(sysroot)s/usr}
```
But indeed from the docs of `ax_check_zlib.m4` that should be correct:

> If the –with-zlib=DIR is specified, it will try to find it in DIR/include/zlib.h and DIR/lib/libz.a.

casparvl:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/haswell

casparvl:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/skylake_avx512
bot: build repo:eessi.io-2023.06-software arch:aarch64/generic
bot: build repo:eessi.io-2023.06-software arch:aarch64/neoverse_n1
bot: build repo:eessi.io-2023.06-software arch:aarch64/neoverse_v1

casparvl:
Just FYI: I'd like to merge https://github.com/easybuilders/easybuild-easyconfigs/pull/19679 so we can be sure that's the final version before we deploy these builds in EESSI. I'm waiting on one more test report from a build I'm running, if that's ok, I think we can merge there and deploy here. I might be able to find a few minutes tonight to do that... otherwise it'll be tomorrow.

APN-Pucky:
Thanks for getting this through despite the hurdles!

casparvl:
> Thanks for getting this through despite the hurdles!

Sure! Thank _you_ for pushing through :) As you've noticed, we are definitely still building experience with the particularities of the EESSI setup, but things go more and more smoothly as we learn :)

casparvl (approved):
Lgtm!

PR #432, opened 2023-12-19
{2023.06}[GCCcore/12.3.0] FFmpeg 6.0

Neves-P:
First installing FFmpeg v6.0 as a necessary dependency of #404
```
7 out of 34 required modules missing:

* ffnvcodec/12.0.16.0 (ffnvcodec-12.0.16.0.eb)
* x264/20230226-GCCcore-12.3.0 (x264-20230226-GCCcore-12.3.0.eb)
* LAME/3.100-GCCcore-12.3.0 (LAME-3.100-GCCcore-12.3.0.eb)
* Yasm/1.3.0-GCCcore-12.3.0 (Yasm-1.3.0-GCCcore-12.3.0.eb)
* x265/3.5-GCCcore-12.3.0 (x265-3.5-GCCcore-12.3.0.eb)
* SDL2/2.28.2-GCCcore-12.3.0 (SDL2-2.28.2-GCCcore-12.3.0.eb)
* FFmpeg/6.0-GCCcore-12.3.0 (FFmpeg-6.0-GCCcore-12.3.0.eb)
```

PR #433, opened 2023-12-19
Explicitly set `$EESSI_CPU_FAMILY` in CI that checks missing installations

bedroge:
This sets/overrides `$EESSI_CPU_FAMILY` to the part before the first slash of `$EESSI_SOFTWARE_SUBDIR_OVERRIDE`, as it's otherwise always set to the architecture of the runner (i.e. `x86_64`). This causes issues with checks in our EB configuration that use `$EESSI_CPU_FAMILY`, e.g. for filtering dependencies for `aarch64` CPUs, which in turn cause the CI to fail for these architectures. This just happened in #432 .


ocaisa:
I was just about to make the same PR. Shouldn't you also do it in the second instance below this line (line 58)?

bedroge:
> I was just about to make the same PR. Shouldn't you also do it in the second instance below this line (line 58)? I know it has no practical consequence but since you know it's a potential issue, might as well keep it consistent

Ah, thanks, overlooked that one. May not be strictly necessary (as it only checks GCC), but I've added it anyway.

PR #434, opened 2023-12-19
Add CUDA, CUDA samples, and CUDA related hooks and lmodrc changes [WIP, do NOT deploy unless sure it works!]

casparvl:
Equivalent of https://github.com/EESSI/software-layer/pull/381, now targetting the new eessi.io domain.

PR #441, opened 2024-01-02
don't include files in .lmod directory twice in tarball

boegel:
This was introduced accidentally in #434 by me via 42e3404a0015d7b4049c5cd3e5a01ee764261d44


trz42 (approved):
Looks good. Thanks for the quick fix.

PR #447, opened 2024-01-10
add zen4 / AMD Genoa

hmeiland:
updated tests and config files;

### DIFF
--- a/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a.yml
+++ b/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a.yml
@@ -1,0 +1,4 @@
+easyconfigs:
+  - OpenFOAM-11-foss-2023a.eb:
+      options:
+        from-pr: 19545


ocaisa:
I don't think we've actually thought about what to do if someone wants to introduce a new architecture into EESSI production. This is another use case for a development repo that sits on top of production EESSI, you figure out how far you can get there and then when things are mature you can rebuild them into production.

boegel:
@hmeiland Somehow there's a merge conflict now here...

boegel:
> I don't think we've actually thought about what to do if someone wants to introduce a new architecture into EESSI production. This is another use case for a development repo that sits on top of production EESSI, you figure out how far you can get there and then when things are mature you can rebuild them into production.

That's totally orthogonal to this PR, no?

Let's follow up on that in https://gitlab.com/eessi/support/-/issues/37

ocaisa:
> That's totally orthogonal to this PR, no?
> 
> Let's follow up on that in https://gitlab.com/eessi/support/-/issues/37

Almost orthogonal but not quite, current CI assumes that if something is supported (in CI) by archdetect, then there should be a corresponding software directory. It was a bit looser before (checking that it had to be available in _some_ EESSI version, which is to accommodate the deprecation of Power).

hmeiland:
moved to #451

PR #457, opened 2024-01-18
Licenses for EESSI

dagonzalezfo:
WIP

### DIFF
--- a/licenses/add_license_info.py
+++ b/licenses/add_license_info.py
@@ -13,7 +13,7 @@ parser=argparse.ArgumentParser(description='Script to ingest licences')
 
 parser.add_argument('--source', help='Project available in GitHub,pypi,cran or user that push the license')
 parser.add_argument('project', help='Project')
-parser.add_argument('--spdx',required=False)
+parser.add_argument('--manual',required=False)
 args=parser.parse_args()
 
 def github(source):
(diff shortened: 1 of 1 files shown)


dagonzalezfo:
Now the add_license_info.py script can process pypi, github and manual license addition to licenses.json file. There is also included work for other license fonts (libraries.io, repology and CRAN). This is fully functional, ready for review.

dagonzalezfo:
Waiting for review/approval

licenses/add_license_info.py
```diff
@@ -0,0 +1,127 @@
...
+    Function that updates json file
+    """
+    with open('licenses.json','r') as licDict:
+        licenses=json.loads(licDict.read())
+    
+    if project in licenses.keys():
+        print('project in licenses.json')
+    else: 
+        print('we do not have the license, adding into licenses.json')
+        licenses[project]=dict(info)
```
ocaisa, line 116: From what I can see, this doesn't store information for the case for a licence change or multiple licences (academic versus commercial).

licenses/add_license_info.py
```diff
@@ -0,0 +1,127 @@
...
+        licJson=json.dumps(licenses, indent=4)
```
ocaisa, line 117: Perhaps we should give 2 options here. There's kind of an assumption that this will make a change but if this is called within a GitHub Action the update to the licences file will be lost (without giving the Action permissions to write to the PR). It would be good to allow the display of the information in patch format so that it can be copied from the GitHub Actions and applied (or even [posted as a review](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request)).

ocaisa (changes requested):
I wonder if we shouldn't break this off into it's own repository and create a GitHub Action for it? If we do it like then it can be easily used by others, and in other repositories for EasyStack files.

dagonzalezfo:
Covered by  Mr 675

PR #460, opened 2024-01-23
only pick a software subdir that is present in the CVMFS repository

trz42:
Without the change, a software subdir that is not present in the CVMFS repository could be chosen.


boegel (approved):
lgtm

PR #475, opened 2024-02-15
Always set `$LMOD_RC` (even when `EESSI_BASIC_ENV` is set), and make sure loading CUDA apps always work when building

ocaisa:


boegel (approved):
lgtm

PR #482, opened 2024-02-29
Temporary fix for CI by pinning archspec version

ocaisa:


boegel (approved):
lgtm

boegel:
@ocaisa Let's also open an issue with more info on why CI with latest `archspec` is broken (as a reminder to fix it)?

ocaisa:
I'm already working on a PR to fix, plan is to also do that today

PR #490, opened 2024-03-04
Workaround for LMOD only allowing a single hook to be registered, which made our first hook unused

casparvl:
Use LMOD_CONFIG_DIR instead of LMOD_RC to allow user overrides of what we specify in our RC file, see search order on https://lmod.readthedocs.io/en/latest/145_properties.html?highlight=lmod_rc#the-properties-file-lmodrc-lua . Also, prefix our hooks with eessi_ to avoid unintentional clashes in namespace between site lmod configuration and eessi lmod configuration

### DIFF
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -85,6 +85,7 @@ if [ -d $EESSI_PREFIX ]; then
           # Allow for site-specific lmod configuration
           host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
           if [ -f $host_lmod_rc ]; then
+              show_msg "Found site Lmod configuration file at $LMOD_RC"
               export LMOD_RC="$host_lmod_rc"
           fi
 
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -82,6 +82,11 @@ if [ -d $EESSI_PREFIX ]; then
           else
             error "Lmod configuration file not found at $LMOD_RC"
           fi
+          # Allow for site-specific lmod configuration
+          host_lmod_rc="$EESSI_CVMFS_REPO"/host_injections/.lmod/lmodrc.lua
+          if [ -f $host_lmod_rc ]; then
+              export LMOD_RC="$host_lmod_rc"
+          fi
 
         else
           error "EESSI software layer at $EESSI_SOFTWARE_PATH not found!"
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -83,7 +83,7 @@ if [ -d $EESSI_PREFIX ]; then
             error "Lmod configuration file not found at $LMOD_RC"
           fi
           # Allow for site-specific lmod configuration
-          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}/.lmod/lmodrc.lua
+          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
           if [ -f $host_lmod_rc ]; then
               export LMOD_RC="$host_lmod_rc"
           fi
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -83,7 +83,7 @@ if [ -d $EESSI_PREFIX ]; then
             error "Lmod configuration file not found at $LMOD_RC"
           fi
           # Allow for site-specific lmod configuration
-          host_lmod_rc="$EESSI_CVMFS_REPO"/host_injections/.lmod/lmodrc.lua
+          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}/.lmod/lmodrc.lua
           if [ -f $host_lmod_rc ]; then
               export LMOD_RC="$host_lmod_rc"
           fi
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -76,8 +76,8 @@ if [ -d $EESSI_PREFIX ]; then
             false
           fi
 
-          export LMOD_RC="$EESSI_SOFTWARE_PATH/.lmod/lmodrc.lua"
-          if [ -f $LMOD_RC ]; then
+          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
             show_msg "Found Lmod configuration file at $LMOD_RC"
           else
             error "Lmod configuration file not found at $LMOD_RC"
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -78,9 +78,9 @@ if [ -d $EESSI_PREFIX ]; then
 
           export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
           if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
-            show_msg "Found Lmod configuration file at $LMOD_RC"
+            show_msg "Found Lmod configuration file at $LMOD_CONFIG_DIR/lmodrc.lua"
           else
-            error "Lmod configuration file not found at $LMOD_RC"
+            error "Lmod configuration file not found at $LMOD_CONFIG_DIR/lmodrc.lua"
           fi
           # Allow for site-specific lmod configuration
           host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -76,17 +76,11 @@ if [ -d $EESSI_PREFIX ]; then
             false
           fi
 
-          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
-          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
-            show_msg "Found Lmod configuration file at $LMOD_CONFIG_DIR/lmodrc.lua"
+          export LMOD_RC="$EESSI_SOFTWARE_PATH/.lmod/lmodrc.lua"
+          if [ -f $LMOD_RC ]; then
+            show_msg "Found Lmod configuration file at $LMOD_RC"
           else
-            error "Lmod configuration file not found at $LMOD_CONFIG_DIR/lmodrc.lua"
-          fi
-          # Allow for site-specific lmod configuration
-          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
-          if [ -f $host_lmod_rc ]; then
-              show_msg "Found site Lmod configuration file at $LMOD_RC"
-              export LMOD_RC="$host_lmod_rc"
+            error "Lmod configuration file not found at $LMOD_RC"
           fi
 
         else
--- a/create_lmodrc.py
+++ b/create_lmodrc.py
@@ -29,7 +29,7 @@ local function read_file(path)
     return content
 end
 
-local function cuda_enabled_load_hook(t)
+local function eessi_cuda_enabled_load_hook(t)
     local frameStk  = require("FrameStk"):singleton()
     local mt        = frameStk:mt()
     local simpleName = string.match(t.modFullName, "(.-)/")
@@ -94,7 +94,7 @@ local function cuda_enabled_load_hook(t)
     end
 end
 
-local function openmpi_load_hook(t)
+local function eessi_openmpi_load_hook(t)
     -- disable smcuda BTL when loading OpenMPI module for aarch64/neoverse_v1,
     -- to work around hang/crash due to bug in OpenMPI;
     -- see https://gitlab.com/eessi/support/-/issues/41
@@ -114,8 +114,8 @@ local function openmpi_load_hook(t)
     end
 end
 
-hook.register("load", cuda_enabled_load_hook)
-hook.register("load", openmpi_load_hook)
+hook.register("load", eessi_cuda_enabled_load_hook)
+hook.register("load", eessi_openmpi_load_hook)
 """
 
 def error(msg):
--- a/create_lmodrc.py
+++ b/create_lmodrc.py
@@ -114,8 +114,15 @@ local function eessi_openmpi_load_hook(t)
     end
 end
 
-hook.register("load", eessi_cuda_enabled_load_hook)
-hook.register("load", eessi_openmpi_load_hook)
+-- Combine both functions into a single one, as we can only register one function as load hook in lmod
+-- Also: make it non-local, so it can be imported and extended by other lmodrc files if needed
+function eessi_load_hook(t)
+    eessi_cuda_enabled_load_hook(t)
+    eessi_openmpi_load_hook(t)
+end
+
+
+hook.register("load", eessi_load_hook)
 """
 
 def error(msg):


casparvl:
Let me test first if I can indeed extend this hook... if extending it doesn't work anyway, the `eessi_load_hook` might as well be made a local function.

casparvl:
I _can_ extend this hook. A solution that works is to have in the `EESSI` `lmodrc` file:

```
-- Combine both functions into a single one, as we can only register one function as load hook in lmod
-- Also: make it non-local, so it can be imported and extended by other lmodrc files if needed
function eessi_load_hook(t)
    eessi_cuda_enabled_load_hook(t)
    eessi_openmpi_load_hook(t)
end

hook.register("load", eessi_load_hook)
```
And then in the site-specific `lmodrc` file:
```
lmod_config_dir = os.getenv("LMOD_CONFIG_DIR")
dofile(lmod_config_dir.."/lmodrc.lua")

local function combined_load_hook(t)
    eessi_load_hook(t)
    site_specific_hook_a(t)
end
hook.register("load", combined_load_hook)
```
(N.B. this is a minimal code example, in reality you'd want to do checking if `LMOD_CONFIG_DIR` is defined, if `lmod_config_dir.."/lmodrc.lua" exists, etc).

While this works and is a solution we can use already _today_, I've also opened a discussion upstream to see if `lmod` can be extended to facilitate this in a more natural way: https://github.com/TACC/Lmod/issues/694

ocaisa:
Updating the Lmod version is going to take a while, I suggest we make this available now as it helps record what the necessary steps are on top of anything else.

It's no longer WIP, is it?

casparvl:
You're right, it isn't. I wanted to test it first, but my message from 3 hours ago was the conclusion that that works :D so, this is ready for review.

casparvl:
One thing I'm wondering: should we set `LMOD_RC` to `.../host_injections/.lmod` by default? As a site, I probably want this `LMOD_RC` to be set if (and only if) my user is using the `EESSI` software stack... How else would I achieve that, other than it being set in the EESSI init script?

This way, I can just put a local `lmodrc.lua` file in `.../host_injections/.lmod` and be done with it - nothing else needed from the host site.

casparvl:
To answer my own question: yes, I think we should do this, and I added it in the commit. Our docs should then just say "If you want to customize lmod behaviour for the EESSI environment, place an `lmodrc.lua` in `host_injections/.lmod`" or something of that nature.

init/eessi_environment_variables
```diff
@@ -76,12 +76,17 @@ if [ -d $EESSI_PREFIX ]; then
...
-          export LMOD_RC="$EESSI_SOFTWARE_PATH/.lmod/lmodrc.lua"
-          if [ -f $LMOD_RC ]; then
+          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
             show_msg "Found Lmod configuration file at $LMOD_RC"
           else
             error "Lmod configuration file not found at $LMOD_RC"
           fi
+          # Allow for site-specific lmod configuration
+          host_lmod_rc="$EESSI_CVMFS_REPO"/host_injections/.lmod/lmodrc.lua
```
ocaisa, line 86: I like the idea in principle, but I'd make it sit within a host_injections equivalent of LMOD_CONFIG_DIR

ocaisa (changes requested):

init/eessi_environment_variables
casparvl: You mean: nested, so that you have e.g. a `/cvmfs/software.eessi.io/host_injections/2023.06/software/linux/x86_64/amd/zen2/.lmod/lmodrc.lua` ?

We could. But do you really need to be able to make architecture-specific hooks? In EESSI itself, we only make _one_ hook file with `create_lmodrc.py` and use logic to do architecture specific things like `if (moduleName == "OpenMPI") and (cpuTarget == "aarch64/neoverse_v1") then ...`. Then we just duplicate the `lmodrc.lua` to all prefixes

The downside of having this nested in an architecture prefix is that you might need to do a lot of duplication (or at the very least symlinking) as a host site if you have a heterogenous system. For the Snellius system at SURF, it'd already mean putting the same `lmodrc.lua` in two different prefixes today, and if we have a `zen4` tree tomorrow, I'll also need to _remember_ to deploy it there...

casparvl:
I'm working on the upstream issue of allowing the registration of multiple hooks of the same type by putting them into a table (instead of overwriting). In the process, I have installed a newer version of `lmod` locally, and I'm seeing that my hooks are being called way less often:
```
{EESSI 2023.06} [casparl@gcn6 ~]$ module load OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1
Loading eessi hooks
in OpenMPI: Loading GCCcore
in OpenMPI: Loading GCC
in OpenMPI: Loading numactl
in OpenMPI: Loading libxml2
in OpenMPI: Loading libpciaccess
in OpenMPI: Loading hwloc
in OpenMPI: Loading OpenSSL
in OpenMPI: Loading libevent
in OpenMPI: Loading UCX
in OpenMPI: Loading libfabric
in OpenMPI: Loading PMIx
in OpenMPI: Loading UCC
in OpenMPI: Loading OpenMPI
in OpenMPI: Loading gompi
in OpenMPI: Loading GDRCopy
in OpenMPI: Loading UCX-CUDA
in OpenMPI: Loading NCCL
in OpenMPI: Loading UCC-CUDA
in OpenMPI: Loading OSU-Micro-Benchmarks
{EESSI 2023.06} [casparl@gcn6 software-layer]$ module --version

Modules based on Lua: Version 8.7.23  2023-03-29 17:19 -05:00
    by Robert McLay mclay@tacc.utexas.edu

{EESSI 2023.06} [casparl@gcn6 ~]$ vi .lmodrc.lua
{EESSI 2023.06} [casparl@gcn6 ~]$ source $HOME/lmod_install/lmod/lmod/init/bash
{EESSI 2023.06} [casparl@gcn6 ~]$ echo $LMOD_CMD
/home/casparl/lmod_install/lmod/lmod/libexec/lmod
{EESSI 2023.06} [casparl@gcn6 ~]$ module load OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1
Loading eessi hooks
in OpenMPI: Loading OSU-Micro-Benchmarks
{EESSI 2023.06} [casparl@gcn6 ~]$ module purge
Loading eessi hooks
{EESSI 2023.06} [casparl@gcn6 ~]$ module load OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1
Loading eessi hooks
in OpenMPI: Loading UCX-CUDA
in OpenMPI: Loading NCCL
in OpenMPI: Loading UCC-CUDA
in OpenMPI: Loading OSU-Micro-Benchmarks
{EESSI 2023.06} [casparl@gcn6 ~]$ module --version

Modules based on Lua: Version 8.7.34 (a97d6819) 2024-01-05 19:34 -07:00
    by Robert McLay mclay@tacc.utexas.edu
```
It seems the load hook is now only called if for modules with the GPU properties assigned. I'm wondering if this may be the result of a change in Lmod of _when_ the `lmodrc.lua` files are sourced/used. As I learned in the Lmod monthly meeting, they are _not_ the designated place for hooks: strictly speaking, they are _only_ there for defining custom module properties (see https://lmod.readthedocs.io/en/latest/145_properties.html?highlight=lmodrc ). I can imagine Lmod now does something like "does this module have a property?" and only _then_ source the `lmodrc` files to get the definition of those properties. This is more efficient, since it means these `lmodrc` files don't need to be read _for every module_.

The _documented_ place to specify hooks is `Site_Package.lua`, see https://lmod.readthedocs.io/en/latest/170_hooks.html?highlight=hook#hook-functions . The disadvantage: there is no hierarchy in this, you can only ever have one `Site_Package.lua`. Of course, we _could_ create a `Site_Package.lua` in EESSI and have that `source` a `Site_Package.lua` from some prefix in `host_injections`...

I'll confirm if the version is the issue by installing different versions of lmod and seeing if I get the original behaviour back...

casparvl:
Ok, I'm stumped. I installed the `8.7.23` version that we also have in the compat layer from a git clone:
```
 1097  git checkout 8.7.23
 1098  ./configure --prefix $HOME/lmod_install
 1099  make install
```
Running with that installation, I still only see the modules with a GPU property invoking the hook:
```
{EESSI 2023.06} [casparl@gcn6 ~]$ module --version

Modules based on Lua: Version 8.7.24  2023-05-04 15:12 -05:00
    by Robert McLay mclay@tacc.utexas.edu

{EESSI 2023.06} [casparl@gcn6 ~]$ module purge
{EESSI 2023.06} [casparl@gcn6 ~]$ module load OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1
Load hook B called on UCX-CUDA
Load hook B called on NCCL
Load hook B called on UCC-CUDA
Load hook B called on OSU-Micro-Benchmarks
```
Why? What's different? I don't know...

casparvl:
Inspecting the debugging output with:
```
module --debug=3 load OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1 2>&1 | less
```
I find:
```
                                  add_property(arch, gpu){
                                    MT:add_property("UCX-CUDA", "arch", "gpu"){
                                      ReadLmodRC:singleton(){
                                      } ReadLmodRC:singleton
                                      ReadLmodRC:validPropValue("arch", "gpu", t){
                                      } ReadLmodRC:validPropValue
                                    } MT:add_property
                                  } add_property
```
I.e. it is indeed the `add_property` in the module which causes `ReadLmodRC:singleton()` to be invoked. This then triggers a read of all the LmodRC files in the documented order. Now my only question is: what triggers this `ReadLmodRC:singleton()` for our EESSI installation? That seems to be the following:
```
          Hub:load(mA={OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1}){
            Hub:load i: 1, userName: OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1
            Cache:singleton(){
              Cache:l_new(){
                ReadLmodRC:singleton(){
                } ReadLmodRC:singleton
                #scDescriptT: 3
                Adding: dir: /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/.lmod/cache, timestamp: 1709303487
                Adding: dir: /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/.lmod/cache, timestamp: 1709303487
              } Cache:l_new
              s_cache.buildCache: nil
            } Cache:singleton
```
Which indeed happens very early in the trace. It is part of initializing the `Cache:singleton()`. In that sense, we are just _lucky_ that it is being called _before_ any of the `module load`'s are called, i.e. that the hooks are registered 'in time'.

casparvl:
FYI: I made two PRs with two possible approaches for supporting registration of more than one function as a hook:
https://github.com/TACC/Lmod/pull/695
https://github.com/TACC/Lmod/pull/696
Whichever of those two approaches we go down, it should make adding (appending) site-customized hooks a bit easier... :)

init/eessi_environment_variables
```diff
@@ -76,12 +76,17 @@ if [ -d $EESSI_PREFIX ]; then
...
-          export LMOD_RC="$EESSI_SOFTWARE_PATH/.lmod/lmodrc.lua"
-          if [ -f $LMOD_RC ]; then
+          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
             show_msg "Found Lmod configuration file at $LMOD_RC"
           else
             error "Lmod configuration file not found at $LMOD_RC"
           fi
+          # Allow for site-specific lmod configuration
+          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}/.lmod/lmodrc.lua
```
ocaisa, line 86:
```suggestion
          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
```

ocaisa:
We'll have to make the switch to using `SitePackage.lua` at some point (issue opened in #491), but this fix is what we need today

ocaisa (dismissed):

ocaisa:

init/eessi_environment_variables
```diff
@@ -76,12 +76,17 @@ if [ -d $EESSI_PREFIX ]; then
...
           fi
 
-          export LMOD_RC="$EESSI_SOFTWARE_PATH/.lmod/lmodrc.lua"
-          if [ -f $LMOD_RC ]; then
+          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
             show_msg "Found Lmod configuration file at $LMOD_RC"
           else
             error "Lmod configuration file not found at $LMOD_RC"
           fi
```
ocaisa, line 84:
```suggestion
          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
            show_msg "Found Lmod configuration file at $LMOD_CONFIG_DIR/lmodrc.lua"
          else
            error "Lmod configuration file not found at $LMOD_CONFIG_DIR/lmodrc.lua"
          fi
```

init/eessi_environment_variables
```diff
@@ -76,11 +76,16 @@ if [ -d $EESSI_PREFIX ]; then
...
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
+            show_msg "Found Lmod configuration file at $LMOD_CONFIG_DIR/lmodrc.lua"
           else
-            error "Lmod configuration file not found at $LMOD_RC"
+            error "Lmod configuration file not found at $LMOD_CONFIG_DIR/lmodrc.lua"
+          fi
+          # Allow for site-specific lmod configuration
+          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
+          if [ -f $host_lmod_rc ]; then
+              export LMOD_RC="$host_lmod_rc"
```
ocaisa, line 88:
```suggestion
              show_msg "Found site Lmod configuration file at $LMOD_RC"
              export LMOD_RC="$host_lmod_rc"
```

ocaisa (approved):

init/eessi_environment_variables
```diff
@@ -76,11 +76,17 @@ if [ -d $EESSI_PREFIX ]; then
...
-            show_msg "Found Lmod configuration file at $LMOD_RC"
+          export LMOD_CONFIG_DIR="$EESSI_SOFTWARE_PATH/.lmod"
+          if [ -f $LMOD_CONFIG_DIR/lmodrc.lua ]; then
+            show_msg "Found Lmod configuration file at $LMOD_CONFIG_DIR/lmodrc.lua"
           else
-            error "Lmod configuration file not found at $LMOD_RC"
+            error "Lmod configuration file not found at $LMOD_CONFIG_DIR/lmodrc.lua"
+          fi
+          # Allow for site-specific lmod configuration
+          host_lmod_rc="${EESSI_PREFIX/versions/host_injections}"/.lmod/lmodrc.lua
```
ocaisa, line 86: While I get that there probably is no good reason to have separate user hooks, I think we can do better here. Let's first try `"${LMOD_CONFIG_DIR/versions/host_injections}"/lmodrc.lua` and then fall back to the more global location if that doesn't exist, that way sites can have arch-specific customisations.

ocaisa (approved):

ocaisa (changes requested):
Meant to request a change in last review

ocaisa:
@casparvl The changes to the initialisation script are not being picked up (according to https://github.com/EESSI/software-layer/pull/490#issuecomment-1981053047). I'm guessing this is deliberate behaviour of our tarring script...but I don't know how to deal with it.

ocaisa:
@casparvl Thinking about this again, I would revert all your changes to `init/eessi_environment_variables` and we focus solely on fixing the immediate problem. We can't integrate them anyway without jumping through some hoops, and in fact they can be part of the follow-up PR to move to using `SitePackage.lua` and setting `LMOD_PACKAGE_PATH`

init/eessi_environment_variables
casparvl: That's a good idea, we can make it hierarchical ourselves. If the more specific one doesn't exist, we just pick up on the generic one. I'll change it.

casparvl:
Also true, ok, I'm stripping that and limiting it to the actual fix.

ocaisa:
bot: build repo:eessi.io-2023.06-software

ocaisa (approved):

PR #479, opened 2024-02-20
{2023.06}[foss/2023b] WSClean 3.4 + DP3 6.0

t1mk1k:
Adding WSClean and DP3 to software.eessi.io.

Including the AOFlagger dependency because of previous issues with an older Lua version in the compat layer.

PR #493, opened 2024-03-08
Allow for open PRs when checking missing installations in build script

ocaisa:
As mentioned in #479 , with the changes we've made to checking for missing installations the errors from the bot are cryptic when a referenced PR is not merged in EasyBuild. This PR allows for open PRs during the build, but still fails them in GitHub Actions (with a bit more helpful text).

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -237,7 +237,7 @@ else
                 copy_build_log "${eb_last_log}" "${build_logs_dir}"
             fi
     
-            $TOPDIR/check_missing_installations.sh ${TOPDIR}/${easystack_file}
+            $TOPDIR/check_missing_installations.sh ${TOPDIR}/${easystack_file} ${TOPDIR}/${pr_diff}
         else
             fatal_error "Easystack file ${easystack_file} not found!"
         fi
--- a/check_missing_installations.sh
+++ b/check_missing_installations.sh
@@ -10,8 +10,15 @@
 
 TOPDIR=$(dirname $(realpath $0))
 
-if [ $# -ne 1 ]; then
-    echo "ERROR: Usage: $0 <path to easystack file>" >&2
+if [ "$#" -eq 1 ]; then
+    true
+elif [ "$#" -eq 2 ]; then
+    echo "Using $2 to give create exceptions for PR filtering of easystack"
+    # Find lines that are added and use from-pr, make them unique, grab the
+    # PR numbers and use them to construct something we can use within awk
+    pr_exceptions=$(grep ^+ $2 | grep from-pr | uniq | awk '{print $3}' | xargs -i echo " || /'{}'/")
+else
+    echo "ERROR: Usage: $0 <path to easystack file> (<optional path to PR diff>)" >&2
     exit 1
 fi
 easystack=$1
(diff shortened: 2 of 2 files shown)


casparvl (commented):
I've tried this interactively using an EasyStack file containing:
```
easyconfigs:
  - GCC-13.2.0.eb
  - foss-2023b.eb
  - SciPy-bundle-2023.11-gfbf-2023b.eb
  - netCDF-4.9.2-gompi-2023b.eb:
      options:
        from-pr: 19534
  - matplotlib-3.8.2-gfbf-2023b.eb:
      options:
        from-pr: 19552
  - CDO-2.2.2-gompi-2023b.eb:
      options:
        from-pr: 19792
  - CFITSIO-4.3.1-GCCcore-13.2.0.eb:
      options:
        from-pr: 19840
```
I downloaded `wget https://github.com/EESSI/software-layer/pull/479.diff` to get a diff that would create an exception for `CFITSIO`.

Then, I ran 
```
./check_missing_installations.sh easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023b.yml 479.diff
```
interactively. That resulted in a succesful run, i.e. it _did not_ print anything with the message `(are you sure all PRs referenced have been merged in EasyBuild?)`.

I'm not sure if this was your intention?

Also, you mentioned that this will fail in Github Actions, but I don't see any change to the github CI. It also made me think: that might be too late, as we do the `deploy` even when github CI is failing (we are used to it, since one of the CI checks is 'was this PR deployed' for every architecture). 

As mentioned in https://github.com/EESSI/software-layer/pull/479#issuecomment-1985367836 I don't disagree with the fact that it results in a failure (so maybe we should) _not_ make an exception, but I would like to give a more clear message (reported back to the PR by the bot) as to why it failed.

I'm currently working on something that'll add a "Reason" to the failure, similar to what I did for the test step in https://github.com/EESSI/software-layer/pull/467/files . I think the work you did here is going to allign well with that (I'd feel bad for telling you it wouldn't after you spend a day on this ;-)). But, it would require some slight modification: 
- create a 2nd EasyStack file in which the exception for the `--from-pr` is made
- do your original thing first, i.e. run _only_ against develop. 
- If the exist status is 1 _and_ if the output file contains a pattern like `ERROR: One or more files not found: AOFlagger-3.4.0-foss-2023b.eb (search paths: /tmp/tmp.d83nFSn6v6/easyconfigs/easybuild/easyconfigs)`, _then_ we want to run again _with_ your exception. 
- If that 2nd succeeds, we know the problem was an umerged PR, and we print the failure message you envisioned  `...(are you sure all PRs referenced have been merged in EasyBuild?)`

In my own PR (that modifies `check-build`, I can pick up on _that_ pattern (which is very specific), and have the bot print it as "Reason" in it's response in the PR. My own PR isn't there yet, but you can take a sneak peek in https://github.com/casparvl/software-layer/tree/improve_error_on_unmerged_pr

casparvl:
N.B. After finishing my part, I'll make a PR to your feature branch to make the changes I suggest above. That should make more clear what I mean :)

ocaisa:
This does work as intended, when the script is called by the bot it will allow the open PR and pass. There is another _GitHub Action_ that we call that will _only_ check against develop...so this will stop us from merging something thay has not been integrated upstream

casparvl:
To answer my own first question: I see you add that to the message if the `check_missing_installation.sh` script is called _without_ a PR diff, i.e. without exceptions. Ok, then I see what your intention was with this, it's just a different one than I anticipated :) You wanted to make it pass (provided the callee passes the prdiff to allow for exceptions). My idea would still be to make it fail, to prevent Reviewers from `deploy`-ing the PR. (does the bot do a hard-fail on deploying tarballs if their latest status was `FAILURE`? I don't know, I assumed/hoped so, but maybe not...)

casparvl:
Ok, https://github.com/EESSI/software-layer/pull/494#issuecomment-1986853137 does what I want now - see the new "Reason" header in there.

Honestly, I think we should try to more extensively `grep` stuff so that we can expose better to the end-user what went wrong during a build _without_ the end-user immediately having to ask us (for context from the `slum-XYZ.log`) or having to immediately try to replicate locally with https://www.eessi.io/docs/adding_software/debugging_failed_builds/ . Both add quite a bit of latency to figuring out what's wrong... :)

ocaisa:
This has been rolled into #494 so closing this

PR #494, opened 2024-03-09
Improve error reported by bot when using unmerged EasyBuild PR in easystack file

casparvl:
https://github.com/EESSI/software-layer/commit/301ab593867335700b576cc9949fe58c408325e7 prevents deploying software for which the PR in EasyBuild is still open. In principle, that's good, but it would lead to failures in the EESSI PR which are unclear to the end user, e.g. as in https://github.com/EESSI/software-layer/pull/479#issuecomment-1981272433 . In these failing cases, the only thing you see is the status is failure, and the details reflect that `ERROR:` was found, as well as a non-zero excit code on `check_missing_installations`. 

This is quite unspecific if the only issue is an umerged PR. It should be clear to the end user that the _build_ succeeded, that all the software ended up in the tarball, but the _only_ issue is an unmerged PR.

In this PR, I use inspiration from @ocaisa 's https://github.com/EESSI/software-layer/pull/493 . He made it possible to make an exception for `--from-pr` statements that are added in the `software-layer` PR that tries to add software. In this PR, I use that conditionally: 
- First, we check for missing installations as was implemented in https://github.com/EESSI/software-layer/commit/301ab593867335700b576cc9949fe58c408325e7 . I.e. we check using the EasyConfigs from `develop`. If there is an unmerged EasyBuild PR, this will fail.
- Upon failure, I use @ocaisa 's logic to run the same check, but now with an exceptions for the `--from-pr` statements added in the `software-layer` PR. If that _does_ pass, it is clear that some of the EasyBuild PR's referred to in this `software-layer` PR are not merged yet. We can then print a clear message in the output file. 
- The final step is that in the `check-build.sh`, I pick up on this pattern. if it is present, I add a `reason` for the failure.

Note that I also refactored the `check-build.sh` a bit. The definition and useage of certain variables was far apart, making it hard to trace what was happening. I've refactored that so that things are defined and used 'in order', which makes it easier to understand what is happening in the code.

### DIFF
--- a/bot/check-build.sh
+++ b/bot/check-build.sh
@@ -207,6 +207,8 @@ elif [[ ${UNMERGED_EASYCONFIG} -eq 1 ]]; then
 else
     # FAILURE
     status="FAILURE"
+    # General failure, we don't know a more specific reason
+    reason=""
     summary=":cry: FAILURE"
 fi
 
--- a/bot/check-build.sh
+++ b/bot/check-build.sh
@@ -174,7 +174,7 @@ fi
 # For now, we only analyse unmerged EasyConfigs as potential cause, but we can easily add checks for other
 # specific scenarios below
 
-# Check for the pattern being added here https://github.com/EESSI/software-layer/pull/493 to the output to
+# Check for the pattern being added here by check_missing_installations.sh to the output to
 # see if EasyConfigs might have been unmerged, and that's causing a failure
 UNMERGED_EASYCONFIG=-1
 if [[ ${SLURM_OUTPUT_FOUND} -eq 1 ]]; then
--- a/bot/check-build.sh
+++ b/bot/check-build.sh
@@ -98,15 +98,15 @@ job_dir=${PWD}
 job_out="slurm-${SLURM_JOB_ID}.out"
 [[ ${VERBOSE} -ne 0 ]] && echo ">> searching for job output file(s) matching '"${job_out}"'"
 if  [[ -f ${job_out} ]]; then
-    SLURM=1
+    SLURM_OUTPUT_FOUND=1
     [[ ${VERBOSE} -ne 0 ]] && echo "   found slurm output file '"${job_out}"'"
 else
-    SLURM=0
+    SLURM_OUTPUT_FOUND=0
     [[ ${VERBOSE} -ne 0 ]] && echo "   Slurm output file '"${job_out}"' NOT found"
 fi
 
 ERROR=-1
-if [[ ${SLURM} -eq 1 ]]; then
+if [[ ${SLURM_OUTPUT_FOUND} -eq 1 ]]; then
   GP_error='ERROR: '
   grep_out=$(grep -v "^>> searching for " ${job_dir}/${job_out} | grep "${GP_error}")
   [[ $? -eq 0 ]] && ERROR=1 || ERROR=0
@@ -116,7 +116,7 @@ if [[ ${SLURM} -eq 1 ]]; then
 fi
 
 FAILED=-1
-if [[ ${SLURM} -eq 1 ]]; then
+if [[ ${SLURM_OUTPUT_FOUND} -eq 1 ]]; then
   GP_failed='FAILED: '
   grep_out=$(grep -v "^>> searching for " ${job_dir}/${job_out} | grep "${GP_failed}")
   [[ $? -eq 0 ]] && FAILED=1 || FAILED=0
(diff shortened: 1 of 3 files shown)


casparvl:

casparvl:

casparvl:

casparvl:

bot/check-build.sh
```diff
@@ -332,42 +349,65 @@ echo -n "comment_description = " >> ${job_result_file}
...
+comment_details_list=${comment_details_list}$(add_detail ${NO_MISSING} 1 "${success_msg}" "${failure_msg}")
 
 success_msg="found message matching <code>${GP_tgz_created}</code>"
 failure_msg="no message matching <code>${GP_tgz_created}</code>"
-CoDeList=${CoDeList}$(add_detail ${TGZ} 1 "${success_msg}" "${failure_msg}")
-
-comment_details="${comment_details_fmt/__DETAILS_LIST__/${CoDeList}}"
-
+comment_details_list=${comment_details_list}$(add_detail ${TGZ} 1 "${success_msg}" "${failure_msg}")
+# Now, do the actual repalcement of __DETAILS_FMT__
```
ocaisa, line 403:
```suggestion
# Now, do the actual replacement of __DETAILS_FMT__
```

check_missing_installations.sh
```diff
@@ -40,6 +47,28 @@ exit_code=${PIPESTATUS[0]}
 
 ok_msg="Command 'eb --missing ...' succeeded, analysing output..."
 fail_msg="Command 'eb --missing ...' failed, check log '${eb_missing_out}'"
+if [ "$exit_code" -eq 1 ]; then
```
ocaisa, line 50: Shouldn't this also be conditional on `[ "$#" -eq 2 ]`?

casparvl: Good point, we can't do that check if the second argument is missing. I'll fix it...

ocaisa: Maybe just check for the existence of `pr_exceptions`, that's a bit more obvious

casparvl: I actually added `&& [ ! -z "$pr_exceptions" ]` instead. If no 2nd argument was passed, `pr_exceptions` will be empty and the `if` condition will be false. But _also_ if a 2nd argument was passed and there _were no `--from-pr` statements_ in the current PR, we can save ourselves the time and skip the second `eb --missing` check, as unmerged PRs cannot be the reason for failure in that scenario.

ocaisa:

ocaisa:

easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023b.yml
```diff
@@ -49,3 +49,6 @@ easyconfigs:
   - libspatialindex-1.9.3-GCCcore-13.2.0.eb:
       options:
         from-pr: 19922
+  - CFITSIO-4.3.1-GCCcore-13.2.0.eb:
+      options:
+        from-pr: 19840
```
ocaisa, line 54: Assuming this needs to be removed before we consider merging

boegel, line 54: This doesn't make much sense anymore, since `CFITSIO/4.3.1-GCCcore-13.2.0` is already installed (via #479)

```suggestion
```

boegel (changes requested):

check_missing_installations.sh
```diff
@@ -40,6 +47,30 @@ exit_code=${PIPESTATUS[0]}
 
 ok_msg="Command 'eb --missing ...' succeeded, analysing output..."
 fail_msg="Command 'eb --missing ...' failed, check log '${eb_missing_out}'"
+if [ "$exit_code" -eq 1 ] && [ ! -z $pr_exceptions ]; then
```
boegel, line 50: Should this be `-ne 0` (EasyBuild but `exit 1` now, but it may be different in the future, see https://github.com/easybuilders/easybuild-framework/issues/4426)

```suggestion
if [ "$exit_code" -ne 0 ] && [ ! -z $pr_exceptions ]; then

File: bot/check-build.sh [hunk 5]
```diff
@@ -169,9 +169,27 @@ fi
 [[ ${VERBOSE} -ne 0 ]] && echo "  NO_MISSING.: $([[ $NO_MISSING -eq 1 ]] && echo 'yes' || echo 'no') (yes)"
 [[ ${VERBOSE} -ne 0 ]] && echo "  TGZ_CREATED: $([[ $TGZ -eq 1 ]] && echo 'yes' || echo 'no') (yes)"
 
+# Here, we try to do some additional analysis on the output file
+# to see if we can print a more clear 'reason' for the failure
+# For now, we only analyse unmerged EasyConfigs as potential cause, but we can easily add checks for other
+# specific scenarios below
+
+# Check for the pattern being added here https://github.com/EESSI/software-layer/pull/493 to the output to
```
boegel, line 177:
```suggestion
# Check for the pattern being added here by check_missing_installations.sh to the output to

boegel (changes requested):

boegel (approved):

ocaisa:
@casparvl Do you think there is something wrong with this as merged? We were planning to test it via #499 but it is not actually commenting in the tests (see https://github.com/EESSI/software-layer/pull/499#issuecomment-1994612204 for example)

PR #488, opened 2024-03-01
Add functionality for rebuilding software: try it on OpenMPI 4.1.x to fix `smcuda` issue

bedroge:
Attempt to implement a solution for #312 with `--fakeroot`.

PR #500, opened 2024-03-13
Make OpenMPI workaround apply to all OpenMPI

ocaisa:
Not sure if we should instead just get #488 over the line

### DIFF
--- a/create_lmodrc.py
+++ b/create_lmodrc.py
@@ -95,14 +95,13 @@ local function eessi_cuda_enabled_load_hook(t)
 end
 
 local function eessi_openmpi_load_hook(t)
-    -- disable smcuda BTL when loading OpenMPI module for aarch64/neoverse_v1,
+    -- disable smcuda BTL when loading OpenMPI module,
     -- to work around hang/crash due to bug in OpenMPI;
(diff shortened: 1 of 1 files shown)


create_lmodrc.py
```diff
@@ -95,14 +95,13 @@
 end
 
 local function eessi_openmpi_load_hook(t)
-    -- disable smcuda BTL when loading OpenMPI module for aarch64/neoverse_v1,
+    -- disable smcuda BTL when loading OpenMPI module,
     -- to work around hang/crash due to bug in OpenMPI;
-    -- see https://gitlab.com/eessi/support/-/issues/41
+    -- Fix should be merged via https://github.com/EESSI/software-layer/pull/488
```
boegel, line 100: I would keep the reference to the support ticket?
```suggestion
    -- see https://gitlab.com/eessi/support/-/issues/41
    -- Fix should be merged via https://github.com/EESSI/software-layer/pull/488

boegel (changes requested):

bedroge:
#488 got merged and deployed, so this PR can be closed now.

PR #486, opened 2024-02-29
Make check_missing_installations.sh check against develop branch of EasyBuild

ocaisa:

PR #501, opened 2024-03-13
replace Z3 dependency for `PyTorch-2.1.2-foss-2023a.eb`

trz42:
A recent change to the easyconfig `PyTorch-2.1.2-foss-2023a.eb` (see https://github.com/easybuilders/easybuild-easyconfigs/pull/20050) lets the check for missing installations fail, for example, see
https://github.com/EESSI/software-layer/actions/runs/8268611041/job/22621916438?pr=500#step:4:2130

This PR adds a hook that attempts to replace the dependency for Z3 with the version that was used before (with suffix `-Python-3.11.3`).

CI should not fail anymore. However, running `eb` with `--missing` prints messages that the dependency has been replaced, but the renamed dependency is then not listed anymore. Running with `--dry-run` does show the renamed dependency. Not sure if this behaviour is expected or not.

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -249,6 +249,30 @@ def parse_hook_pybind11_replace_catch2(ec, eprefix):
             build_deps[idx] = (catch2_name, catch2_version)
 
 
+def parse_hook_pytorch_replace_z3(ec, eprefix):
+    """
+    Replace Z3 dependency in PyTorch/2.1.2 easyconfig to work around a change
+    of the Z3 eb/module name
+    cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/20050
+    """
+    # this is mainly necessary to avoid that --missing keeps reporting Z3/4.12.2
+    # is missing, because previously it was Z3/4.12.2-GCCcore-12.3.0-Python-3.11.3
+    if ec.name == 'PyTorch' and ec.version in ['2.1.2']:
+        deps = ec['dependencies']
+        z3_dep = None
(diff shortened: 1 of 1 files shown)


casparvl:
Hm, this is really due to the difference between the set of EasyConfigs used to _build_ and the set of EasyConfig used when running `--missing`. 

For the build, because PyTorch is in https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a.yml , we run `EasyBuild` version `4.9.0`. For the `--missing`, we run the EasyConfigs from the develop branch, since @ocaisa 's https://github.com/EESSI/software-layer/pull/486 .

Make me wonder if this actually makes sense. There was a good reason for it, as it allowed us to strip a lot of `--from-pr`'s, and enforce that EasyConfigs should be merged. However, it is fundamentally inconsistent to use different versions of the EasyConfig in the `--missing` check.

I need to think about what the best way is to proceed. This PR could work as a workaround. We could also explicitly install `Z3` without the Python prefix (i.e. new naming), to make the `--missing` check pass (it doesn't really hurt to install that). Not sure what's preferable...

boegel:
I would do the latter (installing `Z3` without Python versionsuffix) in short term.

The idea is to switch to using `--from-commit` when that's ready anyway, and then many of the issues shouldn't present themselves anymore?

ocaisa:
I'd be with @boegel here, the easy (and harmless) fix is to just install the new Z3

trz42:
Obsolete after installing Z3 (#502)

PR #512, opened 2024-03-26
Don't update Lmod cache in software installation script

bedroge:
This is now done on the Stratum 0 to ensure that the Lmod cache gets updated for every software that's being ingested (and to prevent it from being overwritten by an older version of the cache): 
https://github.com/EESSI/filesystem-layer/pull/175


trz42 (approved):
Looks fine to me.

PR #514, opened 2024-03-26
debug failing CI after lmod cache is not updated by bot build jobs

trz42:
Duplicate of #512 to debug failing CI.


trz42:
Problem was apparently caused by an update of Apptainer. See temporary fix in #516 and issue #517.

PR #516, opened 2024-03-26
stick to Apptainer < 1.3.0 for running CI workflows

boegel:


trz42 (approved):
Looks good.

Note, actual CI that failed isn't run here, but a similar change was successful in #514.

PR #507, opened 2024-03-19
{2023.06}[2023a] MODFLOW 6.4.4

casparvl:

PR #518, opened 2024-03-27
Fix rebuild check in `bot/build.sh` script, and always resume from the correct temporary directory

bedroge:
Due to the `set -e`, the grep command that checks for rebuild easystacks would cause the entire script to fail if there is nothing to be rebuilt. Fixed that by adding a true command to the end. 

Also included some other fixes, e.g. one to make sure that the tarball step resumes from the right temp dir (either from the rebuild step or build step, depending on whether a rebuild was done or not).

### DIFF
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -182,7 +182,7 @@ fi
 # determine if the removal step has to be run
 # assume there's only one diff file that corresponds to the PR patch file
 pr_diff=$(ls [0-9]*.diff | head -1)
-changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | grep "/rebuilds/")
+changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | grep "/rebuilds/" || true)
 if [[ -z ${changed_easystacks_rebuilds} ]]; then
     echo "This PR does not add any easystack files in a rebuilds subdirectory, so let's skip the removal step."
 else
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -182,6 +182,8 @@ fi
 # determine if the removal step has to be run
 # assume there's only one diff file that corresponds to the PR patch file
 pr_diff=$(ls [0-9]*.diff | head -1)
+# the true at the end of the next command is important: grep will expectedly return 1 if there is no easystack file being added under rebuilds,
+# but due to "set -e" the entire script would otherwise fail
 changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | (grep "/rebuilds/" || true))
 if [[ -z "${changed_easystacks_rebuilds}" ]]; then
     echo "This PR does not add any easystack files in a rebuilds subdirectory, so let's skip the removal step."
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -192,7 +192,7 @@ else
 
     # prepare arguments to eessi_container.sh specific to remove step
     declare -a REMOVAL_STEP_ARGS=()
-    REMOVAL_STEP_ARGS+=("--save" "${TARBALL_TMP_BUILD_STEP_DIR}")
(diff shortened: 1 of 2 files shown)


bot/build.sh
```diff
@@ -182,8 +182,8 @@ fi
 # determine if the removal step has to be run
 # assume there's only one diff file that corresponds to the PR patch file
 pr_diff=$(ls [0-9]*.diff | head -1)
-changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | grep "/rebuilds/")
-if [[ -z ${changed_easystacks_rebuilds} ]]; then
+changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | grep "/rebuilds/" || true)
```
boegel, line 185: Please add a comment above to explain why the `||` true is there

It's clear now, but we will go wondering later why that's there...

Also, maybe it's worth grouping the latter (mostly to avoid confusion):

```suggestion
changed_easystacks_rebuilds=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | (grep "/rebuilds/" || true))

**Review by boegel — CHANGES_REQUESTED** [review]

File: bot/build.sh [hunk 2]
```diff
@@ -247,7 +247,14 @@ declare -a TARBALL_STEP_ARGS=()
 TARBALL_STEP_ARGS+=("--save" "${TARBALL_TMP_TARBALL_STEP_DIR}")
 
 # determine temporary directory to resume from
-TARBALL_STEP_ARGS+=("--resume" "${REMOVAL_TMPDIR}")
+if [[ -z ${REMOVAL_TMPDIR} ]]; then
+    # no rebuild step was done, so the tarball step should resume from the build directory
+    BUILD_TMPDIR=$(grep ' as tmp directory ' ${build_outerr} | cut -d ' ' -f 2)
+    TARBALL_STEP_ARGS+=("--resume" "${BUILD_TMPDIR}")
```
boegel, line 253: correct me if I'm wrong, but are we sure there is always a tmpdir to resume from?

bot/build.sh
bedroge: Done, and added a comment.

bot/build.sh
bedroge: Yes, there should be? The build step is always run, and even if it builds nothing, it contains for instance the CVMFS files.

trz42: Yes, there is always a tmpdir to resume from ***within*** a job. We also store the tmpdir in a tarball, so we can later use that for inspection. We need to do this because the tmpdir should be cleaned up when the job finishes.

trz42:
PR has been extensively tested in #519 

The only scenario that is currently not supported is when a single PR both rebuilds and adds some package. This can be fixed by the following (tested in https://github.com/EESSI/software-layer/pull/519#issuecomment-2022789235)

```diff
diff --git a/EESSI-install-software.sh b/EESSI-install-software.sh
index 470663e..63b68a5 100755
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -204,11 +204,14 @@ ${EESSI_PREFIX}/scripts/gpu_support/nvidia/install_cuda_host_injections.sh -c 12

 # use PR patch file to determine in which easystack files stuff was added
 changed_easystacks=$(cat ${pr_diff} | grep '^+++' | cut -f2 -d' ' | sed 's@^[a-z]/@@g' | grep '^easystacks/.*yml$' | egrep -v 'known-issues|missing')
-if [ -z ${changed_easystacks} ]; then
+if [ -z "${changed_easystacks}" ]; then
     echo "No missing installations, party time!"  # Ensure the bot report success, as there was nothing to be build here
 else

-    for easystack_file in ${changed_easystacks}; do
+    # first process rebuilds if any, then easystack files for new installations
+    rebuild_easystacks=$(echo "${changed_easystacks}" | grep /rebuilds/ || true)
+    new_easystacks=$(echo "${changed_easystacks}" | grep -v /rebuilds/ || true)
+    for easystack_file in ${rebuild_easystacks} ${new_easystacks}; do

         echo -e "Processing easystack file ${easystack_file}...\n\n"
```

or a similar change in #507

trz42 (approved):
Looks good to me. Tested extensively in #519. Issues discovered therein have been addressed here too.

boegel (approved):

PR #528, opened 2024-04-02
Skip cuda sdk host injections install

### LINKED ISSUE #517: Some CI fails when it installs Apptainer 1.3.0
We've seen this for various PRs (#488, #496, #507, #512, #514)

The last successful run was using Apptainer 1.2.5 (https://github.com/EESSI/software-layer/actions/runs/8284497353).

As a short-time workaround we stick to Apptainer < 1.3.0 (#516).

We should revert that change either when a newer Apptainer works again or when we have found another solution to the issue (e.g., running `apptainer` with additional arguments such as `--underlay`)

casparvl:
Fixes #517


boegel (approved):
lgtm

PR #529, opened 2024-04-03
{2023.06}[2023a] Highway v1.0.4

bedroge:
Trying to fix the test issues for Highway 1.0.4 observed in #428.

PR #532, opened 2024-04-04
Fix Highway hook: change optarch in `pre_prepare_hook` and reset it in `post_prepare_hook`

bedroge:
Since the hook is changing a general build option (for setting `optarch=GENERIC`), it has to be reset afterwards. See https://github.com/EESSI/software-layer/pull/529#pullrequestreview-1976123494.

Tested it interactively for Haswell, Neoverse N1+V1. We could also let the bot do it, but then we'll need to add a rebuild easystack file.

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -323,6 +323,8 @@ def post_prepare_hook_highway_handle_test_compilation_issues(self, *args, **kwar
     if self.name == 'Highway':
         tcname, tcversion = self.toolchain.name, self.toolchain.version
         cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
+        # note: keep condition in sync with the one used in 
+        # pre_prepare_hook_highway_handle_test_compilation_issues
         if self.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':
             if cpu_target == CPU_TARGET_NEOVERSE_N1:
                 update_build_option('optarch', self.orig_optarch)
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -116,6 +116,9 @@ def pre_prepare_hook(self, *args, **kwargs):
         print_msg("Updated rpath_override_dirs (to allow overriding MPI family %s): %s",
                   mpi_family, rpath_override_dirs)
 
+    if self.name in PRE_PREPARE_HOOKS:
+        PRE_PREPARE_HOOKS[self.name](self, *args, **kwargs)
+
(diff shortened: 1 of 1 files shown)


eb_hooks.py
```diff
@@ -290,26 +293,38 @@ def parse_hook_lammps_remove_deps_for_CI_aarch64(ec, *args, **kwargs):
...
       - for neoverse_v1 we completely disable the tests
     cfr. https://github.com/EESSI/software-layer/issues/469
     """
-    if ec.name == 'Highway':
-        tcname, tcversion = ec['toolchain']['name'], ec['toolchain']['version']
+    if self.name == 'Highway':
+        tcname, tcversion = self.toolchain.name, self.toolchain.version
         cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
-        if ec.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':
+        if self.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':
```
boegel, line 307:
```suggestion
        # note: keep condition in sync with the one used in 
        # post_prepare_hook_highway_handle_test_compilation_issues
        if self.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':

File: eb_hooks.py [hunk 2]
```diff
@@ -290,26 +293,38 @@ def parse_hook_lammps_remove_deps_for_CI_aarch64(ec, *args, **kwargs):
...
 
 
+def post_prepare_hook_highway_handle_test_compilation_issues(self, *args, **kwargs):
+    """
+    Post-prepare hook for Highway to reset optarch build option.
+    """
+    if self.name == 'Highway':
+        tcname, tcversion = self.toolchain.name, self.toolchain.version
+        cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
+        if self.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':
```
boegel, line 324:
```suggestion
        # note: keep condition in sync with the one used in 
        # pre_prepare_hook_highway_handle_test_compilation_issues
        if self.version in ['1.0.4'] and tcname == 'GCCcore' and tcversion == '12.3.0':

boegel (changes requested):
looks good, I would just add a comment to help ensure that the condition is kept in sync in the pre/post hooks used for Highway...

boegel (approved):
lgtm

PR #543, opened 2024-04-16
make sure that software directory that corresponds to `$EESSI_SOFTWARE_SUBDIR_OVERRIDE` exists

boegel:
Fix for a small but annoying issue that I bumped into when testing builds for the new `amd/zen4` CPU target.

`/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4` must exist before `init/eessi_environment_variables` is sourced, because there's a check in there whether the subdirectory that `archdetect` produces (which is determined by `$EESSI_SOFTWARE_SUBDIR_OVERRIDE`) corresponds to an existing directory.

Without this, we hit:

```
>> Determining software subdirectory to use for current build host...
>> Picking up pre-defined $EESSI_SOFTWARE_SUBDIR_OVERRIDE: x86_64/amd/zen4
ERROR: no value set for $EESSI_SOFTWARE_SUBDIR
ERROR: Failed to determine software subdirectory?!
```


bedroge (approved):
Lgtm

PR #544, opened 2024-04-16
Remove `build_container.sh` script (replaced by more capable `eessi_container.sh`)

bedroge:
It has been replaced by `eessi_container.sh`.


boegel:
There goes my muscle memory 😅 

(I don't oppose removing the script, to be clear, doesn't make sense to have two scripts that roughly have the same purpose)

boegel (approved):

PR #550, opened 2024-04-21
enable mounting previous overlay-upper read-only

trz42:
Use case for this is that we want to test software additions to a CernVM-FS repository. Typically we would first run `eessi_container.sh` with `--access rw` to make the repository writable with an overlayfs. The changes are stored in some `overlay-upper` directory (under some `/tmp` dir or in a tarball of that `/tmp` directory). When we want to test the additions later we can run `eessi_container.sh` with `--resume TMPDIR_OR_TMPTARBALL`. However, until now this only worked when using `--access rw` which would make the repository again writable, which is likely not what is wanted in a test (because in actual use `/cvmfs/*` mounts are read-only). In fact, the `test-suite.sh` run after a build job has finished, uses `--access ro` (read-only), but this access mode did not recognise changes in an `overlay-upper` directory.

This PR merges an existing `overlay-upper` (given via `--resume ...`) on top of a CernVM-FS repository. Because it only uses the `lowerdir` option of `fuse-overlayfs` the merged directory is mounted read-only.

### DIFF
--- a/eessi_container.sh
+++ b/eessi_container.sh
@@ -594,11 +594,47 @@ declare -a EESSI_FUSE_MOUNTS=()
 # always mount cvmfs-config repo (to get access to software.eessi.io)
 EESSI_FUSE_MOUNTS+=("--fusemount" "container:cvmfs2 cvmfs-config.cern.ch /cvmfs/cvmfs-config.cern.ch")
 
+# check if we got some data via --resume and, if so, use the overlayfs
+# example scenario:
+# 1st step: some software is build in rw mode
+# 2nd step: the software is tested in ro mode (same access as when we would use a
+#           repository)
 if [[ "${ACCESS}" == "ro" ]]; then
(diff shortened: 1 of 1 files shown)


eessi_container.sh
```diff
@@ -594,11 +594,47 @@ declare -a EESSI_FUSE_MOUNTS=()
...
+    # the overlay-upper directory is only created in a read-write-session, thus
+    # we are resuming from such a session here (otherwise there shouldn't be such
+    # directory yet as it is only created for read-write-sessions a bit further
+    # below); the overlay-upper directory can only exist because it is part of
+    # the ${RESUME} directory or tarball
+    # to be able to see the contents of the read-write session we have to mount
+    # the fuse-overlayfs (in read-only mode) on top of the CernVM-FS repository
+
+    # make sure the overlay-upper directory exists
+    mkdir -p ${EESSI_TMPDIR}/overlay-upper
```
ocaisa, line 613: Isn't this redundant, you've checked the directory exists to enter this block

ocaisa (changes requested):

PR #547, opened 2024-04-20
zen4-only {2023.06}[system] EasyBuild v4.9.1

trz42:
Add EasyBuild v4.9.1 to `zen4` software subdir.

Only build for `zen4` on cluster ...

SPDX license identifier: `GPL-2.0-only`

Missing packages:
```
```

PR #564, opened 2024-05-07
also check for missing installations in `x86_64/amd/zen4` subdirectory

boegel:
This will fail until all software installations for `x86_64/amd/zen4` are in place, and we just got started with that in:

* #547


casparvl:
Replaced by https://github.com/EESSI/software-layer/pull/765

PR #571, opened 2024-05-14
Only run OSU test for now in test step. 

casparvl:
As the test suite is growing, the test step would start taking longer and longer. This is particularly annoying if you want to do small tweaks during a build. 

In the (near) future, we should make some mapping to determine which tests to run for which software installations. E.g. if your tarball contains a new TensorFlow module, you probably want to run the TensorFlow test (`-n TensorFlow`). If your tarball contains OpenMPI, you probably want to run `OSU` and maybe one MPI-based application (i.e. `-n OSU -n GROMACS.*foss` for example).

### DIFF
--- a/test_suite.sh
+++ b/test_suite.sh
@@ -175,7 +175,15 @@ else
 fi
 
 # List the tests we want to run
-export REFRAME_ARGS='--tag CI --tag 1_node --nocolor'
+export REFRAME_FLAGS='--nocolor'
+export REFRAME_TAGS='--tag CI --tag 1_node'
+# Limit test names to relevant tests
+# For now, we limit to OSU.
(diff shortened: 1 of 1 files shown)


casparvl:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/skylake_avx512

ocaisa:
Something is going wrong, I see a CUDA and PSM2 as missing installations

ocaisa:
PSM2 is coming from https://github.com/easybuilders/easybuild-easyconfigs/pull/20501

ocaisa:
@casparvl This now requires a sync with the default branch for CI to pass

casparvl:
No longer needed, we now have test selection possibilities in https://github.com/EESSI/software-layer/pull/673

PR #574, opened 2024-05-15
Add `psm2` to filtered dependencies for `2023.06`

ocaisa:
`PSM2` was introduced as a dependency of `libfabric` in https://github.com/easybuilders/easybuild-easyconfigs/pull/20501. 

We already have PSM2 in the compat layer, so we can filter this dependency out, but longer term we probably actually want it since it should be built with accelerator support.


ocaisa:
Urf, I made a meal of this, closing for #575

PR #580, opened 2024-05-16
use directory of called script as base dir for other scripts

trz42:
This improves the usability of `bot/inspect.sh` / fixes a bug. If one ran the script from another directory as the working directory of a bot job (or the main directory of a checkout of EESSI/software-layer), it fails to source other scripts and does not find `./eessi_container.sh`.

### DIFF
--- a/bot/inspect.sh
+++ b/bot/inspect.sh
@@ -27,6 +27,8 @@
 # stop as soon as something fails
 set -e
 
+SCRIPT_DIR=$(dirname $(realpath $BASH_SOURCE))
+
 display_help() {
   echo "usage: $0 [OPTIONS]"
   echo "  -h | --help            -  display this usage information"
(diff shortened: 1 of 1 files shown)


ocaisa (approved):
Changes look good, but I'm not sure how to test this, if I try to submit a build job I guess it will run...

ocaisa (changes requested):
Line 260 in this file should be updated to keep things consistent

trz42:
> Line 260 in this file should be updated to keep things consistent
> 
> https://github.com/trz42/software-layer/blob/use_full_path_in_inspect_script/bot/inspect.sh#L258-L263

Updated the script by using a single variable (and only setting its value once). Also changed variable name from uppercase to lowercase.

trz42:
> Changes look good, but I'm not sure how to test this, if I try to submit a build job I guess it will run...

You can use this to inspect a build job by:
1. launching an interactive job on a node with the same architecture as was used by the build job
2. determine the working directory of the build job you want to inspect, say BUILD_JOB_DIR
3. then run `$BUILD_JOB_DIR/bot/inspect.sh --resume $BUILD_JOB_DIR/previous_tmp/build_step/eessi...tgz`

(I think, it should be possible to create a tool/script that does all three steps, such that you could just run `debug_session.sh --job-id SLURM_JOBID` or similar. Thus it would become even easier to debug failing jobs.)

ocaisa (approved):
LGTM

PR #371, opened 2023-10-20
Easyconfig that can extend EESSI

ocaisa:
The idea would be to add this to (it's own?) easystack file, it will be found as it is in the same directory (as well the "sources" since they are just the hook). It currently requires write permissions to the `host_injections` directory as this allows us to automatically make the modules visible.

The name may be a bit suspect, as I would like to have `EESSI` as a module for different EESSI versions.

PR #582, opened 2024-05-20
Rebuild EESSI-extend module to correctly handle `EESSI_SITE_INSTALL`

trz42:
The current module `EESSI-extend/2023.06-easybuild` does not handle the environment setting `EESSI_SITE_INSTALL` correctly. It just falls back to using the standard `CVMFS` installation path where `versions` is replaced with `host_injections`. This is probably due to an oversight when reviewing PR #371.

Note, when rebuilding this for NESSI, we ran into permission issues when EasyBuild tried to set the permissions of the software after it has been built. Will be interesting to see if we run into the same issue here.


trz42:
bot: build inst:aws repo:eessi.io-2023.06-software arch:zen2

ocaisa:
This is deliberate, we rely on having a known location so we set paths as we do for CUDA

ocaisa (commented):
The original behaviour is deliberate so we can create well defined symlinks for the CUDA installation, but maybe I need to rethink this a bit, I guess we could be a bit more loose given we unbreak those links in a very specific way via scripts

ocaisa:
This path is also used to immediately expose the site modules via our initialisation scripts

trz42:
Should have used EESSI_PROJECT_INSTALL, so this was a misunderstanding on my side. Will close the PR.

PR #558, opened 2024-05-01
use `--from-commit` instead of `--from-pr` for installations done with EasyBuild v4.9.1

boegel:
This doesn't affect the ingested installations at all, but going forward we should use `--from-commit` instead of `--from-pr`, and simply use the last commit in the (merged) PR

Issue #192, opened 2022-10-18
Corner cases that fail for lack of `LD_LIBRARY_PATH`


ocaisa:
Python packages may use ctypes to load libraries:
```
cdll.LoadLibrary("libprotoc.so.3.13.0.0")
```
but this relies on the linker being able to find the libraries, which can fail in our current setup.

mboisson:
This will be relevant to you: https://github.com/ComputeCanada/software-stack/issues/100

mboisson:
Relevant issue in cpython upstream: https://github.com/python/cpython/issues/65241

ocaisa:
@mboisson I'm thinking about making a trivial update to `ctypes` via a sed command that would change https://github.com/python/cpython/blob/main/Lib/ctypes/util.py#L316 to
```
libpath = os.environ.get('LD_LIBRARY_PATH', os.environ.get('LIBRARY_PATH'))
```
that should also solve the issue in our case I think.

boegel:
@ocaisa Why don't we install that as a proper patch for the Python installations we have in EESSI (or even always, I guess it wouldn't do any harm)?

ocaisa:
Yes sure, I was thinking of doing it via `sed ... &&` in `preconfigopts`

ocaisa:
I guess I could actually go ahead and change this in the easyblock. The only scenario I can see this being a problem is where you break the detection by defining `LD_LIBRARY_PATH`

ocaisa:
See https://github.com/easybuilders/easybuild-easyblocks/pull/3352

mboisson:
> @mboisson I'm thinking about making a trivial update to `ctypes` via a sed command that would change https://github.com/python/cpython/blob/main/Lib/ctypes/util.py#L316 to
> 
> ```
> libpath = os.environ.get('LD_LIBRARY_PATH', os.environ.get('LIBRARY_PATH'))
> ```
> 
> that should also solve the issue in our case I think.

that would not fix anything if both LD_LIBRARY_PATH and LIBRARY_PATH are defined.

mboisson:
In my opinion, the correct fix is that `find_library` needs to return the full path. It actually does find the library through `_findLib_gcc`, since it calls the compilers which relies on `LIBRARY_PATH`, but it then returns only the file name of the SO instead of the full path as it should.

ocaisa:
[...]

PR #585, opened 2024-05-23
{2023.06}[2023a] PyTorch-Bundle v2.1.2

casparvl:
```
15 out of 137 required modules missing:

* parameterized/0.9.0-GCCcore-12.3.0 (parameterized-0.9.0-GCCcore-12.3.0.eb)
* tqdm/4.66.1-GCCcore-12.3.0 (tqdm-4.66.1-GCCcore-12.3.0.eb)
* LLVM/14.0.6-GCCcore-12.3.0-llvmlite (LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb)
* Scalene/1.5.26-GCCcore-12.3.0 (Scalene-1.5.26-GCCcore-12.3.0.eb)
* gperftools/2.12-GCCcore-12.3.0 (gperftools-2.12-GCCcore-12.3.0.eb)
* SentencePiece/0.2.0-GCC-12.3.0 (SentencePiece-0.2.0-GCC-12.3.0.eb)
* tensorboard/2.15.1-gfbf-2023a (tensorboard-2.15.1-gfbf-2023a.eb)
* imageio/2.33.1-gfbf-2023a (imageio-2.33.1-gfbf-2023a.eb)
* libmad/0.15.1b-GCCcore-12.3.0 (libmad-0.15.1b-GCCcore-12.3.0.eb)
* SoX/14.4.2-GCCcore-12.3.0 (SoX-14.4.2-GCCcore-12.3.0.eb)
* NLTK/3.8.1-foss-2023a (NLTK-3.8.1-foss-2023a.eb)
* numba/0.58.1-foss-2023a (numba-0.58.1-foss-2023a.eb)
* scikit-image/0.22.0-foss-2023a (scikit-image-0.22.0-foss-2023a.eb)
* librosa/0.10.1-foss-2023a (librosa-0.10.1-foss-2023a.eb)
* PyTorch-bundle/2.1.2-foss-2023a (PyTorch-bundle-2.1.2-foss-2023a.eb)
```

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -4,24 +4,20 @@
 
 name: Scorecards supply-chain security
 on:
-  # For Branch-Protection check. Only the default branch is supported. See
-  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
-  branch_protection_rule:
   # To guarantee Maintained check is occasionally updated. See
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
     - cron: '25 15 * * 3'
   push:
-    branches: [ "main" ]
-  pull_request:
-    branches:    
-      - main
+    branches:
+      - '*-software.eessi.io'
 
 # Declare default permissions as read only.
 permissions: read-all
 
 jobs:
   analysis:
+    if: github.repository_owner == 'EESSI'   # Prevent running on forks
     name: Scorecards analysis
     runs-on: ubuntu-latest
     permissions:
@@ -35,12 +31,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@0864cf19026789058feabb7e87baa5f140aac736 # v2.3.1
+        uses: ossf/scorecard-action@dc50aa9510b46c811795eb24b2f1ba02a914e534 # v2.3.3
         with:
           results_file: results.sarif
           results_format: sarif
@@ -62,7 +58,7 @@ jobs:
       # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
       # format to the repository Actions tab.
       - name: "Upload artifact"
-        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
         with:
           name: SARIF file
           path: results.sarif
@@ -70,6 +66,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@807578363a7869ca324a79039e6db9c843e0e100 # v2.1.27
+        uses: github/codeql-action/upload-sarif@9fdb3e49720b44c48891d036bb502feb25684276 # v3.25.6
         with:
           sarif_file: results.sarif
--- a/.github/workflows/scripts/test_init_scripts.sh
+++ b/.github/workflows/scripts/test_init_scripts.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+EESSI_VERSION="2023.06"
+export LMOD_PAGER=cat
+
+# initialize assert framework
+if [ ! -d assert.sh ]; then
+	echo "assert.sh not cloned."
+	echo ""
+	echo "run \`git clone https://github.com/lehmannro/assert.sh.git\`"
+	exit 1
+fi
+. assert.sh/assert.sh
+
+TEST_SHELLS=("bash" "zsh" "fish" "ksh")
+SHELLS=$@
+
+for shell in ${SHELLS[@]}; do
+	echo = | awk 'NF += (OFS = $_) + 100'
+	echo  RUNNING TESTS FOR SHELL: $shell
+	echo = | awk 'NF += (OFS = $_) + 100'
+  if [[ ! " ${TEST_SHELLS[*]} " =~ [[:space:]]${shell}[[:space:]] ]]; then
+		### EXCEPTION FOR CSH ###
+		echo -e "\033[33mWe don't now how to test the shell '$shell', PRs are Welcome.\033[0m" 
+  else
+		# TEST 1: Source Script and check Module Output
+		assert "$shell -c 'source init/lmod/$shell' 2>&1 " "EESSI/$EESSI_VERSION loaded successfully"
+		# TEST 2: Check if module overviews first section is the loaded EESSI module
+		MODULE_SECTIONS=($($shell -c "source init/lmod/$shell 2>/dev/null; module ov 2>&1 | grep -e '---'"))
+		PATTERN="/cvmfs/software\.eessi\.io/versions/$EESSI_VERSION/software/linux/x86_64/(intel/haswell|amd/zen3)/modules/all"
+		assert_raises 'echo "${MODULE_SECTIONS[1]}" | grep -E "$PATTERN"'
+		# TEST 3: Check if module overviews second section is the EESSI init module
+		assert "echo ${MODULE_SECTIONS[4]}" "/cvmfs/software.eessi.io/versions/$EESSI_VERSION/init/modules"
+		# Test 4: Load Python module and check version
+		command="$shell -c 'source init/lmod/$shell 2>/dev/null; module load Python/3.10.8-GCCcore-12.2.0; python --version'"
+		expected="Python 3.10.8"
+		assert "$command" "$expected"
+		# Test 5: Load Python module and check path
+		PYTHON_PATH=$($shell -c "source init/lmod/$shell 2>/dev/null; module load Python/3.10.8-GCCcore-12.2.0; which python")
+		PATTERN="/cvmfs/software\.eessi\.io/versions/$EESSI_VERSION/software/linux/x86_64/(intel/haswell|amd/zen3)/software/Python/3\.10\.8-GCCcore-12\.2\.0/bin/python"
+		echo "$PYTHON_PATH" | grep -E "$PATTERN"
+		assert_raises 'echo "$PYTHON_PATH" | grep -E "$PATTERN"'
+		
+		#End Test Suite
+		assert_end "source_eessi_$shell"
+	fi
+done
+
+
+# RESET PAGER
+export LMOD_PAGER=
--- a/.github/workflows/test-software.eessi.io.yml
+++ b/.github/workflows/test-software.eessi.io.yml
@@ -1,11 +1,15 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Check for missing software installations in software.eessi.io
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
+  workflow_dispatch:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   check_missing:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test_eessi_container_script.yml
+++ b/.github/workflows/test_eessi_container_script.yml
@@ -1,11 +1,15 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Tests for eessi_container.sh script
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
+  workflow_dispatch:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   eessi_container_script:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
@@ -45,7 +49,8 @@ jobs:
               elif [[ ${{matrix.SCRIPT_TEST}} == 'listrepos_default' ]]; then
                 outfile=out_listrepos.txt
                 ./eessi_container.sh --verbose --list-repos | tee ${outfile}
-                grep "EESSI" ${outfile}
+                # make sure that the default EESSI software repository is available
+                grep "software.eessi.io" ${outfile}
 
               # test use of --list-repos with custom repos.cfg
               elif [[ ${{matrix.SCRIPT_TEST}} == 'listrepos_custom' ]]; then
@@ -57,11 +62,12 @@ jobs:
                 echo "[EESSI/20HT.TP]" >> cfg/repos.cfg
                 echo "repo_version = 20HT.TP" >> cfg/repos.cfg
                 ./eessi_container.sh --verbose --list-repos | tee ${outfile}
-                grep "EESSI" ${outfile}
+                # make sure that the default EESSI software repository is available
+                grep "software.eessi.io" ${outfile}
 
                 export EESSI_REPOS_CFG_DIR_OVERRIDE=${PWD}/cfg
                 ./eessi_container.sh --verbose --list-repos | tee ${outfile2}
-                grep "[EESSI/2023.02]" ${outfile2}
+                grep "EESSI/20AB.CD" ${outfile2}
 
               # test use of --mode run
               elif [[ ${{matrix.SCRIPT_TEST}} == 'run' ]]; then
--- a/.github/workflows/test_licenses.yml
+++ b/.github/workflows/test_licenses.yml
@@ -1,11 +1,14 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Test software licenses
-on: [push, pull_request]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
         - name: Check out software-layer repository
           uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,14 +1,17 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Tests
-on: [push, pull_request]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9, '3.10']
+        python: [3.7, 3.8, 3.9, '3.10']
       fail-fast: false
     steps:
     - name: checkout
--- a/.github/workflows/tests_archdetect.yml
+++ b/.github/workflows/tests_archdetect.yml
@@ -1,11 +1,14 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Tests for eessi_archdetect.sh
-on: [push, pull_request]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         proc_cpuinfo:
--- a/.github/workflows/tests_eessi_module.yml
+++ b/.github/workflows/tests_eessi_module.yml
@@ -1,0 +1,85 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+name: Tests for eessi_module_functionality in software.eessi.io
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
+permissions:
+  contents: read # to fetch code (actions/checkout)
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        EESSI_VERSION:
+          - 2023.06        
+    steps:
+      - name: Check out software-layer repository
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+          
+      - name: Mount EESSI CernVM-FS pilot repository
+        uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
+        with:
+          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+          cvmfs_http_proxy: DIRECT
+          cvmfs_repositories: software.eessi.io
+
+      - name: Test for making sure spider cache is being used and not being rebuilt
+        run: |
+          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/compat/linux/$(uname -m)/usr/share/Lmod/init/bash  # Initialise Lmod
+          export MODULEPATH=init/modules
+          configfile="configfile.txt"
+          module -T load EESSI/${{matrix.EESSI_VERSION}}
+          module --config > "${configfile}" 2>&1
+          grep cache "${configfile}" | grep software | grep -v compat
+          if timeout 10s bash -c "LMOD_PAGER=none module --terse avail" && grep cache "${configfile}" | grep software | grep -v compat; then
+              echo "EESSI spider cache is being used"
+          else
+              echo "EESSI spider cache is being rebuilt" >&2
+              exit 1
+          fi
+          env | grep LMOD
+          module purge
+          unset MODULEPATH
+    
+      - name: Test for archdetect_cpu functionality with invalid path
+        run: |
+          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/compat/linux/$(uname -m)/usr/share/Lmod/init/bash  # Initialise Lmod
+          export MODULEPATH=init/modules
+          set +e  # Do not exit immediately if a command exits with a non-zero status          
+          export EESSI_ARCHDETECT_OPTIONS="dummy/cpu"
+          outfile="outfile.txt"
+          module load EESSI/${{matrix.EESSI_VERSION}} > "${outfile}" 2>&1
+          cat "${outfile}"
+          if grep -q  "Software directory check" "${outfile}"; then
+              echo "Test for picking up invalid path on \${archdetect_cpu} PASSED"
+          else
+              echo "Test for picking up invalid path on \${archdetect_cpu} FAILED" >&2
+              exit 1
+          fi
+          unset EESSI_ARCHDETECT_OPTIONS
+          set -e  # Re-enable exit on non-zero status
+              
+      - name: Test for expected variables while adding dummy cpu archs and loading EESSI module 
+        run: |
+          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/compat/linux/$(uname -m)/usr/share/Lmod/init/bash  # Initialise Lmod
+          export MODULEPATH=init/modules
+          CPU_ARCH=$(./init/eessi_archdetect.sh -a cpupath)
+          export EESSI_ARCHDETECT_OPTIONS="dummy/cpu:${CPU_ARCH}:dummy1/cpu1"          
+          moduleoutfile="moduleout.txt"
+          sourceoutfile="sourceout.txt"
+          module load EESSI/${{matrix.EESSI_VERSION}}
+          env | grep -E '^(EESSI_S|EESSI_C)' | sort > "${moduleoutfile}"
+          module unload EESSI/${{matrix.EESSI_VERSION}}
+          source /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/bash
+          env | grep -E '^(EESSI_S|EESSI_C)' | sort > "${sourceoutfile}"
+          cat "${moduleoutfile}"
+          cat "${sourceoutfile}"
+          if (diff "${moduleoutfile}" "${sourceoutfile}" > /dev/null); then
+              echo "Test for checking env variables PASSED"
+          else 
+              echo "Test for checking env variables FAILED"  >&2
+              exit 1
+          fi
+
--- a/.github/workflows/tests_init.yml
+++ b/.github/workflows/tests_init.yml
@@ -1,14 +1,17 @@
 # documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Tests for init scripts
-on: [push, pull_request]
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8, 3.9, '3.10']
+        python: [3.7, 3.8, 3.9, '3.10']
       fail-fast: false
     steps:
     - name: checkout
--- a/.github/workflows/tests_init_module.yml
+++ b/.github/workflows/tests_init_module.yml
@@ -1,0 +1,43 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+name: Check for EESSI init shell scripts to load eessi software module in software.eessi.io
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
+  workflow_dispatch:
+permissions:
+  contents: read # to fetch code (actions/checkout)
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        EESSI_VERSION:
+          - 2023.06
+        EESSI_SOFTWARE_SUBDIR_OVERRIDE:
+          - x86_64/intel/haswell
+    steps:
+      - name: Check out software-layer repository
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Mount EESSI CernVM-FS pilot repository
+        uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
+        with:
+          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+          cvmfs_http_proxy: DIRECT
+          cvmfs_repositories: software.eessi.io
+         
+      - name: Clone assert.sh script
+        run: git clone https://github.com/lehmannro/assert.sh.git
+
+      - name: Install missing shells
+        run: |
+          sudo apt update
+          sudo apt install zsh ksh fish
+          echo "# INIT ZSH" > ~/.zshrc
+
+      - name: Run tests for available shells
+        run: |
+          .github/workflows/scripts/test_init_scripts.sh "bash" "zsh" "ksh" "fish" "csh"
+
--- a/.github/workflows/tests_readme.yml
+++ b/.github/workflows/tests_readme.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
         - name: Check out software-layer repository
           uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -28,7 +28,7 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - name: checkout
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
--- a/EESSI-extend-2023.06-easybuild.eb
+++ b/EESSI-extend-2023.06-easybuild.eb
@@ -139,12 +139,17 @@ setenv ("EASYBUILD_SYSROOT", sysroot)
 setenv ("EASYBUILD_PREFIX", pathJoin(working_dir, "easybuild"))
 setenv ("EASYBUILD_INSTALLPATH", easybuild_installpath)
 setenv ("EASYBUILD_HOOKS", pathJoin(os.getenv("EESSI_PREFIX"), "init", "easybuild", "eb_hooks.py"))
-setenv ("EASYBUILD_UMASK", "002")
+
+-- Make sure to use the general umask that allows a global read
+setenv ("EASYBUILD_UMASK", "022")
+
+-- Allow this module to be loaded when running EasyBuild
+setenv ("EASYBUILD_ALLOW_LOADED_MODULES", "EasyBuild,EESSI-extend")
 
 -- Set all related environment variables if we have project or user installations (including extending MODULEPATH)
 if (user_modulepath ~= nil) then
   -- Use a more restrictive umask for this case
-  setenv ("EASYBUILD_UMASK", "022")
+  setenv ("EASYBUILD_UMASK", "077")
   setenv ("EASYBUILD_STICKY_BIT", "1")
   -- configure MODULEPATH
   if (project_modulepath ~= nil) then
@@ -155,6 +160,7 @@ elseif (project_modulepath ~= nil) then
   setenv ("EASYBUILD_SET_GID_BIT", "1")
   setenv ("EASYBUILD_GROUP_WRITABLE_INSTALLDIR", "1")
   setenv ("EASYBUILD_STICKY_BIT", "0")
+  setenv ("EASYBUILD_UMASK", "002")
   -- configure MODULEPATH
   prepend_path("MODULEPATH", project_modulepath)
 end
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -17,6 +17,11 @@ display_help() {
   echo "  --skip-cuda-install    -  disable installing a full CUDA SDK in the host_injections prefix (e.g. in CI)"
 }
 
+# Function to check if a command exists
+function command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
+
 function copy_build_log() {
     # copy specified build log to specified directory, with some context added
     build_log=${1}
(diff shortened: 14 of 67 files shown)


casparvl:
```
== No easyconfigs left to be built.
ERROR: Missing dependencies: SentencePiece/0.2.0-foss-2023a, SoX/14.4.2-foss-2023a (no easyconfig file or existing module found)
== Build succeeded for 0 out of 0
  >> download succeeded: https://github.com/easybuilders/easybuild-easyconfigs/archive/7124863ed588066e5a988b4073d91381497a7f64.tar.gz
  >> running command:
        [started at: 2024-05-23 09:28:34]
        [working dir: /tmp/eb-dlj1ws2x/eb-9tn8fu3_/tmpp3me5uio/easybuilders]
        [output logged in /tmp/eb-dlj1ws2x/eb-9tn8fu3_/easybuild-run_cmd-t6inmlw4.log]
        tar xzf /tmp/eb-dlj1ws2x/eb-9tn8fu3_/tmpp3me5uio/easybuilders/7124863ed588066e5a988b4073d91381497a7f64.tar.gz
  >> command completed: exit 0, ran in 00h00m01s
== found valid index for /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/EasyBuild/4.9.1/easybuild/easyconfigs, so using it...
== Running parse hook for PyTorch-bundle-2.1.2-foss-2023a.eb...
== Running parse hook for foss-2023a.eb...
== resolving dependencies ...
== Running parse hook for parameterized-0.9.0-GCCcore-12.3.0.eb...
== Running parse hook for GCCcore-12.3.0.eb...
== Running parse hook for GCCcore-12.3.0.eb...
== Running parse hook for scikit-image-0.22.0-foss-2023a.eb...
== Running parse hook for librosa-0.10.1-foss-2023a.eb...
== Running parse hook for imageio-2.33.1-gfbf-2023a.eb...
== Running parse hook for gfbf-2023a.eb...
== Running parse hook for gfbf-2023a.eb...
== Running parse hook for GCC-12.3.0.eb...
== Running parse hook for FlexiBLAS-3.3.1-GCC-12.3.0.eb...
== Running parse hook for GCC-12.3.0.eb...
== Running parse hook for FFTW-3.3.10-GCC-12.3.0.eb...
== Running parse hook for NLTK-3.8.1-foss-2023a.eb...
== Running parse hook for numba-0.58.1-foss-2023a.eb...
== Running parse hook for Scalene-1.5.26-GCCcore-12.3.0.eb...
== Running parse hook for tqdm-4.66.1-GCCcore-12.3.0.eb...
== Running parse hook for LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb...
== Running parse hook for tensorboard-2.15.1-gfbf-2023a.eb...
```
I guess that with `--from-pr` we got `SentencePiece` and `Sox` correctly since they were already in develop, but with `--from-commit` we don't? Should I combine multiple `--from-commit`'s for each of those (i.e. look up the commit that provided the required `SentencePiece`, etc)?

bedroge:
> ```
> == No easyconfigs left to be built.
> ERROR: Missing dependencies: SentencePiece/0.2.0-foss-2023a, SoX/14.4.2-foss-2023a (no easyconfig file or existing module found)
> == Build succeeded for 0 out of 0
>   >> download succeeded: https://github.com/easybuilders/easybuild-easyconfigs/archive/7124863ed588066e5a988b4073d91381497a7f64.tar.gz
>   >> running command:
>         [started at: 2024-05-23 09:28:34]
>         [working dir: /tmp/eb-dlj1ws2x/eb-9tn8fu3_/tmpp3me5uio/easybuilders]
>         [output logged in /tmp/eb-dlj1ws2x/eb-9tn8fu3_/easybuild-run_cmd-t6inmlw4.log]
>         tar xzf /tmp/eb-dlj1ws2x/eb-9tn8fu3_/tmpp3me5uio/easybuilders/7124863ed588066e5a988b4073d91381497a7f64.tar.gz
>   >> command completed: exit 0, ran in 00h00m01s
> == found valid index for /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/EasyBuild/4.9.1/easybuild/easyconfigs, so using it...
> == Running parse hook for PyTorch-bundle-2.1.2-foss-2023a.eb...
> == Running parse hook for foss-2023a.eb...
> == resolving dependencies ...
> == Running parse hook for parameterized-0.9.0-GCCcore-12.3.0.eb...
> == Running parse hook for GCCcore-12.3.0.eb...
> == Running parse hook for GCCcore-12.3.0.eb...
> == Running parse hook for scikit-image-0.22.0-foss-2023a.eb...
> == Running parse hook for librosa-0.10.1-foss-2023a.eb...
> == Running parse hook for imageio-2.33.1-gfbf-2023a.eb...
> == Running parse hook for gfbf-2023a.eb...
> == Running parse hook for gfbf-2023a.eb...
> == Running parse hook for GCC-12.3.0.eb...
> == Running parse hook for FlexiBLAS-3.3.1-GCC-12.3.0.eb...
> == Running parse hook for GCC-12.3.0.eb...
> == Running parse hook for FFTW-3.3.10-GCC-12.3.0.eb...
> == Running parse hook for NLTK-3.8.1-foss-2023a.eb...
> == Running parse hook for numba-0.58.1-foss-2023a.eb...
> == Running parse hook for Scalene-1.5.26-GCCcore-12.3.0.eb...
> == Running parse hook for tqdm-4.66.1-GCCcore-12.3.0.eb...
> == Running parse hook for LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb...
> == Running parse hook for tensorboard-2.15.1-gfbf-2023a.eb...
> ```
> 
> I guess that with `--from-pr` we got `SentencePiece` and `Sox` correctly since they were already in develop, but with `--from-commit` we don't? Should I combine multiple `--from-commit`'s for each of those (i.e. look up the commit that provided the required `SentencePiece`, etc)?

I (and @trz and @ocaisa ) also saw issues with using `--from-commit`, see for instance https://github.com/EESSI/software-layer/pull/558#issuecomment-2090836084.

bedroge:
Could you try using the merge commit (see bottom of the PR: 04ccd901a613631b00ccbe504d6d66d6a6c2febb) and check if that does work?

casparvl:
I tried manually
```
eb -D PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb --from-commit 04ccd901a613631b00ccbe504d6d66d6a6c2febb
```
But that still shows missing EasyConfigs.

bedroge:
> I tried manually
> 
> ```
> eb -D PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb --from-commit 04ccd901a613631b00ccbe504d6d66d6a6c2febb
> ```
> 
> But that still shows missing EasyConfigs.

Guess we need to stick to `--from-pr` then until we find a solution for this...

casparvl:

casparvl:
This is the actual failure:
```
== 2024-05-23 12:17:16,011 build_log.py:171 ERROR EasyBuild crashed with an error (at easybuild/tools/build_log.py:111 in caller_info): Sanity check failed: extensions sanity check failed for 1 extensions: soundfile
failing sanity check for 'soundfile' extension: command "python -c "import soundfile"" failed; output:
Traceback (most recent call last):
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/librosa/0.10.1-foss-2023a/lib/python3.11/site-packages/soundfile.py", line 161, in <module>
    import _soundfile_data  # ImportError if this doesn't exist
    ^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named '_soundfile_data'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/librosa/0.10.1-foss-2023a/lib/python3.11/site-packages/soundfile.py", line 171, in <module>
    _snd = _ffi.dlopen(_libname)
           ^^^^^^^^^^^^^^^^^^^^^
OSError: cannot load library 'libsndfile.so.1': libsndfile.so.1: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/librosa/0.10.1-foss-2023a/lib/python3.11/site-packages/soundfile.py", line 192, in <module>
    _snd = _ffi.dlopen(_explicit_libname)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: cannot load library 'libsndfile.so': libsndfile.so: cannot open shared object file: No such file or directory,  (at easybuild/framework/easyblock.py:3669 in _sanity_check_step)
```
I guess this should be provide by the module `libsndfile/1.2.2-GCCcore-12.3.0`, but I'm not sure what path's get searched by this `dlopen` call. I _think_ that searches `LD_LIBRARY_PATH`, which we don't set in EESSI.

I guess this is a pretty fundamental question: how do we make `dlopen` calls succesfully find libs from the EESSI software prefix?

ocaisa:
See https://github.com/EESSI/software-layer/issues/192 , the Alliance have a solution for this

casparvl:
Spot on, it is indeed the issue of `ctypes.util`'s `find_library` only returning the filename, not the full path. Or at least: I see that it is using `find_library` [here](https://github.com/bastibe/python-soundfile/blob/8ebb523725e315f24c5592677353c43c4562be54/soundfile.py#L168) to ge tthe `_libname`, which is then used as the `dlopen` argument. I.e. I expect that if `find_library` correctly returns the full path, the `dlopen` call would have succeeded.

The downside is that the Alliance's solution looks quite involved... The upside is we can probably use their shadowing lib from https://github.com/ComputeCanada/custom_ctypes/tree/main/lib . What I don't fully understand is the `sitecustomize` and `ebpythonprefixes` stuff they do. Also, they seem to make a seperate module out of it, I'm not entirely sure why (do they only load it when they need to?).

I guess my main consideration would be if we shouldn't just always have this patched `find_library` function in place. In that case, a simple patch to the installation that normally contains `ctypes` (I guess that's in the standard Python installation?) would then be enough...

ocaisa:
I was also thinking that maybe a patch on ctypes is enough, I don't fully understand all the other stuff going on with them

trz42:
The changes they apply to `ctypes` are quite small. See below for Python/3.11.3 Maybe we could apply these changes "in-place" in a build container to test if they solve the issue?
```diff
diff -u /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/skylake_avx512/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/ctypes/util.py custom_ctypes/lib/python3.11/site-packages/ctypes/util.py
--- /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/skylake_avx512/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/ctypes/util.py      2024-04-30 16:38:09.000000000 +0200
+++ custom_ctypes/lib/python3.11/site-packages/ctypes/util.py   2024-05-30 16:17:44.000000000 +0200
@@ -326,7 +326,10 @@

         def find_library(name):
             # See issue #9998
+            lib = _findLib_gcc(name)
+            # return absolute path
             return _findSoname_ldconfig(name) or \
+                    os.path.join(os.path.dirname(lib), _get_soname(lib)) or \
                    _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))

 ################################################################
```

trz42:
I tried to replace the `utils.py` globally (for all installations in https://github.com/NorESSI/software-layer/pull/387), but that leads to a failure when building/installing `scikitimage` already (third package). See below for details. When I don't use the modified `utils.py` it fails with the same error @casparvl has hit when building `librosa`.

```
    File "/cvmfs/pilot.nessi.no/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/ctypes/util.py", line 332, in find_library
      os.path.join(os.path.dirname(lib), _get_soname(lib)) or \
                   ^^^^^^^^^^^^^^^^^^^^
    File "<frozen posixpath>", line 152, in dirname
  TypeError: expected str, bytes or os.PathLike object, not NoneType
  error: subprocess-exited-with-error
```

Will try to use that modified file only when building/using `librosa`.

trz42:
I've worked out a fix for the `import soundfile` issue. See https://github.com/NorESSI/software-layer/pull/391

If it works out there, I'll test it with PyTorch-bundle. We can dicuss how we should employ this fix (maybe it's better to ship the custom `ctypes` with EESSI, but for lack of better idea where to put it the above PR puts it under `host_injections`).

trz42:
I updated https://github.com/NorESSI/software-layer/pull/387 with the fixes in https://github.com/NorESSI/software-layer/pull/391 to work around the failing sanity check (`python -c 'import soundfile'`). PyTorch (with CUDA) builds for `x86_64/{generic,intel/skylake_avx512,amd/zen2}`. It fails for `aarch64/generic` and `x86_64/intel/broadwell` with a different issue. It could be worth applying the fixes also here and see which builds work (and which don't).

casparvl:
@trz42 I remember you said in a meeting that simply patching `ctypes` caused issues in other packages. I think the idea was then to pick up a 'patched' `ctypes` only for a specific phase of the build (the test step? I don't fully remember...). However, it was also brought up in that meeting that this fix would make the _build_ pass, but users would still run into it at runtime, right?

I was thinking: what if we patch `ctypes` to _add_ a different API call. I.e. a `find_library` with an extra argument `full_path` (which defaults to `false`, i.e. the default behaviour). And then, we patch `librosa` to call `find_library(..., full_path=true)`. That way, you _only_ get the full path back if you _intentionaly_ patch an application that depends on this `find_library` call. That should have no unintended fallout (because the default function call retains it's prior behaviour of only returning the library name, not the full library path), while giving us an easy way to fix future similar issues (simply patch the function calls to `find_library` to add the `full_path=true` argument).

ocaisa:

casparvl:
```
=========================== short test summary info ============================
FAILED test/test_image.py::test_decode_jpeg[None-ImageReadMode.UNCHANGED-grace_hopper_517x606.jpg]
FAILED test/test_image.py::test_decode_jpeg[None-ImageReadMode.UNCHANGED-cmyk_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[None-ImageReadMode.UNCHANGED-gray_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[None-ImageReadMode.UNCHANGED-rgb_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[L-ImageReadMode.GRAY-grace_hopper_517x606.jpg]
FAILED test/test_image.py::test_decode_jpeg[L-ImageReadMode.GRAY-cmyk_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[L-ImageReadMode.GRAY-gray_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[L-ImageReadMode.GRAY-rgb_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[RGB-ImageReadMode.RGB-grace_hopper_517x606.jpg]
FAILED test/test_image.py::test_decode_jpeg[RGB-ImageReadMode.RGB-cmyk_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[RGB-ImageReadMode.RGB-gray_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg[RGB-ImageReadMode.RGB-rgb_pytorch.jpg]
FAILED test/test_image.py::test_decode_jpeg_errors - AssertionError: Regex pa...
FAILED test/test_image.py::test_decode_bad_huffman_images - RuntimeError: dec...
FAILED test/test_image.py::test_damaged_corrupt_images[corrupt.jpg] - Asserti...
FAILED test/test_image.py::test_damaged_corrupt_images[corrupt34_2.jpg] - Ass...
FAILED test/test_image.py::test_damaged_corrupt_images[corrupt34_3.jpg] - Ass...
FAILED test/test_image.py::test_damaged_corrupt_images[corrupt34_4.jpg] - Ass...
FAILED test/test_image.py::test_encode_jpeg_errors - AssertionError: Regex pa...
FAILED test/test_image.py::test_encode_jpeg[grace_hopper_517x606.jpg] - Runti...
FAILED test/test_image.py::test_write_jpeg[grace_hopper_517x606.jpg] - Runtim...
= 21 failed, 48811 passed, 50354 skipped, 2503 deselected, 2220 warnings in 965.82s (0:16:05) =
```

trz42:
@casparvl  the torchvision issue should be fixed with an updated easyblock. Either use that or use EB v4.9.2 which should come with the fix. See #603

casparvl:

casparvl:
bot: build arch:aarch64/generic repo:eessi.io-2023.06-software

casparvl:
bot: build arch:aarch64/generic repo:eessi.io-2023.06-software

casparvl:
Yep, reproduced @trz42 's issue:

```
== 2024-08-13 10:17:49,272 easyblock.py:3632 WARNING Sanity check: sanity check command python -c 'import sentencepiece' exited with code 1 (output: Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/SentencePiece/0.2.0-GCC-12.3.0/lib/python3.11/site-packages/sentencepiece/__init__.py", line 10, in <module>
    from . import _sentencepiece
ImportError: /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so.4: cannot allocate memory in static TLS block
)
```

casparvl:
Related: https://github.com/gluster/glusterfs/issues/2913 and probably why @trz42 disabled it in his patch :)

But, maybe we should  dig into a solution without disabling it, as now we get `segmentation fault`s in the tests, which might be related to disabling the use of `tcmalloc`...

casparvl:
I think [this](https://github.com/opencv/opencv/issues/14884#issuecomment-1146880241) might put us on track for a solution: maybe we should patch `sentencepiece` to load `libtcmalloc_minimal.so.4` earlier on. I've asked ChatGPT for help:

=======
I get an error from libtcmalloc: 'cannot allocate memory in static TLS block'. Can you help me understand this error? What is static TLS?

The error message `cannot allocate memory in static TLS block` typically occurs when there's an issue with memory allocation related to Thread-Local Storage (TLS) in a program. Let's break down the concepts involved to understand this error better.

### Understanding Thread-Local Storage (TLS)

1. **Thread-Local Storage (TLS)**:
   - TLS is a mechanism that allows threads to have their own copy of variables. Each thread maintains its own separate storage, enabling data to be thread-specific without interference from other threads.
   - This is particularly useful for variables that maintain state across function calls in a thread-safe manner without using global or static variables.

2. **Static TLS vs. Dynamic TLS**:
   - **Static TLS**: This is allocated at program startup. It is used for variables that have been marked with the `__thread` or `thread_local` specifier in languages like C and C++. The size of the static TLS block is determined at compile time and allocated in the program’s binary.
   - **Dynamic TLS**: Allocated at runtime, allowing for flexibility when the exact amount of TLS needed isn't known until the program is running.

### The Error Explained

The error `cannot allocate memory in static TLS block` usually occurs in the following scenarios:

1. **Exceeding Static TLS Limit**:
   - Every process in a Linux system has a fixed size for its static TLS block. If you declare too many thread-local variables or if the size of these variables is too large, it might exceed the allocated static TLS block size. This results in the inability to allocate more thread-local storage statically.

2. **Library Compatibility Issues**:
   - Sometimes, the error may be due to compatibility issues between different libraries and the way they handle TLS. For example, if a library like `libtcmalloc` (a high-performance memory allocator) is not fully compatible with how another library uses TLS, this can cause the error.

3. **Platform Limitations**:
   - Some platforms have more restrictive limits on TLS size. For example, some older versions of glibc (GNU C Library) had limitations that affected how much static TLS could be allocated.

### Solutions

1. **Reduce TLS Usage**:
   - Review your code to ensure you’re not unnecessarily using TLS. Try to reduce the number of thread-local variables or their size.
   
2. **Switch to Dynamic TLS**:
   - If possible, switch to dynamic TLS allocation for your variables. This can be achieved by changing compiler flags or the way you declare thread-local storage.

3. **Upgrade or Reconfigure Libraries**:
   - Check if there are updates or patches available for the libraries you’re using that might fix the compatibility issues. Sometimes, newer versions of libraries have better handling of TLS.

4. **Increase Static TLS Size**:
   - Depending on your system and libraries, it might be possible to increase the size of the static TLS block through compiler or linker options.

5. **Use an Alternative Memory Allocator**:
   - If the issue is related to `libtcmalloc`, consider switching to a different memory allocator that is compatible with your setup.

6. **Check Compiler and Linker Options**:
   - Ensure that your compiler and linker options are set up to handle the necessary TLS requirements for your application.

By understanding the underlying mechanisms of TLS and its limitations, you can diagnose and address the `cannot allocate memory in static TLS block` error more effectively.
=======

All in all, not a bad answer. 

Another nice source: [this](https://github.com/opencv/opencv/issues/14884#issuecomment-815632861) mentions the root cause. With more information [here](https://bugzilla.redhat.com/show_bug.cgi?id=1722181)

casparvl:
This seems to discuss an even more fundamental 'fix' at the `glibc`  level: https://bugzilla.redhat.com/show_bug.cgi?id=1871396 . I'm wondering if any of that ever made it in, and if e.g. a newer compat layer (with newer `glibc`) would not even have this issue anymore...

casparvl:
More resources: https://www.akkadia.org/drepper/tls.pdf and https://gcc.gnu.org/onlinedocs/gcc-3.4.6/gcc/Thread_002dLocal.html#Thread_002dLocal

casparvl:
bot: build arch:aarch64/generic repo:eessi.io-2023.06-software

casparvl:

casparvl:
Hm, https://github.com/EESSI/software-layer/pull/585#issuecomment-2302912876 still fails with the same static TLS issue. I realize why though: the sanity check is run _before_ generating the module file. It will generate a temporary module file for this step, but since the hook only gets applied when generating the _final_ module file, it doesn't get applied here!

Fix should be relatively simple: make the hook apply earlier, as a `pre_sanitycheck_hook`.

casparvl:
bot: build arch:aarch64/generic repo:eessi.io-2023.06-software

trz42:
Copying some findings from Slack here:

To me it seems the problem is a combination of what EasyBuild uses to run commands (it uses `/bin/bash`) and that we currently set `LD_PRELOAD` too early via the modified module file. Below are a few examples illustrating what happens.

The original TLS (Thread-Local Storage) allocation error... (withou `LD_PRELOAD`, just running the import after loading Python, gperftools and setting `PATH` and `PYTHONPATH` to the build directory for SentencePiece)
```bash
bot@aarch64-generic-node3 /tmp/bot $ python -c 'import sentencepiece'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/SentencePiece/0.2.0-GCC-12.3.0/lib/python3.11/site-packages/sentencepiece/__init__.py", line 10, in <module>
    from . import _sentencepiece
ImportError: /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so.4: cannot allocate memory in static TLS block
```

With `LD_PRELOAD` this succeeds (same env otherwise)...
```bash
bot@aarch64-generic-node3 /tmp/bot $ LD_PRELOAD=/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so python -c 'import sentencepiece'
```

However, that is not how EasyBuild runs the sanitycheck command. It rather runs the following (which fails)...
```bash
bot@aarch64-generic-node3 /tmp/bot $ LD_PRELOAD=/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so /bin/bash -c "python -c 'import sentencepiece'"
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.35' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libgcc_s.so.1)
/bin/bash: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/GCCcore/12.3.0/lib64/libgcc_s.so.1)
```

The above error is what we got in the last build job for `aarch64/generic`. If we run the original command in a subshell (as EasyBuild does), we get the original error (just to illustrate that we "correctly" emulate what EasyBuild does)...
```bash
bot@aarch64-generic-node3 /tmp/bot $ /bin/bash -c "python -c 'import sentencepiece'"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/SentencePiece/0.2.0-GCC-12.3.0/lib/python3.11/site-packages/sentencepiece/__init__.py", line 10, in <module>
    from . import _sentencepiece
ImportError: /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so.4: cannot allocate memory in static TLS block
```

If we set `LD_PRELOAD` _just_ before we run python, it works...
```bash
bot@aarch64-generic-node3 /tmp/bot $ /bin/bash -c "LD_PRELOAD=/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so python -c 'import sentencepiece'"
```

I think, setting `LD_PRELOAD` in the module for SentencePiece could work. However, when running EasyBuild we'll likely run into issues because it uses `/bin/bash` to run commands. If it would use bash from the compat layer it would work. See example below

```bash
bot@aarch64-generic-node3 /tmp/bot $ LD_PRELOAD=/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so.4 /cvmfs/software.eessi.io/versions/2023.06/compat/linux/aarch64/bin/bash -c "python -c 'import sentencepiece'"
```

To me it seems that `/bin/bash` and `/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so.4` depend on different symbols (which sounds logical), hence it is critical to only preload the latter library after `/bin/bash`'s dependencies have been resolved.

boegel:
@trz42 Doesn't this mean that EasyBuild should be using the `/bin/bash` from the compat layer, so prefixed with `sysroot` in EasyBuild lingo?

trz42:
> @trz42 Doesn't this mean that EasyBuild should be using the `/bin/bash` from the compat layer, so prefixed with `sysroot` in EasyBuild lingo?

Maybe. If `sysroot` implies that it can expect a `sysroot/bin/bash` it could work. However, it has only resulted in a problem when we use `LD_PRELOAD`. So, maybe we should look for another solution.

I'm trying to solve the issue with a parse hook where I just add `LD_PRELOAD=...` in front of the failing sanity check command and another hook to add `LD_PRELOAD=...` in the module file. However, the latter has to be done after the sanity check has been run.

A better fix could be what you suggest, in some cases or always, we prefix the `exec_cmd = "/bin/bash"` (/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/EasyBuild/4.9.2/lib/python3.11/site-packages/easybuild/tools/run.py:229) with `sysroot` when it is present. Then we could just add the setting of `LD_PRELOAD=...` in the module file and it should work both while using the module and while running the sanity check.

casparvl:
> @trz42 Doesn't this mean that EasyBuild should be using the /bin/bash from the compat layer, so prefixed with sysroot in EasyBuild lingo?

To me, this makes a lot of sense actually. If you're explicitly invoking a shell to run your command, and if a `sysroot` is set, it should be the shell from that `sysroot` prefix imho. 

What _is_ the reason that EasyBuild is running this in a subshell actually? I mean that is not typically how I would test the module _manually_ and could _potentially_ lead to differences with running it in the parent shell (this example begin a case in point).

boegel:
@casparvl All shell commands run by EasyBuild are run in a subshell...

boegel:
> A better fix could be what you suggest, in some cases or always, we prefix the `exec_cmd = "/bin/bash"` (/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/generic/software/EasyBuild/4.9.2/lib/python3.11/site-packages/easybuild/tools/run.py:229) with `sysroot` when it is present. Then we could just add the setting of `LD_PRELOAD=...` in the module file and it should work both while using the module and while running the sanity check.

I think that's the right way forward...

It's a relatively easy change to make in EasyBuild (though in some sense a breaking one, so perhaps we need to make it configurable).

trz42:
We may even test this change already by copying the `bash` files from the two compat layers (`x86_64` and `aarch64`) to some directory in the PR and then modify the launch of the containers such that the right file is bind mounted to `/bin/bash` inside the container. Before we run `eessi_container.sh` we can set `SINGULARITY_BIND`.

casparvl:

casparvl:
Ok, good that the test now works on `x86_64`.

For [this issue](https://github.com/EESSI/software-layer/pull/585#issuecomment-2313143191) on ARM, I made a fix https://github.com/easybuilders/easybuild-framework/pull/4646 for EasyBuild framework, only to realize afterwards that the whole `run_cmd` thing is completely overhauled in EasyBuild 5.0. Looking at the 5.0.X code [here](https://github.com/easybuilders/easybuild-framework/blob/a2550eb8fab479f517badbf45925c3cebda2880c/easybuild/tools/run.py#L450), I see:
```
    # use bash as shell instead of the default /bin/sh used by subprocess.run
    # (which could be dash instead of bash, like on Ubuntu, see https://wiki.ubuntu.com/DashAsBinSh)
    # stick to None (default value) when not running command via a shell
    if use_bash:
        bash = shutil.which('bash')
        _log.info(f"Path to bash that will be used to run shell commands: {bash}")
        executable, shell = bash, True
    else:
        executable, shell = None, False
```
I tested a build of SentencePiece, including the `LD_PRLOAD` hook:
```
eb --hooks $HOME/EESSI/software-layer/eb_hooks.py SentencePiece-0.2.0-GCC-12.3.0.eb --rebuild
```
with EasyBuild 5.0.X (from the current branch), and that worked _without_ encountering the previous issue.

In other words, there is not much to fix, we just need to wait for EasyBuild 5.X to be released (soon TM)

casparvl:
Hmm, while the issue for SentencePiece is solved (this now installs succesfully), I'm getting

```
  -- Check for working C compiler: /tmp/eb-cw54zzvr/tmprgti6_vm/rpath_wrappers/gcc_wrapper/gcc - broken
  CMake Error at /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/CMake/3.26.3-GCCcore-12.3.0/share/cmake-3.26/Modules/CMakeTestCCompiler.cmake:67 (message):
    The C compiler

      "/tmp/eb-cw54zzvr/tmprgti6_vm/rpath_wrappers/gcc_wrapper/gcc"

    is not able to compile a simple test program.

    It fails with the following output:

      Change Dir: /tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext/text-0.16.2/build/temp.linux-aarch64-cpython-311/CMakeFiles/CMakeScratch/TryCompile-XrjNFV

      Run Build Command(s):/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Ninja/1.11.1-GCCcore-12.3.0/bin/ninja -v cmTC_64b77 && [1/2] /tmp/eb-cw54zzvr/tmprgti6_vm/rpath_wrappers/gcc_wrapper/gcc   -O
2 -ftree-vectorize -mcpu=native -fno-math-errno -o CMakeFiles/cmTC_64b77.dir/testCCompiler.c.o -c /tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext/text-0.16.2/build/temp.linux-aarch64-cpython-311/CMakeFiles/CMakeScra
tch/TryCompile-XrjNFV/testCCompiler.c
      FAILED: CMakeFiles/cmTC_64b77.dir/testCCompiler.c.o
      /tmp/eb-cw54zzvr/tmprgti6_vm/rpath_wrappers/gcc_wrapper/gcc   -O2 -ftree-vectorize -mcpu=native -fno-math-errno -o CMakeFiles/cmTC_64b77.dir/testCCompiler.c.o -c /tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext
/text-0.16.2/build/temp.linux-aarch64-cpython-311/CMakeFiles/CMakeScratch/TryCompile-XrjNFV/testCCompiler.c
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /home/casparvl/eessi/versions/2023.06/software/linux/aarch64/neoverse_n1/software/gperftools/2.12-GCCcore-12.3.0/lib64/libtcmalloc_minimal.so)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.36' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libstdc++.so.6)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.35' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libgcc_s.so.1)
      /bin/sh: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/GCCcore/12.3.0/lib64/libgcc_s.so.1)
      ninja: build stopped: subcommand failed.
```
when it is installing `torchtext` from `PyTorch-Bundle`. I _think_ the `/bin/sh` here comes from the fact that _some_ python process invokes `subprocess.run()`, which uses `/bin/sh` according to https://github.com/easybuilders/easybuild-framework/blob/a2550eb8fab479f517badbf45925c3cebda2880c/easybuild/tools/run.py#L450

The last part of the stack trace I'm getting:

```
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
      dist.run_commands()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
      self.run_command(cmd)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/dist.py", line 1244, in run_command
      super().run_command(command)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
      cmd_obj.run()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/wheel/bdist_wheel.py", line 343, in run
      self.run_command("build")
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/cmd.py", line 318, in run_command
      self.distribution.run_command(command)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/dist.py", line 1244, in run_command
      super().run_command(command)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
      cmd_obj.run()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/command/build.py", line 131, in run
      self.run_command(cmd_name)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/cmd.py", line 318, in run_command
      self.distribution.run_command(command)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/dist.py", line 1244, in run_command
      super().run_command(command)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
      cmd_obj.run()
    File "/tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext/text-0.16.2/tools/setup_helpers/extension.py", line 46, in run
      super().run()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/command/build_ext.py", line 84, in run
      _build_ext.run(self)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python-bundle-PyPI/2023.06-GCCcore-12.3.0/lib/python3.11/site-packages/Cython/Distutils/old_build_ext.py", line 186, in run
      _build_ext.build_ext.run(self)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/command/build_ext.py", line 345, in run
      self.build_extensions()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python-bundle-PyPI/2023.06-GCCcore-12.3.0/lib/python3.11/site-packages/Cython/Distutils/old_build_ext.py", line 195, in build_extensions
      _build_ext.build_ext.build_extensions(self)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/command/build_ext.py", line 467, in build_extensions
      self._build_extensions_serial()
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/setuptools/_distutils/command/build_ext.py", line 493, in _build_extensions_serial
      self.build_extension(ext)
    File "/tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext/text-0.16.2/tools/setup_helpers/extension.py", line 108, in build_extension
      subprocess.check_call(["cmake", str(_ROOT_DIR)] + cmake_args, cwd=self.build_temp)
    File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/subprocess.py", line 413, in check_call
      raise CalledProcessError(retcode, cmd)
```

That's annoying to say the least. We can fix it, but it might require a patch to Python to alter which `sh` command is used by default by `subprocess.run`. Alternatively, we change the `subprocess call done by `/tmp/casparvl/easybuild/build/PyTorchbundle/2.1.2/foss-2023a/torchtext/text-0.16.2/tools/setup_helpers/extension.py`. That's much smaller impact, but also a less complete fix. It means that any other software using SentencePiece and calling `subprocess.run` will still run into this issue.

boegel:
> In other words, there is not much to fix, we just need to wait for EasyBuild 5.X to be released (soon, I hope :D). Or we need to reinstall 4.9.3 with a patch based on [easybuilders/easybuild-framework#4646](https://github.com/easybuilders/easybuild-framework/pull/4646) so we can proceed here.

@casparvl There's an EasyBuild v4.9.4 release coming really soon (in next couple of days), because the GCC easyblock in EasyBuild v4.9.3 has a serious bug that many people will easily run into (see [here](https://github.com/easybuilders/easybuild-easyblocks/pull/3449)), so it's worth trying to get https://github.com/easybuilders/easybuild-framework/pull/4646 merged ASAP.

casparvl:
From the sources, it seems to be equally broken in Gentoo Prefix:
```
$ cat /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib/python3.11/subprocess.py | grep -A 5 "/bin/sh"
    >>> check_output(["/bin/sh", "-c",
    ...               "ls -l non_existent_file ; exit 0"],
    ...              stderr=STDOUT)
    b'ls: non_existent_file: No such file or directory\n'

    There is an additional optional argument, "input", allowing you to
--
                # On Android the default shell is at '/system/bin/sh'.
                unix_shell = ('/system/bin/sh' if
                          hasattr(sys, 'getandroidapilevel') else '/bin/sh')
                args = [unix_shell, "-c"] + args
                if executable:
                    args[0] = executable

            if executable is None:
```

casparvl:
I confirmed that if I run a `subprocess.run("sleep 5", shell=True)` with the python from the compat layer, it will use `/bin/sh` to execute this command. So yes, it's just as broken in the Python in Gentoo-Prefix.

The fix _should_ be very simple: prepend the sysroot to the path on [this line](https://github.com/python/cpython/blob/306368c9c881040e383fa9f43ee671620ff133fc/Lib/subprocess.py#L1814) in the source code. I guess this could (and should) be done at the EasyBlock level. I'll look at that later...

laraPPr:
@Caparvl Can you pick this on up again now that EasyBuild5 is here? Maybe closing and reopening a new pr will be easiest since the pr will also need to retargeted. The hooks file is now also in the [software-layer-scripts](https://github.com/EESSI/software-layer-scripts) repository so those changes will need to be made there.

PR #591, opened 2024-06-04
Update scorecards to latest version + prevent running on forks

ocaisa:


.github/workflows/scorecards.yml
```diff
@@ -4,24 +4,20 @@
...
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
     - cron: '25 15 * * 3'
   push:
-    branches: [ "main" ]
-  pull_request:
-    branches:    
-      - main
+    branches:
+      - '2023.06-software.eessi.io'
```
boegel, line 13: Do we need to limit this to specific branches?

We're bound to forget to update this when we introduce additonal branches, and we're already missing some relevant ones (like `riscv-*`)

ocaisa: There is no riscv-* : https://github.com/EESSI/software-layer/branches/all

boegel: Good enough for now, hoping that a scan for current branch would bring this workflow back in view when we start a new EESSi version...

PR #606, opened 2024-06-13
Dynamically create ReFrame config with correct amount of memory

casparvl:
Make sure we bind-mount `/sys/fs/cgroup` into the container, so that we can use it to determine available memory for dynamically configuring ReFrame in the test step


eessi_container.sh
```diff
@@ -70,40 +70,43 @@ export EESSI_REPOS_CFG_FILE="${EESSI_REPOS_CFG_DIR}/repos.cfg"
...
+  echo "                            tarball's name will be {REPO_ID}-{TIMESTAMP}.tgz"
+  echo "                            [default: not set]"
+  echo "  -v | --verbose          - display more information [default: false]"
+  echo "  -x | --http-proxy URL   - provides URL for the env variable http_proxy"
+  echo "                            [default: not set]; uses env var \$http_proxy if set"
+  echo "  -y | --https-proxy URL  - provides URL for the env variable https_proxy"
+  echo "                            [default: not set]; uses env var \$https_proxy if set"
+  echo "  -b | --extra-bind-paths - specify extra paths to be bound into the container."
+  echo "                            To specify multiple bind paths, seperate by comma."
+  echo "                            Example: '/src:/dest:ro,/src2:/dest2:rw'"
```
bedroge, line 109:
```suggestion
  echo "  -a | --access {ro,rw}   - ro (read-only), rw (read & write) [default: ro]"
  echo "  -b | --extra-bind-paths - specify extra paths to be bound into the container."
  echo "                            To specify multiple bind paths, seperate by comma."
  echo "                            Example: '/src:/dest:ro,/src2:/dest2:rw'"
  echo "  -c | --container IMG    - image file or URL defining the container to use"
  echo "                            [default: docker://ghcr.io/eessi/build-node:debian11]"
  echo "  -f | --fakeroot         - run the container with --fakeroot [default: false]"
  echo "  -g | --storage DIR      - directory space on host machine (used for"
  echo "                            temporary data) [default: 1. TMPDIR, 2. /tmp]"
  echo "  -h | --help             - display this usage information [default: false]"
  echo "  -i | --host-injections  - directory to link to for host_injections "
  echo "                            [default: /..storage../opt-eessi]"
  echo "  -l | --list-repos       - list available repository identifiers [default: false]"
  echo "  -m | --mode MODE        - with MODE==shell (launch interactive shell) or"
  echo "                            MODE==run (run a script or command) [default: shell]"
  echo "  -n | --nvidia MODE      - configure the container to work with NVIDIA GPUs,"
  echo "                            MODE==install for a CUDA installation, MODE==run to"
  echo "                            attach a GPU, MODE==all for both [default: false]"
  echo "  -r | --repository CFG   - configuration file or identifier defining the"
  echo "                            repository to use [default: EESSI via"
  echo "                            default container, see --container]"
  echo "  -u | --resume DIR/TGZ   - resume a previous run from a directory or tarball,"
  echo "                            where DIR points to a previously used tmp directory"
  echo "                            (check for output 'Using DIR as tmp ...' of a previous"
  echo "                            run) and TGZ is the path to a tarball which is"
  echo "                            unpacked the tmp dir stored on the local storage space"
  echo "                            (see option --storage above) [default: not set]"
  echo "  -s | --save DIR/TGZ     - save contents of tmp directory to a tarball in"
  echo "                            directory DIR or provided with the fixed full path TGZ"
  echo "                            when a directory is provided, the format of the"
  echo "                            tarball's name will be {REPO_ID}-{TIMESTAMP}.tgz"
  echo "                            [default: not set]"
  echo "  -v | --verbose          - display more information [default: false]"
  echo "  -x | --http-proxy URL   - provides URL for the env variable http_proxy"
  echo "                            [default: not set]; uses env var \$http_proxy if set"
  echo "  -y | --https-proxy URL  - provides URL for the env variable https_proxy"
  echo "                            [default: not set]; uses env var \$https_proxy if set"
```

bedroge: The diff looks really weird, but I just wanted to keep it sorted alphabetically :smile:

eessi_container.sh
```diff
@@ -197,6 +200,10 @@ while [[ $# -gt 0 ]]; do
       export https_proxy=${HTTPS_PROXY}
       shift 2
       ;;
+    -b|--extra-bind-paths)
```
bedroge, line 203: Can you move this up as well, so the options are ordered alphabetically?

eessi_container.sh
```diff
@@ -70,40 +70,43 @@ export EESSI_REPOS_CFG_FILE="${EESSI_REPOS_CFG_DIR}/repos.cfg"
...
-  echo "                           tarball's name will be {REPO_ID}-{TIMESTAMP}.tgz"
-  echo "                           [default: not set]"
-  echo "  -v | --verbose         - display more information [default: false]"
-  echo "  -x | --http-proxy URL  - provides URL for the env variable http_proxy"
-  echo "                           [default: not set]; uses env var \$http_proxy if set"
-  echo "  -y | --https-proxy URL - provides URL for the env variable https_proxy"
-  echo "                           [default: not set]; uses env var \$https_proxy if set"
+  echo "  -a | --access {ro,rw}   - ro (read-only), rw (read & write) [default: ro]"
+  echo "  -b | --extra-bind-paths - specify extra paths to be bound into the container."
+  echo "                            To specify multiple bind paths, seperate by comma."
```
bedroge, line 75:
```suggestion
  echo "                            To specify multiple bind paths, separate by comma."
```

bedroge (approved):
Looks good to me.

PR #607, opened 2024-06-13
Fix `--extra-bind-paths` argument in `bot/test.sh`

bedroge:
This should fix:
```
ERROR: Unknown option: --extra-bind-paths /sys/fs/cgroup:/hostsys/fs/cgroup:ro
```


casparvl (approved):
Lgtm

PR #608, opened 2024-06-13
add support in archdetect for detecting A64FX

boegel:


init/arch_specs/eessi_arch_arm.spec
```diff
@@ -1,5 +1,6 @@
 # ARM CPU architecture specifications (see https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html for guidance)
 # Software path in EESSI 	| Vendor ID 	| List of defining CPU features
+"aarch64/aarch64"	""		"asimdhp sve"		# Fujitsu A64FX
```
ocaisa, line 3:
```suggestion
"aarch64/a64fx"	""		"asimdhp sve"		# Fujitsu A64FX
```

boegel: This wasn't the only issue, fixed both in cb1672b with force push

init/arch_specs/eessi_arch_arm.spec
```diff
@@ -1,5 +1,6 @@
 # ARM CPU architecture specifications (see https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html for guidance)
 # Software path in EESSI 	| Vendor ID 	| List of defining CPU features
+"aarch64/a64fx"	""		"asimdhp sve"		# Fujitsu A64FX
```
boegel, line 3: Hmm, this doesn't fully work, because `asimdhp` is also available for `neoverse_v1`, which also has features that start with `sve` (like `svei8mm`).

The CPU features check should be checking for *exact* strings though, and not consider `svei8mm` as a valid match for `sve`...

BMalaca: I think both have `sve` according to archspec:
```bash
[malaca@cna0002 A64FX_auto_detection]$ python3 -c "from archspec import cpu;import archspec; print(archspec.cpu.TARGETS['neoverse_v1'].features)"
{'flagm', 'pmull', 'rng', 'asimd', 'dgh', 'bf16', 'cpuid', 'asimdfhm', 'evtstrm', 'fphp', 'jscvt', 'ilrcpc', 'fp', 'i8mm', 'asimdrdm', 'dcpop', 'dcpodp', 'aes', 'sha2', 'asimdhp', 'fcma', 'sha3', 'crc32', 'svebf16', 'sha1', 'uscat', 'sha512', 'lrcpc', **'sve'**, 'dit', 'asimddp', 'svei8mm', 'atomics'}
```

boegel: The problem is that we'll need to take into account the processor vendor to discriminate them, just looking at CPU features is not enough here...

BMalaca: From what I tested locally, adding 0x46 in the Vendor ID solves both a64fx and graviton 3 identifications. You still need to create a file: tests/archdetect/aarch64/a64fx/Deucalion-Rocky85.output thate should read `aarch64/a64fx:aarch64/generic`

boegel: @BMalaca Thanks for the feedback & help!

Not exactly, since only the "`Vendor ID`" field from `/proc/cpuinfo` is considered.

That doesn't exist for `aarch64`, there we need to grab the "`CPU implementer`" field, so we need to update the `eessi_archdetect.sh` accordingly.

I'll look into that, shouldn't be much work...

boegel:
bot: build repo:eessi.io-2023.06-software arch:aarch64/neoverse_n1

init/arch_specs/eessi_arch_arm.spec
```diff
@@ -1,6 +1,9 @@
...
+# CPU implementers: 0x41 (ARM), 0x46 (Fujitsu) - also see https://github.com/hrw/arm-socs-table/blob/main/data/socs.yml
+
+# Software path in EESSI 	| 'Vendor ID' or 'CPU implementer' 	| List of defining CPU features
+"aarch64/a64fx"		"0x46"		"asimdhp sve"		# Fujitsu A64FX
 "aarch64/neoverse_n1"	"ARM"		"asimddp"		# Ampere Altra
-"aarch64/neoverse_n1"	""		"asimddp"		# AWS Graviton2
+"aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
-"aarch64/neoverse_v1"	""		"asimddp svei8mm"	# AWS Graviton3
+"aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
```
ocaisa, line 9: What happens when there are multiple vendors implementing the instruction set, like with Gravition/Grace for "aarch64/neoverse_v2"

ocaisa: I guess this is only necessary when there is a doubt about what the match is

ocaisa: Nevermind, it is right there in the file, just means we need to add each supported CPU explicitly. That seems like a bit of a pity, doesn't it make the archdetect selection very conservative if it encounters a CPU it hasn't seen before? It doesn't matter if the CPU has all the required features

boegel:
`eb_hooks.py` is in the generated tarball, but untouched in this PR...
Did we overlook a deploy in a previous already merged PR?

ocaisa:
Sure did, https://github.com/EESSI/software-layer/pull/841

init/arch_specs/eessi_arch_arm.spec
ocaisa: > For `neoverse_v2` with Graviton 4 vs Grace, there's a more complex situation though, because Graviton 4 supports CPU instructions like `paca`, `pacg`, `rng` which Grace doesn't, and the other way around (like `sm3`, `sm4`, `svesm4`). It gets even more interesting when Google Axion is taken into account, since that doesn't support `sbss`, which the other two do...

Isn't that quite a big problem? Can we disable some instructions? I guess compilers would be caught by surprise...

ocaisa: This data is probably pretty useful for us: https://github.com/hrw/arm-socs-table/tree/main/cpuinfo-data

ocaisa: I wonder if we shouldn't have a much more considered set of CPU features that reflect the CPU where the software was built for the target and then match that? This would do a much better job of future matching CPUs. Is it really the case that CPU features alone are not enough to distinguish A64FX from Neoverse_N1?

Not going to hold this PR back for that discussion, will open an issue.

boegel: I'm also wondering if we can instruct the compiler to not emit particular instructions even if they're supported.
My gut says that should be possible, but I'm not sure it actually is...

Let's follow up on that in a dedicated issue though => https://github.com/EESSI/software-layer/issues/845

bedroge (approved):

init/arch_specs/eessi_arch_arm.spec
boegel: For A64FX vs Neoverse N1/V1, see the `/proc/cpuinfo` dumps we have under `tests/archdetect/`.

bedroge:
The tarball has been approved and ingested.

Issue #25, opened 2020-09-30
OpenFOAM 8 in broadwell tree broken


bedroge:
By running our OpenFOAM test script on broadwell, I found that lots of files are missing in the installation folder, especially in the tutorials folder, e.g. the following dirs are completely empty (and shouldn't be):
```
/cvmfs/pilot.eessi-hpc.org/2020.09/software/x86_64/intel/broadwell/software/OpenFOAM/8-foss-2020a/OpenFOAM-8/tutorials/resources/geometry/

/cvmfs/pilot.eessi-hpc.org/2020.09/software/x86_64/intel/broadwell/software/OpenFOAM/8-foss-2020a/OpenFOAM-8/tutorials/incompressible/simpleFoam/motorBike/system/

/cvmfs/pilot.eessi-hpc.org/2020.09/software/x86_64/intel/broadwell/software/OpenFOAM/8-foss-2020a/OpenFOAM-8/tutorials/resources/blockMesh/
```

and more...

The pilot build script did crash at some point, probably during the installation of OpenFOAM. Could it be that a reinstall for OpenFOAM does something weird due to building in the installation dir?

bedroge:
Fixed it for now by recompiling OpenFOAM. Maybe we can still test what happens if you abort an OpenFOAM installation somewhere near the end, and try to reinstall it?

boegel:
We now run a tutorial case in the sanity check for OpenFOAM (see https://github.com/easybuilders/easybuild-easyblocks/pull/2201), so we should catch problems like this in the future...

PR #623, opened 2024-06-26
Workaround to fix broken curl installation

hvelab:
As reported in the EESSI support portal issue [#25](https://gitlab.com/eessi/support/-/issues/25), curl fails in RHEL 8 and above systems as CA files location differs. Works for all rhel based systems as I could reproduce the issue also in them.

```
$ cat /etc/os-release 
NAME="Rocky Linux"
VERSION="9.3 (Blue Onyx)"
ID="rocky"
ID_LIKE="rhel centos fedora"

$ source software-layer/init/bash 
Found EESSI repo @ /cvmfs/software.eessi.io/versions/2023.06!
archdetect says x86_64/intel/skylake_avx512
Using x86_64/intel/skylake_avx512 as software subdirectory.
[...]
Environment set up to use EESSI (2023.06), have fun!

{EESSI 2023.06} $ ml CMake
{EESSI 2023.06} $ curl https://www.example.com
<!doctype html>
<html>
<head>
    <title>Example Domain</title>
[...]
```

I am not a fan of hardcording in the init script but investigating the issue I haven't found a more suitable way to solve it for our use case. Maybe another solution would be to add a modlua footer exporting this variable in the affected modules?

### DIFF
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -15,6 +15,20 @@ function show_msg {
   fi
 }
 
+function check_rhel_ca {
+
+  os=$(grep -oP '(?<=^ID_LIKE=).+' /etc/os-release | tr -d '"')
+
+  version_id=$(grep -oP '(?<=^VERSION_ID=).+' /etc/os-release | tr -d '"')
+  version_major=${version_id%\.*}
+
+  if [[ $os =~ "rhel" ]] && [ $version_major -gt 7 ]
+  then
+        export CURL_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
+  fi
+}
+
+
 # set up minimal environment: $EESSI_PREFIX, $EESSI_VERSION, $EESSI_OS_TYPE, $EESSI_CPU_FAMILY, $EPREFIX
 source $EESSI_INIT_DIR_PATH/minimal_eessi_env
 
@@ -106,6 +120,8 @@ if [ -d $EESSI_PREFIX ]; then
             false
           fi
 
+        # Fix wrong path for RHEL >=8 libcurl 
+	check_rhel_ca
 
         else
           error "EESSI software layer at $EESSI_SOFTWARE_PATH not found!"
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -15,20 +15,6 @@ function show_msg {
   fi
 }
 
-function check_rhel_ca {
-
-  os=$(grep -oP '(?<=^ID_LIKE=).+' /etc/os-release | tr -d '"')
-
-  version_id=$(grep -oP '(?<=^VERSION_ID=).+' /etc/os-release | tr -d '"')
-  version_major=${version_id%\.*}
-
-  if [[ $os =~ "rhel" ]] && [ $version_major -gt 7 ]
-  then
-        export CURL_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
-  fi
-}
-
-
 # set up minimal environment: $EESSI_PREFIX, $EESSI_VERSION, $EESSI_OS_TYPE, $EESSI_CPU_FAMILY, $EPREFIX
 source $EESSI_INIT_DIR_PATH/minimal_eessi_env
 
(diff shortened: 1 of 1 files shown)


init/eessi_environment_variables
```diff
@@ -15,6 +15,20 @@ function show_msg {
...
+function check_rhel_ca {
+
+  os=$(grep -oP '(?<=^ID_LIKE=).+' /etc/os-release | tr -d '"')
+
+  version_id=$(grep -oP '(?<=^VERSION_ID=).+' /etc/os-release | tr -d '"')
+  version_major=${version_id%\.*}
+
+  if [[ $os =~ "rhel" ]] && [ $version_major -gt 7 ]
+  then
+        export CURL_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
```
ocaisa, line 27: Given that you know the path where the certs should be, isn't it enough to check if it exists and if so export the variable?

ocaisa:
Currently we ship `curl` in the compat layer, so there's no way to avoid setting this in the initialisation script. If it ever came to be that we stopped doing that, then it would be enough to set this via an Lmod hook

init/eessi_environment_variables
ocaisa: No need for an OS check with that approach

hvelab: True! I wanted to make sure that nothing breaks, but you are right, my implementation was over-engineered, now its much more simple

ocaisa (changes requested):
Looks good but can you add a comment to mention that if curl we're removed from the compat layer we could do this via the cURL module file (or an Lmod hook) instead

init/eessi_environment_variables
```diff
@@ -106,6 +106,12 @@ if [ -d $EESSI_PREFIX ]; then
             false
           fi
 
+        # Fix wrong path for RHEL >=8 libcurl 
+	rhel_libcurl_file="/etc/pki/tls/certs/ca-bundle.crt"
```
ocaisa, line 110:
```suggestion
	# This is required here because we ship curl in our compat layer. If we only provided
	# curl as a module file we could instead do this via a `modluafooter` in an EasyBuild
	# hook (or via an Lmod hook)
	rhel_libcurl_file="/etc/pki/tls/certs/ca-bundle.crt"
```

init/eessi_environment_variables
```diff
@@ -106,6 +106,15 @@ if [ -d $EESSI_PREFIX ]; then
...
 
+        # Fix wrong path for RHEL >=8 libcurl 
+	# This is required here because we ship curl in our compat layer. If we only provided
+	# curl as a module file we could instead do this via a `modluafooter` in an EasyBuild
+	# hook (or via an Lmod hook)
+	rhel_libcurl_file="/etc/pki/tls/certs/ca-bundle.crt"
+        if [ -f $rhel_libcurl_file ]; then
+          show_msg "Found libcurl CAs file at RHEL location, setting CURL_CA_BUNDLE"
+          export CURL_CA_BUNDLE=$rhel_libcurl_file
+        fi
```
ocaisa, line 117:
```suggestion
          fi

          # Fix wrong path for RHEL >=8 libcurl 
	  # This is required here because we ship curl in our compat layer. If we only provided
	  # curl as a module file we could instead do this via a `modluafooter` in an EasyBuild
	  # hook (or via an Lmod hook)
	  rhel_libcurl_file="/etc/pki/tls/certs/ca-bundle.crt"
          if [ -f $rhel_libcurl_file ]; then
            show_msg "Found libcurl CAs file at RHEL location, setting CURL_CA_BUNDLE"
            export CURL_CA_BUNDLE=$rhel_libcurl_file
          fi
```
indentation was off

ocaisa (changes requested):

xinan1911:
@ocaisa Could you take a final check before merging this PR?

ocaisa:
I can't merge this because I don't know how to deploy it

bedroge:
I'm not completely sure anymore if this can now be done with the bot. Let's try it, and otherwise I can do it manually.

bedroge:

bedroge:
Looks like it actually worked :tada:.

I also tested the fix on our Rocky 8 cluster: I got an error without the fix, but after applying the fix the curl command worked fine.

PR #618, opened 2024-06-17
support for multiple CVMFS repositories in `eessi_container.sh`

trz42:
Support for multiple CVMFS repositories is needed for `dev.eessi.io`.

Ready for review. Example test:
```bash
# ./eessi_container.sh --repository software.eessi.io --repository dev.eessi.io,access=rw
Using /tmp/eessi.jLuBerqoFu as tmp directory (to resume session add '--resume /tmp/eessi.jLuBerqoFu').
Pulling container image from docker://ghcr.io/eessi/build-node:debian11 to /tmp/eessi.jLuBerqoFu/ghcr.io_eessi_build_node_debian11.sif
Launching container with command (next line):
singularity -q shell  --fusemount container:cvmfs2 cvmfs-config.cern.ch /cvmfs/cvmfs-config.cern.ch --fusemount container:cvmfs2 software.eessi.io /cvmfs/software.eessi.io --fusemount container:cvmfs2 dev.eessi.io /cvmfs_ro/dev.eessi.io --fusemount container:fuse-overlayfs -o lowerdir=/cvmfs_ro/dev.eessi.io -o upperdir=/tmp/dev.eessi.io/overlay-upper -o workdir=/tmp/dev.eessi.io/overlay-work /cvmfs/dev.eessi.io /tmp/eessi.jLuBerqoFu/ghcr.io_eessi_build_node_debian11.sif
INFO:    Environment variable SINGULARITY_BIND is set, but APPTAINER_BIND is preferred
INFO:    Environment variable SINGULARITY_HOME is set, but APPTAINER_HOME is preferred
CernVM-FS: pre-mounted on file descriptor 3
CernVM-FS: pre-mounted on file descriptor 3
CernVM-FS: pre-mounted on file descriptor 3
Apptainer> CernVM-FS: loading Fuse module... done
CernVM-FS: loading Fuse module... CernVM-FS: loading Fuse module... done
done

Apptainer> touch /cvmfs/dev.eessi.io/foo
Apptainer> touch /cvmfs/software.eessi.io/foo
touch: cannot touch '/cvmfs/software.eessi.io/foo': Permission denied
```

Partially fixes #616 (doesn't provide repository-specific bind mounts for `host_injections`).

PR #632, opened 2024-07-04
{2023.06,a64fx}[foss/2023a] ParaView 5.11.2

boegel:

PR #633, opened 2024-07-04
add missing `/cvmfs/` in `$EESSI_CVMFS_REPO_OVERRIDE` defined by `bot/build.sh` and `bot/test.sh`

boegel:
This is required after the changes in #618, which used to include this in `eessi_container.sh` (but not anymore):
```shell
export EESSI_CVMFS_REPO_OVERRIDE="/cvmfs/${repo_name}"
```

Removing that exposed the incorrect setting of `$EESSI_CVMFS_REPO_OVERRIDE` in `bot/build.sh`, which never was correct without the `/cvmfs/`


boegel:
Build for #632 is actually running now, which confirms this fix

bedroge (approved):

PR #634, opened 2024-07-05
{2023.06}[GCCcore/12.3.0-GCC/12.3.0-gompi/2023a] bio-packages

TopRichard:
```
missing packages :

BWA-0.7.17-GCCcore-12.3.0.
CapnProto-1.0.1-GCCcore-12.3.0
DendroPy-4.6.1-GCCcore-12.3.0
DIAMOND-2.1.8-GCC-12.3.0 
FastME-2.1.6.3-GCC-12.3.0
fastp-0.23.4-GCC-12.3.0
HMMER-3.4-gompi-2023a
IQ-TREE-2.3.5-gompi-2023a
KronaTools-2.8.1-GCCcore-12.3.0
LSD2-2.4.1-GCCcore-12.3.0
MAFFT-7.520-GCC-12.3.0-with-extensions
ncbi-vdb-3.0.10-gompi-2023a
```

PR #635, opened 2024-07-05
Fix path to overlay upper directory in create_tarball.sh

bedroge:
#618 introduced a new structure where each repo has its own overlay-upper dir, see https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/eessi_container.sh#L410 and https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/eessi_container.sh#L744. This was not changed in the tarball script yet, leading to errors in https://github.com/EESSI/software-layer/pull/634:
```
Software directory overlay /tmp/overlay-upper/versions/2023.06 does not exist?!
```

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -26,7 +26,7 @@ if [ ! -d ${software_dir} ]; then
     exit 2
 fi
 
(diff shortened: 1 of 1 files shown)


Neves-P (approved):
LGTM!

PR #636, opened 2024-07-05
Remove `/cvmfs/` in path to overlay upper directory in tarball script

bedroge:
Follow-up to #635 , which didn't fully fix the issue. The `cvmfs_repo` variable not only contained the repo name, but also had a leading `/cvmfs`. This has to be removed when setting the path to the overlay upper directory, which looks like `/tmp/<repo name>/overlay-upper`.

### DIFF
--- a/create_tarball.sh
+++ b/create_tarball.sh
@@ -26,7 +26,8 @@ if [ ! -d ${software_dir} ]; then
     exit 2
 fi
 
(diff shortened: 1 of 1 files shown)


bedroge:
Let's test this in #634 before merging it.

bedroge:
The build completed successfully (https://github.com/EESSI/software-layer/pull/634/#issuecomment-2211359075), so this fix indeed solves the issue.

boegel (approved):

PR #645, opened 2024-07-19
Remove duplicate testing of PRs

ocaisa:
Right now we are testing the commit via the PR and the push (if someone silly like me just creates a patch via the GitHub interface)


ocaisa:
These also removes all (or at least most) of the testing on peoples forks

bedroge (approved):

Issue #68, opened 2021-02-05
provide module file to set up EESSI environment


boegel:
Idea by @akesandgren, a module file that sites can install on their system to provide easy access to EESSI.

Instead of init script to be sourced.

Reasons:

* rollback of changes to environment
* customisations where needed

@akesandgren Anything to add?

akesandgren:
That's the basic ideas for it at least. Can't currently come up with anything else.

Just being able to hop between SW-stacks, EESSI, CC, Site-stuff, would be nice and even somewhat fancy.

akesandgren:
here is a sketch of how to do things....
```
local posix = require("posix")
require("capture")

whatis([[Description: EESSI Software environment]])

help([[
Sets up the environment to use the EESSI software stack.
]])

conflict("EESSI")

local eessi_pilot_version = "2020.12"
local eessi_prefix = "/cvmfs/pilot.eessi-hpc.org/"..eessi_pilot_version

local osname           = posix.uname("%s")
if osname == "Linux" then
    setenv("EESSI_OS_TYPE", "linux")
else
    setenv("EESSI_OS_TYPE", "macos")
end
setenv("EESSI_CPU_FAMILY", posix.uname("%m"))
setenv("EPREFIX", eessi_prefix.."/compat/"..os.getenv("EESSI_OS_TYPE").."/"..os.getenv("EESSI_CPU_FAMILY"))
local q=capture("echo hejsan åke")
setenv("AKE", q)
```
the capture("echo hejsan åke") is a short substitute for calling the ${EESSI_PREFIX}/init/eessi_software_subdir_for_host.py script.

klust:
Here is my current take on a module. Currently it requires the system manager to set a few things by hand (basically the OS, though  I don't think it is anywhere ready for macOS) and the processor as would be retuned by archspec cpu.
```lua
local eessi_version = "2020.12"
local eessi_root = "/cvmfs/pilot.eessi-hpc.org/"

-- Discovering which software directories to use:
-- - Option 1: Detect the cpu architecture as returned by archspec cpu 
--   and use a mapping in the module file to map that onto the triple 
--   for the host type. That mapping table could always be generated
--   offline during the building of a new EESSI stack.
-- - Option 1bis: Simply ask system managers who want to provide EESSI
--   through a module to set a certain environment variable which this
--   module then picks up. This can avoid calling external programs
--   that may or may not slow down the loading and unloading of the
--   module.
-- - Option 2:
[...]

PR #667, opened 2024-08-12
EESSI bash initialization to module file

TopRichard:
This is a follow up PR to the issue: https://gitlab.com/eessi/support/-/issues/83

The CI tests are intended to evaluate the following:
```
Check the archdetect function works when there are no valid paths
Check the archdetect function works (by setting a value with valid path in the middle), Load the module and check the existence and values a few expected variables
```


init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+
+local eessi_version = myModuleVersion()
+local eessi_repo = "/cvmfs/software.eessi.io"
+local eessi_prefix = pathJoin(eessi_repo, "versions", eessi_version)
+local eessi_os_type = "linux"
+setenv("EESSI_VERSION", eessi_version)
+setenv("EESSI_CVMFS_REPO", eessi_repo)
+setenv("EESSI_OS_TYPE", eessi_os_type)
+function archdetect_cpu()
+    local script = pathJoin(eessi_repo, 'versions', eessi_version, 'init', 'lmod_eessi_archdetect_wrapper.sh')
```
ocaisa, line 21:
```suggestion
    local script = pathJoin(eessi_prefix, 'init', 'lmod_eessi_archdetect_wrapper.sh')
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+    if not os.getenv("EESSI_ARCHDETECT_OPTIONS") then
+        if convertToCanonical(LmodVersion()) < convertToCanonical("8.6") then
+            LmodMessage("Loading this modulefile requires using Lmod version > 8.6, but you can export EESSI_ARCHDETECT_OPTIONS to the available cpu architecture in the form of: x86_64/intel/haswell or aarch64/neoverse_v1")
+            os.exit(1)
```
ocaisa, line 25:
```suggestion
            LmodError("Loading this modulefile requires using Lmod version >= 8.6, but you can export EESSI_ARCHDETECT_OPTIONS to the available cpu architecture in the form of: x86_64/intel/haswell:x86_64/generic or aarch64/neoverse_v1:aarch64/generic")
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+        end
+        source_sh("bash", script)
+    end
+    for archdetect_filter_cpu in string.gmatch(os.getenv("EESSI_ARCHDETECT_OPTIONS"), "([^" .. ":" .. "]+)") do
+        if isDir(pathJoin(string.gsub(script, "init/lmod_eessi_archdetect_wrapper.sh", "software/" .. eessi_os_type), archdetect_filter_cpu)) then
```
ocaisa, line 30:
```suggestion
        if isDir(pathJoin(eessi_prefix, "software", eessi_os_type, archdetect_filter_cpu, "software")) then
```

ocaisa: We can test this by setting strange option for `EESSI_ARCHDETECT_OPTIONS`:
```
export EESSI_ARCHDETECT_OPTIONS="path/does/not exist:x86_64/generic:other/path/that/does/not/exist"
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+        end
+    end
+    LmodError("Software directory check for the detected architecture failed")
+end
+local archdetect = archdetect_cpu()
+local eessi_cpu_family = archdetect:match("([^/]+)")
+local eessi_software_subdir = os.getenv("EESSI_SOFTWARE_SUBDIR_OVERRIDE") or archdetect
+local eessi_eprefix = pathJoin(eessi_prefix, "compat", eessi_os_type, eessi_cpu_family)
+local eessi_software_path = pathJoin(eessi_prefix, "software", eessi_os_type, eessi_software_subdir)
+local eessi_module_path = pathJoin(eessi_software_path, "modules/all")
```
ocaisa, line 41:
```suggestion
local eessi_module_path = pathJoin(eessi_software_path, "modules", "all")
local eessi_site_module_path = string.gsub(eessi_module_path, "versions", "host_injections")
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+setenv("EESSI_SITE_MODULEPATH", string.gsub(eessi_module_path, eessi_repo, "host_injections"))
```
ocaisa, line 42:
```suggestion
setenv("EESSI_SITE_MODULEPATH", eessi_site_module_path)
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+setenv("EESSI_SOFTWARE_SUBDIR", eessi_software_subdir)
+setenv("EESSI_PREFIX", eessi_prefix)
+setenv("EESSI_EPREFIX", eessi_eprefix)
```
ocaisa, line 45: Maybe we should use `pushenv`? What Lmod version supports that?

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+prepend_path("PATH", pathJoin(eessi_eprefix, "bin"))
+prepend_path("PATH", pathJoin(eessi_eprefix, "usr/bin"))
+setenv("EESSI_SOFTWARE_PATH", eessi_software_path)
+setenv("EESSI_MODULEPATH", eessi_module_path)
+prepend_path("MODULEPATH", os.getenv("EESSI_SITE_MODULEPATH") .. ":" .. os.getenv("EESSI_MODULEPATH"))
```
ocaisa, line 50:
```suggestion
prepend_path("MODULEPATH", eessi_module_path)
prepend_path("MODULEPATH", eessi_site_module_path)
```

ocaisa (commented):
Some comments from a visual review, will take it for a test drive. 

We should also add CI to test this module file (can be done using the EESSI GitHub Action)

init/modules/EESSI/2023.06.lua
ocaisa: Since 2013, https://github.com/TACC/Lmod/commit/49d44a8fefadcd0e4fbc3b7881713be0718ed499 so not really a worry

ocaisa:
I'm not sure if it is really a concern, but we add an Lmod `family()` related to the `EESSI` as well so that we have a way to manage "compatible" modules (for example module files related to `dev.eessi.io`)

ocaisa:
This also doesn't cover the cURL issue currently fixed in our initialisation script.

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,55 @@
...
+setenv("LMOD_CONFIG_DIR", pathJoin(eessi_software_path, ".lmod"))
+setenv("LMOD_PACKAGE_PATH", pathJoin(eessi_software_path, ".lmod"))
+if mode() == "load" then
+    LmodMessage("EESSI/" .. eessi_version .. " loaded successfully")
+end
```
MaKaNu, line 55: Should we manipulate PS1 and other variables below inside the if statment?
While I was able to successful implement a solution for prompts like ohmyposh or starship, my implementations for PS1 failed.

It seems, reading the current PS1 via `os.getenv("PS1")` result in nil value. Setting the PS1 inside the bash init script is working, but I think it should be handled here, so the unload could also manipulate it.

MaKaNu: The solution for `ohmyposh` and co:

```suggestion
    setenv("EESSI_ENABLED", 1)
    setenv("EESSI_PROMPT", "{EESSI " .. eessi_version .. "}")
end
```

this allows as example for ohmyposh following config:

```toml
[[blocks.segments]]
                type = 'command'
                style = 'plain'
                background = "transparent"
                foreground = "#14AD9E" # NICE EESSI COLOR
                template = "{{ if .Output }}{{ .Env.EESSI_PROMPT }}{{ end }}"
                [blocks.segments.properties]
                        shell = "sh"
                        command = "echo $EESSI_ENABLED"
```

MaKaNu (changes requested):

ocaisa:
The one other thing missing is the current redirection we do for Zen4:

https://github.com/EESSI/software-layer/blob/1fc0cb75a076a37184b9abddba2093118bdcfca6/init/eessi_environment_variables#L51-L59

init/modules/EESSI/2023.06.lua
ocaisa: > It seems, reading the current PS1 via `os.getenv("PS1")` result in nil value. Setting the PS1 inside the bash init script is working, but I think it should be handled here, so the unload could also manipulate it.

This seems unusual, as I see previous Lmod posts talk about setting/restoring `PS1`. Is it set in the environment before loading the module? `pushenv()` should be pretty safe here (and shouldn't be guarded). We can make it optional by guarding it with an `if not os.getenv("EESSI_RETAIN_PROMPT") then`

ocaisa:
A quick test, including my suggested changes, shows that this is missing `EESSI_CPU_FAMILY`. This seems to be the only one of consequence

ocaisa:
Here's the one that works for me (but still misses `EPREFIX` and `EESSI_CPU_FAMILY`) and includes the cert bundle check and PS1 change:
```lua
help([[
Description
===========
The European Environment for Scientific Software Installations (EESSI, pronounced as easy) is a collaboration between different European partners in HPC community.The goal of this project is to build a common stack of scientific software installations for HPC systems and beyond, including laptops, personal workstations and cloud infrastructure. 

More information
================
 - URL: https://www.eessi.io/docs/
]])
whatis("Description: The European Environment for Scientific Software Installations (EESSI, pronounced as easy) is a collaboration between different European partners in HPC community. The goal of this project is to build a common stack of scientific software installations for HPC systems and beyond, including laptops, personal workstations and cloud infrastructure.")
whatis("URL: https://www.eessi.io/docs/:")

local eessi_version = myModuleVersion()
local eessi_repo = "/cvmfs/software.eessi.io"
local eessi_prefix = pathJoin(eessi_repo, "versions", eessi_version)
local eessi_os_type = "linux"
pushenv("EESSI_VERSION", eessi_version)
pushenv("EESSI_CVMFS_REPO", eessi_repo)
pushenv("EESSI_OS_TYPE", eessi_os_type)
function archdetect_cpu()
    local script = pathJoin(eessi_prefix, 'init', 'lmod_eessi_archdetect_wrapper.sh')
    if not os.getenv("EESSI_ARCHDETECT_OPTIONS") then
        if convertToCanonical(LmodVersion()) < convertToCanonical("8.6") then
            LmodError("Loading this modulefile requires using Lmod version > 8.6, but you can export EESSI_ARCHDETECT_OPTIONS to the available cpu architecture in the form of: x86_64/intel/haswell or aarch64/neoverse_v1")
        end
        source_sh("bash", script)
    end
    for archdetect_filter_cpu in string.gmatch(os.getenv("EESSI_ARCHDETECT_OPTIONS"), "([^" .. ":" .. "]+)") do
        if isDir(pathJoin(eessi_prefix, "software", eessi_os_type, archdetect_filter_cpu, "software")) then
            return archdetect_filter_cpu
        end
    end
    LmodError("Software directory check for the detected architecture failed")
end
local archdetect = archdetect_cpu()
local eessi_cpu_family = archdetect:match("([^/]+)")
local eessi_software_subdir = os.getenv("EESSI_SOFTWARE_SUBDIR_OVERRIDE") or archdetect
local eessi_eprefix = pathJoin(eessi_prefix, "compat", eessi_os_type, eessi_cpu_family)
local eessi_software_path = pathJoin(eessi_prefix, "software", eessi_os_type, eessi_software_subdir)
local eessi_module_path = pathJoin(eessi_software_path, "modules", "all")
local eessi_site_module_path = string.gsub(eessi_module_path, "versions", "host_injections")
pushenv("EESSI_SITE_MODULEPATH", eessi_site_module_path)
pushenv("EESSI_SOFTWARE_SUBDIR", eessi_software_subdir)
pushenv("EESSI_PREFIX", eessi_prefix)
pushenv("EESSI_EPREFIX", eessi_eprefix)
prepend_path("PATH", pathJoin(eessi_eprefix, "bin"))
prepend_path("PATH", pathJoin(eessi_eprefix, "usr/bin"))
pushenv("EESSI_SOFTWARE_PATH", eessi_software_path)
pushenv("EESSI_MODULEPATH", eessi_module_path)
prepend_path("MODULEPATH", eessi_module_path)
prepend_path("MODULEPATH", eessi_site_module_path)
pushenv("LMOD_CONFIG_DIR", pathJoin(eessi_software_path, ".lmod"))
pushenv("LMOD_PACKAGE_PATH", pathJoin(eessi_software_path, ".lmod"))
-- update the prompt (unless overridden)
if not os.getenv("EESSI_RETAIN_PROMPT") then
  pushenv("PS1", "{EESSI " .. eessi_version .. "} " .. (os.getenv("PS1") or ""))
end
-- check for RHEL certificate locatioin
local rhel_certificates = "/etc/pki/tls/certs/ca-bundle.crt"
if isFile(rhel_certificates) then
  pushenv("", rhel_certificates)
end 
if mode() == "load" then
    LmodMessage("EESSI/" .. eessi_version .. " loaded successfully")
end
```

MaKaNu:
Setting the PS1 only would create difficulties for prompts, which don't use the PS1 Variable (Starship, OhMyPosh, etc).
An easy solution could be the following 

> ```diff
> if not os.getenv("EESSI_RETAIN_PROMPT") then
>-  pushenv("PS1", "{EESSI " .. eessi_version .. "} " .. (os.getenv("PS1") or ""))
>+ local eessi_prompt = "{EESSI " .. eessi_version .. "}"
>+ pushenv("PS1", eessi_prompt .. (os.getenv("PS1") or ""))
>+ pushenv("EESSI_PROMPT", eessi_prompt)
> end
> ```

But the `os.getenv("PS1")` still results in nil and so the original prompt will be gone. I have no clue why the variable could not be read by lua, since the bash script has no issues with it.

TopRichard:
> The one other thing missing is the current redirection we do for Zen4:
> 
> https://github.com/EESSI/software-layer/blob/1fc0cb75a076a37184b9abddba2093118bdcfca6/init/eessi_environment_variables#L51-L59

Handled in the last commit

MaKaNu:
> We can only set `PS1` if it is in the environment already
> 
> ```
> if os.getenv("PS1") and not os.getenv("EESSI_RETAIN_PROMPT") then
> ```

That's the Thing I don't get. In my test scenario, `PS1` was set and was also readable by the bash script, but `os.getenv("PS1")` returns the nil. I believe the problem is that the PS1 variable can't be exported, so it's not available in the module.

MaKaNu:
might it be possible that your jobs already fail at test "Test for archdetect_cpu functionality with only one valid path" and not as indicated at test "Test for archdetect_cpu functionality with invalid path".

`/home/runner/work/_temp/c11c9c6f-66bb-4412-991a-68d31eb8184c.sh: line 4: module: command not found`

This causes the if to fail, but the job task succeed.

ocaisa:
@TopRichard, @MaKaNu is right, you don't have an initialised Lmod when you run the tests. You'll need to add another step that does that, see https://github.com/EESSI/software-layer/blob/b14cef843a13e51c2871fd9a4d8e4d7e74cb349a/init/bash from @MaKaNu PR

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,75 @@
...
+          cvmfs_http_proxy: DIRECT
+          cvmfs_repositories: software.eessi.io
+              
+      - name: Test for archdetect_cpu functionality with only one valid path
+        run: |
+          export MODULEPATH=/cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/modules/:$MODULEPATH
+          CPU_ARCH=$(./init/eessi_archdetect.sh cpupath)
+          export EESSI_ARCHDETECT_OPTIONS=dummy/cpu:$CPU_ARCH:dummy1/cpu1
+          if module load EESSI/${{matrix.EESSI_VERSION}}; then
+              if [[ $CPU_ARCH == "${EESSI_SOFTWARE_SUBDIR}" ]]; then
```
trz42, line 34: Is it intended that this would fail on an architecture like `zen4` where `eessi_archdetect.sh cpupath` would return `x86_64/amd/zen4` while `EESSI_SOFTWARE_SUBDIR` would be set to `x86_64/amd/zen3` (current redirection implemented in `archdetect_cpu` function in the lua file in this PR). For other architectures that are not present in EESSI there might be a difference too (think of `zen5`, `sapphire rapids`, `neoverse_v2`, `a64fx`, ...).

It makes me wonder a bit what this test actually tests? It looks like it tests if a software stack for the _architecture_ that `eessi_archdetect.sh cpupath` returns is present on CVMFS.

Minor issue: use (double) quotes and curly brackets around all variables.

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,75 @@
...
+          result=$(grep "Software" $outfile)
+          if [[ "$result" == *"Software directory check for the detected architecture failed"* ]]; then
+              echo "Test for picking up invalid path on \$archdetect_cpu PASSED"
+          else
+              echo "Test for picking up invalid path on \$archdetect_cpu FAILED" >&2
+              exit 1
+          fi
+          unset EESSI_ARCHDETECT_OPTIONS
+              
+      - name: Test for expected variables after loading EESSI module
```
trz42, line 59: To me this test looks more relevant as it aims to ensure that the EESSI environment settings after `module load EESSI/2023.06` are the same as to those after running `source /cvmfs/software.eessi.io/versions/2023.06/init/bash`

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,67 @@
...
+Description
+===========
+The European Environment for Scientific Software Installations (EESSI, pronounced as easy) is a collaboration between different European partners in HPC community.The goal of this project is to build a common stack of scientific software installations for HPC systems and beyond, including laptops, personal workstations and cloud infrastructure. 
+
+More information
+================
+ - URL: https://www.eessi.io/docs/
+]])
+whatis("Description: The European Environment for Scientific Software Installations (EESSI, pronounced as easy) is a collaboration between different European partners in HPC community. The goal of this project is to build a common stack of scientific software installations for HPC systems and beyond, including laptops, personal workstations and cloud infrastructure.")
+whatis("URL: https://www.eessi.io/docs/:")
```
trz42, line 11: Just curious, is the last `:` part of the URL or part of the `whatis`-syntax? Maybe it should be

```suggestion
whatis("URL: https://www.eessi.io/docs/")
```

trz42 (changes requested):
Nice work overall, particularly the module file and the last test. Not sure what the other tests ("with only one valid path" and "with invalid path") are supposed to test nor what they actually test.

One might also want to have a peek at differences in the Lmod configuration but meaningful tests may be more difficult to come up with.

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,75 @@
...
+          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+          cvmfs_http_proxy: DIRECT
+          cvmfs_repositories: software.eessi.io
+              
+      - name: Test for archdetect_cpu functionality with only one valid path
+        run: |
+          export MODULEPATH=/cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/modules/:$MODULEPATH
+          CPU_ARCH=$(./init/eessi_archdetect.sh cpupath)
+          export EESSI_ARCHDETECT_OPTIONS=dummy/cpu:$CPU_ARCH:dummy1/cpu1
+          if module load EESSI/${{matrix.EESSI_VERSION}}; then
```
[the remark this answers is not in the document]
TopRichard: I modified the script and added Lmod initialization(not needed after merging #668)

.github/workflows/tests_eessi_module.yml
TopRichard: Merged this test along with the value comparison test when sourcing the bash script.

MaKaNu:
Could it be that:

```yaml
- name: Check out software-layer repository
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1

      - name: Mount EESSI CernVM-FS pilot repository
        uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
        with:
          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
          cvmfs_http_proxy: DIRECT
          cvmfs_repositories: software.eessi.io
```
Loads the release version of eessi and so your changes to the init system are not available?

TopRichard:
> Could it be that:
> 
> ```yaml
> - name: Check out software-layer repository
>         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
> 
>       - name: Mount EESSI CernVM-FS pilot repository
>         uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
>         with:
>           cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
>           cvmfs_http_proxy: DIRECT
>           cvmfs_repositories: software.eessi.io
> ```
> 
> Loads the release version of eessi and so your changes to the init system are not available?

Yes exactly, so the module EESSI/2023.06 is not available, this should work out after merging the PR

.github/workflows/tests_eessi_module.yml
TopRichard: Merged adding dummy/cpu arch to the current test.

trz42:
Looks like the test is still failing. Maybe show the result file to verify if the desired string is included?

trz42 (commented):
Looks good. If it actually works as intended can only really be checked after the module file has been ingested.

What is needed now, is to "build" once for a single architecture and verify if the module file is included in the tarball, then ingest the tarball, rerun the CI ... if the test succeeds this PR can be merged.

TopRichard:

TopRichard:
@trz42 the module file is included in the tarball.

ocaisa:
> > Looks like the test is still failing. Maybe show the result file to verify if the desired string is included?
> 
> Scratch that. The module file to be loaded will only be added by this PR. Hence it will never be able to load it from CVMFS before the module file has been ingested.

You can't load it from CVMFS, but you can
```bash
module use init/modules
```
and perform the tests

trz42:
> > > Looks like the test is still failing. Maybe show the result file to verify if the desired string is included?
> > 
> > 
> > Scratch that. The module file to be loaded will only be added by this PR. Hence it will never be able to load it from CVMFS before the module file has been ingested.
> 
> You can't load it from CVMFS, but you can
> 
> ```shell
> module use init/modules
> ```
> 
> and perform the tests

Agreed. This sounds like the correct way to test the module file. If it succeeds we can ingest it. If it fails (now or a future changed version), we must not ingest it.

ocaisa:
Ok, I think this is pretty much there. One thing I wonder is whether we should be deploying this to a higher level directory? The module file is version specific already so there is no real requirement to deploy it under the versions directory, it could be deployed under
```shell
/cvmfs/software.eessi.io/init
```
and that would make it much easier to have multiple EESSI versions available at once

TopRichard:

install_scripts.sh
```diff
@@ -102,6 +102,16 @@ mc_files=(
...
 
+# Copy for init/modules/EESSI directory
+mc_files=(
+   2023.06.lua
+)
+if [ -n "$EESSI_CVMFS_REPO" ]; then
+    copy_files_by_list ${TOPDIR}/init/modules/EESSI ${EESSI_CVMFS_REPO}/modules/EESSI "${mc_files[@]}"
+else
+    echo "Error: EESSI_CVMFS_REPO is required for copying the module files." >&2
+fi
```
trz42, line 113: This adds a new dependency on an environment variable. I'd suggest to add a new argument to the script, say, `--cvmfs-repo` or `--cvmfs-repo-path`, which can then be used in a similar fashion to `--prefix`.

trz42 (dismissed):
I think, we should try to minimise relying on environment settings in `install_scripts.sh`

trz42:
> I think we can easily include it in the tarball, e.g. by adding another step to the create_tarball.sh script. However, fixing the ingestion is a bit more tricky: it currently ingest all tarballs with `/cvmfs/software.eessi.io/versions` as base dir (i.e. it assumes that the contents of any tarball should be added there), and it's not straightforward to change that. We would either have to generate all tarballs from a higher-level directory, or have some exception based on the contents of the tarball (e.g. if it finds files in the tarball that start with `modules/EESSI` or something).

For now, we could go on with the original location, say, `/cvmfs/software.eessi.io/versions/2023.06/init/modules/EESSI` or so. If we ever want to put it in another place, e.g., version independent, we later can create a symlink to that location.

install_scripts.sh
TopRichard: I will wait for a final decision on where to place the module and act accordingly

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,67 @@
...
+          moduleoutfile="moduleout.txt"
+          sourceoutfile="sourceout.txt"
+          module load EESSI/${{matrix.EESSI_VERSION}}
+          env | grep -E '^(EESSI_S|EESSI_C)' | sort > "${moduleoutfile}"
+          module unload EESSI/${{matrix.EESSI_VERSION}}
+          source /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/init/bash
+          env | grep -E '^(EESSI_S|EESSI_C)' | sort > "${sourceoutfile}"
+          cat "${moduleoutfile}"
+          cat "${sourceoutfile}"
+          if (diff "${moduleoutfile}" "${sourceoutfile}" > /dev/null); then
```
ocaisa, line 61: This is a great test for today, but is brittle once we (hopefully) transition to using the module file exclusively

ocaisa: We can cross that bridge when the test actually breaks

install_scripts.sh
ocaisa: Since this is not a trivial change, I think we go with the original approach for now. This only really becomes relevant when there are a few version options (and a symlink would solve that)

ocaisa (changes requested):
Aside from the prompt situation I think this is good to go, and we can deal with the prompt in a follow up PR

ocaisa:
@bedroge Do we need any trick here to ensure that we pick up the Lmod cache created by EESSI?

ocaisa:
@TopRichard Timing a `module avail` is a good way of making sure that our spider cache is being used and not being rebuilt. Anything over say 15 seconds should be a failure.

TopRichard:

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,67 @@
...
+setenv("EESSI_SITE_MODULEPATH", eessi_site_module_path)
+setenv("EESSI_SOFTWARE_SUBDIR", eessi_software_subdir)
+setenv("EESSI_PREFIX", eessi_prefix)
+setenv("EESSI_EPREFIX", eessi_eprefix)
+prepend_path("PATH", pathJoin(eessi_eprefix, "bin"))
+prepend_path("PATH", pathJoin(eessi_eprefix, "usr/bin"))
+setenv("EESSI_SOFTWARE_PATH", eessi_software_path)
+setenv("EESSI_MODULEPATH", eessi_module_path)
+prepend_path("MODULEPATH", eessi_module_path)
+prepend_path("MODULEPATH", eessi_site_module_path)
```
ocaisa, line 62: Each one of these needs to be guarded so they are not included in a spider search (or at the very least the first one does), and `LMOD_RC` needs to be updated to find their spider caches

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
+        with:
+          cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+          cvmfs_http_proxy: DIRECT
+          cvmfs_repositories: software.eessi.io
+
+      - name: Test for making sure spider cache is being used and not being rebuilt
+        run: |
+          . /cvmfs/software.eessi.io/versions/${{matrix.EESSI_VERSION}}/compat/linux/$(uname -m)/usr/share/Lmod/init/bash  # Initialise Lmod
+          export MODULEPATH=init/modules
```
ocaisa, line 31:
```suggestion
          export MODULEPATH=init/modules
          module load EESSI/${{matrix.EESSI_VERSION}}
```

.github/workflows/tests_eessi_module.yml
```diff
@@ -0,0 +1,78 @@
...
+          if timeout 10s bash -c "LMOD_PAGER=none module --terse avail"; then
+              echo "EESSI spider cache is being used"
+          else
+              echo "EESSI spider cache is being rebuilt" >&2
+              exit 1
+          fi
```
ocaisa, line 37:
```suggestion
          fi
          module purge
          unset MODULEPATH
```

bedroge:
> @bedroge Do we need any trick here to ensure that we pick up the Lmod cache created by EESSI (and it doesn't get rebuilt by spider)? I see you have a few clever tricks in [#68 (comment)](https://github.com/EESSI/software-layer/issues/68#issuecomment-2173437988) :
> 
> ```lua
> if ( mode() ~= "spider" ) then
>   prepend_path("MODULEPATH", pathJoin(eessi_software_path, eessi_module_subdir, classes[i]))
> end
> ```
> 
> ```lua
> -- Set path to the Lmod cache of the the EESSI stack.
> prepend_path("LMOD_RC", pathJoin(eessi_software_path, "/.lmod/lmodrc.lua"))
> ```

We indeed set `LMOD_RC` in our EESSI modulefile, but it looks like setting `LMOD_CONFIG_DIR` to a directory that contains a file `lmodrc.lua` should also work, see https://lmod.readthedocs.io/en/latest/145_properties.html#the-properties-file-lmodrc-lua.

Tested it on my laptop:
```
$ export LMOD_CONFIG_DIR=/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/.lmod/
$ module --config
Active RC file(s):
------------------
/usr/share/lmod/lmod/libexec/../init/lmodrc.lua
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/.lmod/lmodrc.lua


Cache Directory                                                                            Time Stamp File
---------------                                                                            ---------------
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/.lmod/cache  /cvmfs/software.eessi.io/versions/2023.06/software/linux/
x86_64/intel/haswell/.lmod/cache/timestamp
```

init/modules/EESSI/2023.06.lua
bedroge: Unless something is different here because of not setting `LMOD_RC` (but I wouldn't expect so), this should indeed be guarded as described here: https://lmod.readthedocs.io/en/latest/350_community.html

Otherwise Lmod is going to recursively search through this path to find all modulefiles in the directory that you're adding to $MODULEPATH, while it doesn't have to, since the EESSI-provided cache already contains all of those.

ocaisa:
I'm not sure we _should_ be setting `LMOD_CONFIG_DIR`, I think that may be too intrusive if people are using their own Lmod

bedroge:
> I'm not sure we _should_ be setting `LMOD_CONFIG_DIR`, I think that may be too intrusive if people are using their own Lmod

True. I'm not sure if it can contain multiple paths (then we could prepend to it), but that's definitely possible with LMOD_RC.

MaKaNu:
> I'm not sure we _should_ be setting `LMOD_CONFIG_DIR`, I think that may be too intrusive if people are using their own Lmod

If prepending is an Option, is this part of the module or do we need this in the init scripts?

TopRichard:
> > I'm not sure we _should_ be setting `LMOD_CONFIG_DIR`, I think that may be too intrusive if people are using their own Lmod
> 
> True. I'm not sure if it can contain multiple paths (then we could prepend to it), but that's definitely possible with LMOD_RC.

`EESSI` `LMOD_CONFIG_DIR` is set when we source the bash script no ?

ocaisa:
The difference now is that we are allowing the case where people are using their own Lmod, we should only make the minimal changes to the environment we need to gets things working as we expect. That means using our spider caches and setting up our hooks. I think @casparvl has a good overview of this somewhere in a PR

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,70 @@
...
+setenv("EESSI_OS_TYPE", eessi_os_type)
+function archdetect_cpu()
+    local script = pathJoin(eessi_prefix, 'init', 'lmod_eessi_archdetect_wrapper.sh')
+    if not os.getenv("EESSI_ARCHDETECT_OPTIONS") then
+        if convertToCanonical(LmodVersion()) < convertToCanonical("8.6") then
+            LmodError("Loading this modulefile requires using Lmod version >= 8.6, but you can export EESSI_ARCHDETECT_OPTIONS to the available cpu architecture in the form of: x86_64/intel/haswell:x86_64/generic or aarch64/neoverse_v1:aarch64/generic")
+        end
+        source_sh("bash", script)
+    end
+local archdetect_options = os.getenv("EESSI_ARCHDETECT_OPTIONS") or ""
```
ocaisa, line 28:
```suggestion
    local archdetect_options = os.getenv("EESSI_ARCHDETECT_OPTIONS") or ""
```

init/modules/EESSI/2023.06.lua
```diff
@@ -0,0 +1,70 @@
...
+setenv("EESSI_SOFTWARE_SUBDIR", eessi_software_subdir)
+setenv("EESSI_PREFIX", eessi_prefix)
+setenv("EESSI_EPREFIX", eessi_eprefix)
+prepend_path("PATH", pathJoin(eessi_eprefix, "bin"))
+prepend_path("PATH", pathJoin(eessi_eprefix, "usr/bin"))
+setenv("EESSI_SOFTWARE_PATH", eessi_software_path)
+setenv("EESSI_MODULEPATH", eessi_module_path)
+if ( mode() ~= "spider" ) then
+    prepend_path("MODULEPATH", eessi_module_path)
+    prepend_path("MODULEPATH", eessi_site_module_path)
```
ocaisa, line 64: Probably need to move this out, we don't actually know if they have generated a spider cache for these modules. Maybe protect this with an environment variable (if the variable is set, don't spider, otherwise do it normally)?

ocaisa (changes requested):

init/modules/EESSI/2023.06.lua
ocaisa: I think we can deal with `PS1` in a follow up PR, let' s get this merged so people can try it out

ocaisa (approved):
LGTM, thanks for all the effort!

ocaisa:

bedroge:
This has been ingested, so I'll merge the PR.
https://github.com/EESSI/filesystem-layer/pull/198 adds a symlink to this new modulefile in a top-level dir `init/modules/EESSI`.

PR #668, opened 2024-08-12
Add new init scripts for new initialization module

MaKaNu:
This is a followup PR for https://github.com/EESSI/software-layer/pull/667

After successful merge of #667 I have some test cases prepared, to test against the different shell.

For the moment, I struggle with an implementation for csh. It only seems to happen when I try to load the module inside csh. Error response of csh is not the brightest I've seen.


init/bash
```diff
@@ -1,41 +1,9 @@
...
-fi
+# Choose an EESSI version
+EESSI_VERSION="${EESSI_VERSION:-2023.06}"
+# Automatically load the module for EESSI version
+export LMOD_SYSTEM_DEFAULT_MODULES="${LMOD_SYSTEM_DEFAULT_MODULES:-EESSI/${EESSI_VERSION}}"
+# Path to top-level module tree
+export MODULEPATH=/cvmfs/software.eessi.io/versions/"$EESSI_VERSION"/init/modules
+. /cvmfs/software.eessi.io/versions/"$EESSI_VERSION"/compat/linux/$(uname -m)/usr/share/Lmod/init/bash
+
+module load "$LMOD_SYSTEM_DEFAULT_MODULES"
```
ocaisa, line 9: The module load should not be necessary, they should be loaded when Lmod is initialised

ocaisa: I'm not 100% though...https://lmod.readthedocs.io/en/latest/070_standard_modules.html#providing-a-standard-set-of-modules-for-all-users

MaKaNu:
Ok it expanded the PS1 inside the git commit message :smiling_face_with_tear:

init/bash
ocaisa: Was there already an Lmod loaded when you `source`d it?

`module load $LMOD_SYSTEM_DEFAULT_MODULES` is a little brittle as this won't work if there is more than module in the variable (I think). It's probably best to take the official route as per the Lmod docs (shell dependent):
```bash
if [ -z "$__Init_Default_Modules" ]; then
   export __Init_Default_Modules=1;

   ## ability to predefine elsewhere the default list
   LMOD_SYSTEM_DEFAULT_MODULES=${LMOD_SYSTEM_DEFAULT_MODULES:-"StdEnv"}
   export LMOD_SYSTEM_DEFAULT_MODULES
   module --initial_load --no_redirect restore
else
   module refresh
fi
```

MaKaNu: > Was there already an Lmod loaded when you `source`d it?

No, Lmod was not loaded at that moment. Only loaded with the script.

MaKaNu:
>module load $LMOD_SYSTEM_DEFAULT_MODULES is a little brittle as this won't work if there is more than module in the variable (I think). It's probably best to take the official route as per the Lmod docs (shell dependent):

Implemented in last commit. Works as expected, with exclusion for csh.

MaKaNu:
Pushed tests are expected to fail for now (getting interesting after merge of #667)

init/bash
```diff
@@ -1,41 +1,17 @@
...
-    #show_msg ""
-    #show_msg "*** Known problems in the ${EESSI_VERSION} software stack ***"
-    #show_msg ""
-    #show_msg "1) ..."
-    #show_msg ""
-    #show_msg ""
-
-    echo "Environment set up to use EESSI (${EESSI_VERSION}), have fun!"
-
+# Choose an EESSI version
```
ocaisa, line 1: Changing the behaviour of the original initialisation is too disruptive since many are sure to rely on it (I know I do in my EESSI GitHub Action). Perhaps all the options for Lmod initialisation can be moved to a subdirectory (e.g., `init/lmod/bash`)

init/bash
```diff
@@ -1,41 +1,17 @@
...
+# Path to top-level module tree
+export MODULEPATH=/cvmfs/software.eessi.io/versions/"$EESSI_VERSION"/init/modules
+. /cvmfs/software.eessi.io/versions/"$EESSI_VERSION"/compat/linux/$(uname -m)/usr/share/Lmod/init/bash
+
+export PS1
+if [ -z "$__Init_Default_Modules" ]; then
+   export __Init_Default_Modules=1;
+
+   ## ability to predefine elsewhere the default list
+   LMOD_SYSTEM_DEFAULT_MODULES=${LMOD_SYSTEM_DEFAULT_MODULES:-"StdEnv"}
```
ocaisa, line 12:
```suggestion
   LMOD_SYSTEM_DEFAULT_MODULES=${LMOD_SYSTEM_DEFAULT_MODULES:-"EESSI"}
```
You'll need this everywhere

ocaisa: Some sites also force full name matching so you probably can't leave it versionless

ocaisa:
We never included PS1 in #667 so your tests will fail for that currently. If you are keen to have that, it could be part of this PR or a follow-up after this is merged (this PR has higher priority in my opinion)

MaKaNu:
> We never included PS1 in #667 so your tests will fail for that currently. If you are keen to have that, it could be part of this PR or a follow-up after this is merged (this PR has higher priority in my opinion)

Yes I was following the complete process and agree. The commits are reverted and can easily be reactivated later.

init/bash
MaKaNu: This is now implemented with Version number. It doesn't matter to me but like to understand the difference why `StdEnv.lua` is not the correct solution.

init/bash
MaKaNu: Done with fcf4937

MaKaNu:
I still struggle with correctly testing the `csh`. It seems the script is working correctly, but redirecting is a pain and I rely on redirecting in the tests a lot.

.github/workflows/scripts/test_init_scripts.sh
```diff
@@ -0,0 +1,43 @@
...
+for shell in ${SHELLS[@]}; do
+	echo = | awk 'NF += (OFS = $_) + 100'
+	echo  RUNNING TESTS FOR SHELL: $shell
+	echo = | awk 'NF += (OFS = $_) + 100'
+
+	# TEST 1: Source Script and check Module Output
+	assert "$shell -c 'source init/lmod/$shell' 2>&1 " "EESSI/$EESSI_VERSION loaded successfully"
+	# TEST 2: Check if module overviews first section is the loaded EESSI module
+	MODULE_SECTIONS=($($shell -c "source init/lmod/$shell 2>/dev/null; module ov 2>&1 | grep -e '---'"))
+	assert "echo ${MODULE_SECTIONS[1]}" "/cvmfs/software.eessi.io/versions/$EESSI_VERSION/software/linux/x86_64/intel/haswell/modules/all"
```
ocaisa, line 25: You've got no guarantees what architecture you will end up with so better to override the architecture to give you something predictable.

ocaisa:
```
export EESSI_ARCHDETECT_OPTIONS="x86_64/generic"
```

.github/workflows/tests_init_module.yml
```diff
@@ -0,0 +1,43 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        EESSI_VERSION:
+          - 2023.06
+        EESSI_SOFTWARE_SUBDIR_OVERRIDE:
+          - x86_64/haswell
```
ocaisa, line 19:
```suggestion
          - x86_64/intel/haswell
```
Ah you have it here

init/lmod/csh
```diff
@@ -0,0 +1,17 @@
...
+if (! $?__Init_Default_Modules ) then
+	setenv __Init_Default_Modules 1;
+
+	## ability to predefine elsewhere the default list
+	if (! $?LMOD_SYSTEM_DEFAULT_MODULES) then; setenv LMOD_SYSTEM_DEFAULT_MODULES "EESSI/$EESSI_VERSION"; endif
+	module --initial_load --no_redirect restore
+else
+	module refresh
+endif
+#module load "$LMOD_SYSTEM_DEFAULT_MODULES"
```
ocaisa, line 16:
```suggestion
```

.github/workflows/scripts/test_init_scripts.sh
```diff
@@ -0,0 +1,48 @@
...
+	PATTERN="/cvmfs/software\.eessi\.io/versions/$EESSI_VERSION/software/linux/x86_64/(intel/haswell|amd/zen3)/software/Python/3\.10\.8-GCCcore-12\.2\.0/bin/python"
+	echo "$PYTHON_PATH" | grep -E "$PATTERN"
+	assert_raises 'echo "$PYTHON_PATH" | grep -E "$PATTERN"'
+	
+	#End Test Suite
+	assert_end "source_eessi_$shell"
+done
+
+### EXCEPTION FOR CSH ###
+echo -e "\033[33mWe don't now how to test csh correctly, PRs are Welcome.\033[0m"
```
ocaisa, line 44: It would be good to add the exception to your loop, check if `$shell` is `csh` and if so just do the echo. Then we can also add it in workflow even if it doesn't really do anything

init/bash
ocaisa: There is no existing `StdEnv.lua` when Lmod is configured (we don't ship one)

ocaisa (dismissed):
LGTM

ocaisa:

ocaisa:
You need something similar to https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/install_scripts.sh#L94-L97 so that the new scripts are actually deployed

ocaisa:

MaKaNu:

ocaisa (approved):
LGTM, thanks for steering this over the line!

ocaisa:
Staging PR merged, this is now in the wild!

PR #674, opened 2024-08-19
use upper directory of former 'rw' session in 'ro' mode

trz42:
This PR is essentially redoing #550. Original description below. It is also doing a little code polishing replacing <kbd>TAB</kbd> with spaces.
___
Use case for this is that we want to test software additions to a CernVM-FS repository. Typically we would first run `eessi_container.sh` with `--access rw` to make the repository writable with an overlayfs. The changes are stored in some `overlay-upper` directory (under some `/tmp` dir or in a tarball of that `/tmp` directory). When we want to test the additions later we can run `eessi_container.sh` with `--resume TMPDIR_OR_TMPTARBALL`. However, until now this only worked when using `--access rw` which would make the repository again writable, which is likely not what is wanted in a test (because in actual use `/cvmfs/*` mounts are read-only). In fact, the `test-suite.sh` run after a build job has finished, uses `--access ro` (read-only), but this access mode did not recognise changes in an `overlay-upper` directory.

This PR merges an existing `overlay-upper` (given via `--resume ...`) on top of a CernVM-FS repository. Because it only uses the `lowerdir` option of `fuse-overlayfs` the merged directory is mounted read-only.
___

Because the mounting of directories under `/cvmfs` is done late in the `eessi_container.sh` script, this should also work with the changes introduced in #618, e.g., one should be able to run something like
```bash
eessi_container.sh --repository software.eessi.io,access=ro --repository dev.eessi.io,access=rw --storage=/tmp
```
which reports about a temporary location used such as `/tmp/eessi.abc123defg` and then
```bash
eessi_container.sh --repository software.eessi.io,access=ro --repository dev.eessi.io,access=ro --resume /tmp/eessi.abc123defg
```

### DIFF
--- a/eessi_container.sh
+++ b/eessi_container.sh
@@ -254,7 +254,7 @@ if [[ ${LIST_REPOS} -eq 1 ]]; then
             default_label=", default"
         else
             default_label=""
-	fi
+        fi
         echo "    ${cvmfs_repo} [CVMFS config repo${default_label}]"
     done
     for cfg_repo in "${!cfg_cvmfs_repos[@]}"
@@ -323,7 +323,7 @@ do
     if [[ ! -n "${eessi_cvmfs_repos[${cvmfs_repo_name}]}" ]] ; then
         [[ ${VERBOSE} -eq 1 ]] && echo "repo '${cvmfs_repo_name}' is not an EESSI CVMFS repository..."
         # cvmfs_repo_name is actually a repository ID, use that to obtain
-	#   the actual name from the EESSI_REPOS_CFG_FILE
+        #   the actual name from the EESSI_REPOS_CFG_FILE
         cfg_repo_id=${cvmfs_repo_name}
         cvmfs_repo_name=$(cfg_get_value ${cfg_repo_id} "repo_name")
     fi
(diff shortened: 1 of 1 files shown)


eessi_container.sh
```diff
@@ -736,15 +736,46 @@ do
...
+            # the left-most directory given for the lowerdir argument is put on top,
+            #   and with no upperdir=... the whole overlayfs is made available read-only
+            EESSI_READONLY_OVERLAY+=" -o lowerdir=${TMP_IN_CONTAINER}/${cvmfs_repo_name}/overlay-upper:/cvmfs_ro/${cvmfs_repo_name}"
+            EESSI_READONLY_OVERLAY+=" /cvmfs/${cvmfs_repo_name}"
+            export EESSI_READONLY_OVERLAY
+
+            EESSI_FUSE_MOUNTS+=("--fusemount" "${EESSI_READONLY_OVERLAY}")
+            export EESSI_FUSE_MOUNTS
+        else
+            # basic "ro" access that doesn't require any fuseoverlay-fs
```
casparvl, line 768: Maybe also add an explicit echo here, just so we easily see in which branch we went.

casparvl (changes requested):
Once the two explicit `echo`'s are added to make it more transparent into which branch we went, I'd say this is good to go.

https://github.com/EESSI/software-layer/pull/676#issuecomment-2298319163 proves that this enables the test step of running in `ro` mode _with_ the previous `overlay-upper` mounted as lowerdir (i.e. read-only). It shows that the `patchelf` we just built in the build step is available in the test step. At the same time, https://github.com/EESSI/software-layer/pull/677 proves that without this PR (#674 ), the `patchelf` we just built in the build step is _not_ available in the test step.

In summary: this PR does exactly what it says on the label :)

PR #675, opened 2024-08-19
[WIP] update licenses

MartinsNadia:
Open the new one to fix the origin branch


licenses/check_licenses.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  license_update:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
ocaisa, line 20: Should use specific version here, see other workflows

licenses/check_licenses.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          python update_licenses.py --source=pypi TensorFlow
+          python update_licenses.py --source=github:easybuilders/easybuild EasyBuild
+          if [ -f license_update.patch ] && [ -s license_update.patch ]; then
+            PATCH_CONTENT=$(cat license_update.patch)
+            echo "patch=$PATCH_CONTENT" >> $GITHUB_OUTPUT
+          fi 
+
+      - name: Create a PR (if changes detected)
+        uses: peter-evans/create-pull-request@v5
```
ocaisa, line 36: Also needs specific commit

licenses/check_licenses.yml
```diff
@@ -0,0 +1,57 @@
...
+        if: steps.check_licenses.outputs.patch != ''
+        with: 
+        commit-message: "Auto PR: Update licenses"
+        title: "Auto PR: Update licenses"
+         body: ${{ steps.check_licenses.outputs.patch }}
+         branch: main #fork branch
```
ocaisa, line 42: This is not the correct branch for EESSI

ocaisa (commented):
Just had an initial look, perhaps you can add the current output of the script in a gist and link to this issue?

.github/workflows/check_licenses.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          python-version: '3.9' 
+
+      - name: Run license script and generate patch
+        id: check_licenses
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python update_licenses.py --source=pypi TensorFlow
+          python update_licenses.py --source=github:easybuilders/easybuild EasyBuild
```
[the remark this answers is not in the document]
ocaisa: This could also help you decide where to look for licenses (since you have access to `source_urls`)

ocaisa (changes requested):

hvelab:
spare thoughts right now: 
- i kept the retrieval date because the original script did so, but i dont see much sense in keeping it, I would keep the license url instead 
- there are a lot of not found but i expect them to be less of them as right now some source_urls are still formatted with EB syntax but it my priority to fix it tomorrow
- need to fix the "needs manual action" part as its case sensitive right now with the "Other" licenses

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+        "Retrieved From": "not found"
+    },
+    "ASE/3.22.1-gfbf-2022b": {
+        "License": "not found",
+        "Needs Manual Check": true,
+        "Retrieved From": "not found"
+    },
+    "ATK/2.38.0-GCCcore-13.2.0": {
+        "License": [
+            "Other"
```
ocaisa, line 19: How did this end up with `Other`? I see it is GPL v2 at https://gitlab.gnome.org/Archive/atk

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+    "Armadillo/12.8.0-foss-2023b": {
+        "License": [
+            "Apache-2.0"
+        ],
+        "Needs Manual Check": false,
+        "Retrieved From": "https://arma.sourceforge.net/"
+    },
+    "Arrow/16.1.0-gfbf-2023b": {
+        "License": [
+            "Other"
```
ocaisa, line 47: License for this is https://github.com/apache/arrow/blob/main/LICENSE.txt ("Apache-2.0")

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+        "Needs Manual Check": true,
+        "Retrieved From": "not found"
+    },
+    "BamTools/2.5.2-GCC-12.2.0": {
+        "License": "not found",
+        "Needs Manual Check": true,
+        "Retrieved From": "not found"
+    },
+    "Bazel/6.3.1-GCCcore-12.3.0": {
+        "License": "not found",
```
ocaisa, line 80: Apache 2.0 from https://github.com/bazelbuild/bazel/blob/master/LICENSE

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+        "Needs Manual Check": true,
+        "Retrieved From": "not found"
+    },
+    "BeautifulSoup/4.12.2-GCCcore-12.3.0": {
+        "License": [
+            "Other"
```
ocaisa, line 86: MIT from https://pypi.org/project/beautifulsoup4/

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+    },
+    "FFTW.MPI/3.3.10-gompi-2023b": {
+        "License": [
+            "Other"
+        ],
+        "Needs Manual Check": false,
+        "Retrieved From": "https://www.fftw.org"
+    },
+    "FFmpeg/6.0-GCCcore-13.2.0": {
+        "License": "not found",
```
ocaisa, line 368: https://www.ffmpeg.org/legal.html

licenses/licenses.json
```diff
@@ -1,10 +1,2751 @@
...
+    "Abseil/20240116.1-GCCcore-13.2.0": {
+        "License": [
+            "Apache-2.0"
+        ],
+        "Needs Manual Check": false,
+        "Retrieved From": "https://abseil.io/"
+    },
+    "Archive-Zip/1.68-GCCcore-12.2.0": {
+        "License": [
+            "Other"
```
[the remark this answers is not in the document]
hvelab: Agree, this is the plain response from the ecosyste.ms API: https://repos.ecosyste.ms/api/v1/repositories/lookup?url=https%3A%2F%2Fgithub.com%2Fredhotpenguin%2Fperl-Archive-Zip

Will work today on a "smarter" way to do this and to scrape other sources because there are much more not found licenses than found or other

hvelab:
Updates: 
- Fixed bug and now finds more licenses
- For the "Other" and "not found", does an scraping to go find the LICENSE file -> we need to find a way to retrieve the spdx from there 
- This needs to be improved for the packages with "Others" 
- For the totally "not found", shows source_urls and homepages, seems that most of them need to be sanitized or do the scraping from there, add more keywords

hvelab:
finished doing manual checks (while watching some random movie on the TV), we would just need permission for PGPLOT (https://sites.astro.caltech.edu/~tjp/pgplot/#copyright), and for CUDA im not sure if the propietary license allows us to redistribute it 
the scraper (aside from update_licenses.json) is good for an initial populate but to cover every single case is very difficult and required much more manual checks than expected, at least is done for now, for all software available in EESSI excluding extensions, which is my next focus, such as enforcing specifying the license from EB files

could try from a hook and if it fails then request the manual action? again, it's very difficult to scrape it from webpages and sometimes requires human interpretation

laraPPr:
@ocaisa should this be closed and moved to https://github.com/EESSI/software-layer-scripts?

ocaisa:
Licences check is now in CI (but still being tweaked #1492)

PR #678, opened 2024-08-20
Fix failing test step on AWS

casparvl:
Currently, the test step on AWS fails because we fail to get a memory limit from the cgroup. I'll add some more verbose output as a first step to debugging this.

### DIFF
--- a/test_suite.sh
+++ b/test_suite.sh
@@ -164,6 +164,11 @@ if [[ "${cpuinfo}" =~ (Core\(s\) per socket:[^0-9]*([0-9]+)) ]]; then
 else
     fatal_error "Failed to get the number of cores per socket for the current test hardware with lscpu."
 fi
+
+# The /sys inside the container is not the same as the /sys of the host
+# We want to extract the memory limit from the cgroup on the host (which is typically set by SLURM).
+# Thus, bot/test.sh bind-mounts the host's /sys/fs/cgroup into /hostsys/fs/cgroup
+# and that's the prefix we use to extract the memory limit from
 cgroup_v1_mem_limit="/hostsys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.limit_in_bytes"
 cgroup_v2_mem_limit="/hostsys/fs/cgroup/$(</proc/self/cpuset)/memory.max"
 if [ -f "$cgroup_v1_mem_limit" ]; then
--- a/test_suite.sh
+++ b/test_suite.sh
@@ -164,11 +164,11 @@ if [[ "${cpuinfo}" =~ (Core\(s\) per socket:[^0-9]*([0-9]+)) ]]; then
 else
     fatal_error "Failed to get the number of cores per socket for the current test hardware with lscpu."
 fi
-cgroup_v1_mem_limit="/sys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.limit_in_bytes"
-cgroup_v2_mem_limit="/sys/fs/cgroup/$(</proc/self/cpuset)/memory.max"
+cgroup_v1_mem_limit="/hostsys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.limit_in_bytes"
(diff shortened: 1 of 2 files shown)


casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
```
ERROR: Both files /sys/fs/cgroup/memory//slurm/uid_60008/job_16702/step_batch/memory.limit_in_bytes and /sys/fs/cgroup//slurm/uid_60008/job_16702/step_batch/memory.max couldn't be found. Failed to get the memory limit from the current cgroup
```
Interactively, I got the _correct_ output, from file `/sys/fs/cgroup/memory//slurm/uid_60014/job_16701/step_interactive/memory.limit_in_bytes`. Now, I don't know why `step_batch` doesn't seem to have the `memory.limit_in_bytes` file.

I'll try to manually submit some batch jobs on AWS to figure this out. It's tricky that we don't see this interactively...

My bet is, it is due to the replacement with `$(</proc/self/cpuset)`. This seems like a more robust way of doing things, but maybe it isn't - we can always revert that choice, at least for AWS.

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
The difference is that we are in a container... so we should use the info from the mounted directories of the host, not from the container's `/sys`  and `/proc`.

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen3

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen2

casparvl:
bot: build repo:eessi.io-2023.06-software arch:zen4

casparvl:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/haswell

test_suite.sh
```diff
@@ -164,11 +164,13 @@ if [[ "${cpuinfo}" =~ (Core\(s\) per socket:[^0-9]*([0-9]+)) ]]; then
 else
     fatal_error "Failed to get the number of cores per socket for the current test hardware with lscpu."
 fi
-cgroup_v1_mem_limit="/sys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.limit_in_bytes"
-cgroup_v2_mem_limit="/sys/fs/cgroup/$(</proc/self/cpuset)/memory.max"
+cgroup_v1_mem_limit="/hostsys/fs/cgroup/memory/$(</proc/self/cpuset)/memory.limit_in_bytes"
+cgroup_v2_mem_limit="/hostsys/fs/cgroup/$(</proc/self/cpuset)/memory.max"
```
trz42, line 168: This path probably makes only sense if run in a very specific environment (e.g., testing software built for EESSI). While this is fine, how about checking whether `/sys` or `/hostsys` is available and use that?

If there would be a comment that explains what `/hostsys` is and how it is made available, it might make debugging a little easier.

trz42 (changes requested):
Looks fine overall. Added a little question if the script could have made a little 'smarter' (using `/sys` or `/hostsys`) and a suggestion to explain a bit what this `/hostsys` is (and where it comes from).

test_suite.sh
casparvl: Yes, we bind-mount this additional path in `bot/test.sh`. You're absolutely right about the commenting part: I'll make that clear. 

Regarding a fallback on `/sys`, I'm not sure if we want to do that. If `/hostsys` isn't there, it means the bind-mount failed / was not executed. I'd probably prefer there to be a hard error, than a silent success here that maybe extracts the wrong amount of memory.

casparvl: Ok, added the description now

trz42 (approved):
Thanks for the update. Makes sense that we use `/hostsys` and don't fall back to `/sys`.

PR #690, opened 2024-09-02
list broken modules 

MartinsNadia:
Add script to list broken modules using `--module-only `

### DIFF
--- a/licenses/check_licenses.yml
+++ b/licenses/check_licenses.yml
@@ -1,0 +1,57 @@
+name: Check and update licenses
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  license_update:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.X' 
+
+      - name: Run license script and generate patch
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python update_licenses.py --source=pypi TensorFlow
+          python update_licenses.py --source=github:easybuilders/easybuild EasyBuild
+          if [ -f license_update.patch ] && [ -s license_update.patch ]; then
+            PATCH_CONTENT=$(cat license_update.patch)
+            echo "patch=$PATCH_CONTENT" >> $GITHUB_OUTPUT
+          fi 
+
+      - name: Create a PR (if changes detected)
+        uses: peter-evans/create-pull-request@v5
+        if: steps.check_licenses.outputs.patch != ''
+        with: 
+        commit-message: "Auto PR: Update licenses"
+        title: "Auto PR: Update licenses"
+         body: ${{ steps.check_licenses.outputs.patch }}
+         branch: main #fork branch
+         base: main #specify right brancg here
+
+      - name: Apply patch (if no PR created)
+        if: steps.create_pull_request.outputs.pull-request-number == '' && steps.check_licenses.outputs.patch != ''        
+        run: |
+          if [ -f license_update.patch ] && [ -s license_update.patch ]; then
+            git apply license_update.patch
+          else
+            echo "No changes to apply"
+          fi
+          git add licenses.json
+          git diff --cached --exit-code || git commit -m "Update licenses.json"
+          git push
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


modules/module_check.sh
```diff
@@ -0,0 +1,79 @@
...
+locks_dir="/tmp/$USER/EESSI/locks"
+
+# Log file to record broken modules 
+broken_modules_log="broken_modules.log"
+> $broken_modules_log
+
+declare -A checked_modules
+
+# Locate all .eb files within the base dir
+easyconfig_files=$(find $base_dir -name "*.eb")
```
ocaisa, line 35: This is quite a hammer depending on how this is used. It will run on the entire stack, which can take a long time as sanity checks are also run when using `--module-only`

ocaisa: It would be better if we can selectively do this for easyconfigs that are introduced by a PR.

One way to do this is to gather that information using `--missing` based on the easy stack files touched by a PR.

As for implementation, to me this would have to be done as part of the build script:
* Gather the easyconfigs to be installed using `--missing` on the easystack file touched by the PR (which is already known in the build script)
* Do the builds as currently implemented 
* Go through the missing easyconfig list with `--module-only` and the other necessary options, compare the two modules and make sure that they are identical

ocaisa (changes requested):

modules/module_check.sh
ocaisa: Specifically, I mean targeting the loop https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/EESSI-install-software.sh#L276-L295

ocaisa: So add a step before 279 to gather the easyconfig files that will be built, then add an `else` clause to L283 (i.e., when the build succeeds) which rebuilds the module and checks the module contents against the original

modules/module_check.sh
```diff
@@ -0,0 +1,79 @@
...
+            # Checks if the module has already been tested 
+            if [ -n "${checked_modules[$module_name]}" ]; then
+                echo "Module $module_name already checked. Skipping."
+                continue
+            fi
+
+            echo "Testing module: $module_name"
+
+            # Try loading the module 
+            if module --ignore_cache load $module_name 2>/dev/null; then
```
ocaisa, line 65: This is checking if the module can be loaded, but we probably want more than that, we want to check that the resulting environment (apart from Lmod variables) is the same as when the original module was loaded.

ocaisa: You can do this pretty easily enough on a module by module basis, just by changing the `MODULEPATH`:

- Load the original module, store the environment using `env` filtering out Lmod related content
- `module use `$module_install_dir`
- reload the module (making sure it comes from where you expect)
- Store the new environment (also filtering)
- Make sure they are identical

modules/module_check.sh
```diff
@@ -0,0 +1,79 @@
...
+        echo "EasyBuild command failed for $package_name. Skipping..."
+        echo "$package_name: EasyBuild command failed" >> $broken_modules_log
+        continue
+    fi
+
+    # Check the generated modules and iterate over the modules in the 'all' dir 
+    echo "Checking generated modules for $package_name..."
+    for module_category in $(ls $module_install_dir/all); do
+        for module_version in $(ls $module_install_dir/all/$module_category); do
+            module_name="$module_category/$module_version"
```
ocaisa, line 54:
```suggestion
    for module_name in $(ls $module_install_dir/all); do
        for module_version in $(ls $module_install_dir/all/$module_category); do
            module_name="$module_name/$module_version"
```

ocaisa, line 54:
```suggestion
    for module_software in $(ls $module_install_dir/all); do
        for module_version in $(ls $module_install_dir/all/$module_software); do
            module_name="$module_software/$module_version"
```

licenses/check_licenses.yml
```diff
@@ -0,0 +1,57 @@
+name: Check and update licenses
```
ocaisa, line 1: This is still mixing up two PRs, the one for licenses and the one for checking `--module-only`

modules/module_check.sh
```diff
@@ -0,0 +1,203 @@
+#!/bin/bash
```
ocaisa, line 1: Does this work? Have you got sample output?

ocaisa (changes requested):

laraPPr:
@hvelab Is this something you can pick up on again. It should probably also be moved to https://github.com/EESSI/software-layer-scripts.

ocaisa:
@MartinsNadia Rather than in CI, I think the best place for a check like this is in the build script right after a successful build. That would be in https://github.com/EESSI/software-layer-scripts

PR #700, opened 2024-09-12
install_cuda_host_injections: Don't fail if temp dir exists or parent is missing

Flamefire:
Use `mkdir -p` such that  `--temp-dir /tmp/$USER/EESSI` works out of the box. It also avoids failures if the user got "ERROR: Could not create directory /tmp/root/EESSI/temp" and did `mkdir -p /tmp/root/EESSI/temp` manually.

PR #701, opened 2024-09-12
install_cuda_host_injections: Fix CUDA-EC-not-found error message

Flamefire:
The search query is a regexp not a glob, so add the missing dot. Also add the missing "of CUDA" in the message


ocaisa:
@Flamefire Can you merge this correction with your other PR, we have to build/deploy this for all architectures which has a good bit or overhead for us.

Flamefire:
Done -> Both are in #700

PR #705, opened 2024-09-17
{2023.06}[2023a,zen4] LAMMPS 2Aug2023 for zen4

laraPPr:
Building LAMMPS 2Aug2023 requires https://github.com/easybuilders/easybuild-easyblocks/pull/3336/files which is part of EasyBuild 4.9.3

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -26,6 +26,8 @@ CPU_TARGET_NEOVERSE_V1 = 'aarch64/neoverse_v1'
 CPU_TARGET_AARCH64_GENERIC = 'aarch64/generic'
 CPU_TARGET_A64FX = 'aarch64/a64fx'
 
+CPU_TARGET_ZEN4 = 'x86_64/amd/zen4'
+
 EESSI_RPATH_OVERRIDE_ATTR = 'orig_rpath_override_dirs'
 
 SYSTEM = EASYCONFIG_CONSTANTS['SYSTEM'][0]
@@ -514,6 +516,23 @@ def pre_configure_hook_wrf_aarch64(self, *args, **kwargs):
         raise EasyBuildError("WRF-specific hook triggered for non-WRF easyconfig?!")
 
 
+def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
+    """
+    pre-configure hook for LAMMPS:
+    - set kokkos_arch on x86_64/amd/zen4
+    """
+
+    cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
+    if self.name == 'LAMMPS':
+        if self.version == '2Aug2023':
+            if get_cpu_architecture() == X86_64:
+                if cpu_target == CPU_TARGET_ZEN4:
+                    # There is no support for ZEN4 in LAMMPS yet so falling back to ZEN3
+                    self.cfg['kokkos'] = 'ZEN3'
+    else:
+        raise EasyBuildError("LAMMPS-specific hook triggered for non-LAMMPS easyconfig?!")
+
+
 def pre_test_hook(self,*args, **kwargs):
     """Main pre-test hook: trigger custom functions based on software name."""
     if self.name in PRE_TEST_HOOKS:
@@ -783,6 +802,7 @@ PRE_CONFIGURE_HOOKS = {
     'MetaBAT': pre_configure_hook_metabat_filtered_zlib_dep,
     'OpenBLAS': pre_configure_hook_openblas_optarch_generic,
     'WRF': pre_configure_hook_wrf_aarch64,
+    'LAMMPS': pre_configure_hook_LAMMPS_zen4,
 }
 
 PRE_TEST_HOOKS = {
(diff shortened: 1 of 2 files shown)


laraPPr:
bot: build repo:eessi.io-2023.06-software arch:zen4

laraPPr:
@boegel You can also try to build for A64FX has the same amount of missing dependencies as the zen4 stack

boegel:
> @boegel You can also try to build for A64FX has the same amount of missing dependencies as the zen4 stack

I can, but if I do this for this PR, it will first try to install everything else already included in `easystacks/software.eessi.io/2023.06/zen4/*.yml`, so we should open a dedicated PR for `easystacks/software.eessi.io/2023.06/a64fx/*.yml`

laraPPr:
> > @boegel You can also try to build for A64FX has the same amount of missing dependencies as the zen4 stack
> 
> I can, but if I do this for this PR, it will first try to install everything else already included in `easystacks/software.eessi.io/2023.06/zen4/*.yml`, so we should open a dedicated PR for `easystacks/software.eessi.io/2023.06/a64fx/*.yml`

Ah I'll open a seperate pr for `easystacks/software.eessi.io/2023.06/a64fx/*.yml`

laraPPr:
since I started a new stack for EasyBuild 4.9.3 it will only pick up the changes in that EasyStack right. So in this case it would only pick up a64fx. But for clarity it might be best that we keep them seperate to some extend.

laraPPr:
```
== 2024-09-17 12:01:03,066 run.py:689 DEBUG cmd "python -c 'from archspec.cpu import host; print(host())'" exited with exit code 0 and output:

x86_64_v4
```

The output should be zen4 not sure why this happened this did not happen when building at Gent with `zen4`

boegel:
> since I started a new stack for EasyBuild 4.9.3 it will only pick up the changes in that EasyStack right. So in this case it would only pick up a64fx. But for clarity it might be best that we keep them seperate to some extend.

That's true, I overlooked that.

boegel:
> ```
> == 2024-09-17 12:01:03,066 run.py:689 DEBUG cmd "python -c 'from archspec.cpu import host; print(host())'" exited with exit code 0 and output:
> 
> x86_64_v4
> ```
> 
> The output should be zen4 not sure why this happened this did not happen when building at Gent with `zen4`

This probably indicates that one of the CPU features expected by archspec for zen4 can't be found, so it "falls back" to `x86_64_v4`; see also https://github.com/archspec/archspec-json/blob/master/cpu/microarchitectures.json#L2140

It isn't the first time this happens in AWS or Azure, see for example https://github.com/archspec/archspec-json/issues/38

laraPPr:
So than set it in a hook or add this to the easyBlock?

ocaisa:
In this case it would need to be a hook, perhaps we should be doing this anyway given that we use archdetect rather than archspec.

You can use a parse hook that sets the correct value for `kokkos_arch`, with a table that maps between the environment variable `EESSI_SOFTWARE_SUBDIR` and the value Kokkos expects

laraPPr:
Yeah we did the same in the pilot so I can copy and update that

laraPPr:
bot: build repo:eessi.io-2023.06-software arch:zen4

eb_hooks.py
```diff
@@ -514,6 +516,23 @@ def pre_configure_hook_wrf_aarch64(self, *args, **kwargs):
...
+    - set kokkos_arch on x86_64/amd/zen4
+    """
+
+    cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
+    if self.name == 'LAMMPS':
+        if self.version == '2Aug2023':
+            if get_cpu_architecture() == X86_64:
+                if cpu_target == CPU_TARGET_ZEN4:
+                    # There is no support for ZEN4 in LAMMPS yet so falling back to ZEN3
+                    self.cfg['kokkos'] = 'ZEN3'
```
laraPPr, line 531:
```suggestion
                    self.cfg['kokkos_arch'] = 'ZEN3'
```

laraPPr:
Made a stupid mistake should be fixed now

laraPPr:
bot: build repo:eessi.io-2023.06-software arch:zen4

eb_hooks.py
```diff
@@ -514,6 +516,23 @@ def pre_configure_hook_wrf_aarch64(self, *args, **kwargs):
...
 
+def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
+    """
+    pre-configure hook for LAMMPS:
+    - set kokkos_arch on x86_64/amd/zen4
+    """
+
+    cpu_target = get_eessi_envvar('EESSI_SOFTWARE_SUBDIR')
+    if self.name == 'LAMMPS':
+        if self.version == '2Aug2023':
```
laraPPr, line 527:
```suggestion
        if self.version == '2Aug2023_update2':
```

laraPPr:
bot: build repo:eessi.io-2023.06-software arch:zen4

laraPPr:
Succes!!

boegel (approved):
lgtm

boegel:
Ah snap, I merged before doing the deploy 🤦 

Should still work though, deploy triggered...

PR #710, opened 2024-09-17
take into account accelerator target when configuring EasyBuild

boegel:


trz42:
We have to tweak the `bot/check-build.sh` so it recognises the modules/software under the different path. For example, see output in https://github.com/boegel/software-layer/pull/33#issuecomment-2356559701 where module files and installed files are shown under 'other'.

boegel:
@trz42 As discussed, this should be good to merge so we can start building CUDA software to be included in `software.eessi.io...

The improvements to `bot/check-build.sh` will be done in a follow-up PR

boegel:
Testing of this was done in https://github.com/boegel/software-layer/pull/33, see [last comment](https://github.com/boegel/software-layer/pull/33#issuecomment-2356559701) for success build of `beagle-lib-4.0.1-GCC-12.3.0-CUDA-12.1.1.eb`

trz42 (approved):
lgtm, thanks a lot @boegel !

PR #717, opened 2024-09-18
{2023.06}[foss/2023a] TensorFlow v2.15.1 w/ CUDA 12.1.1

casparvl:


casparvl:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

casparvl:
Let's see how this goes. Note that we need a proper `cuDNN` deployment that strips the necessary files first... So we will need to rebuild in any case.

boegel (changes requested):
@casparvl Please use easystack file under `/accel/nvidia`

boegel:
> Let's see how this goes. Note that we need a proper `cuDNN` deployment that strips the necessary files first... So we will need to rebuild in any case.

I've marked this a draft, we definitely don't want to deploy with full cuDNN installation

bedroge:
The build succeeded, but many tests failed due to:
```
ImportError: libnccl.so.2: cannot open shared object file: No such file or directory
```
This is already available in the CPU-only stack, so I'm not sure why it didn't pick up the library from that module.

trz42:
> The build succeeded, but many tests failed due to:
> 
> ```
> ImportError: libnccl.so.2: cannot open shared object file: No such file or directory
> ```
> 
> This is already available in the CPU-only stack, so I'm not sure why it didn't pick up the library from that module.

Just opened https://github.com/easybuilders/easybuild-easyblocks/pull/3497 which may fix the `libnccl.so.2` error.

laraPPr:
@casparvl Can you retarget this pr?

PR #720, opened 2024-09-18
{2023.06}[foss/2023a] CUDA 12.1.1 (rebuild)

casparvl:
requires:
* #735

PR #735, opened 2024-09-24
{2023.06}[foss/2023a] CUDA 12.1.1 (rebuild) + limit CUDA hook to EESSI installs only, and remove duplication when creating symlinks

ocaisa:
Also includes #720 to allow for testing

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -32,6 +32,9 @@ EESSI_RPATH_OVERRIDE_ATTR = 'orig_rpath_override_dirs'
 
 SYSTEM = EASYCONFIG_CONSTANTS['SYSTEM'][0]
 
+EESSI_INSTALLATION_REGEX = r"^/cvmfs/[^/]*.eessi.io/versions/"
+HOST_INJECTIONS_LOCATION = "/cvmfs/software.eessi.io/host_injections/"
+
 
 def get_eessi_envvar(eessi_envvar):
     """Get an EESSI environment variable from the environment"""
@@ -419,7 +422,7 @@ def pre_configure_hook_extrae(self, *args, **kwargs):
         # and problem due to use of which only pops up when running make ?!
         self.cfg.update(
             'prebuildopts',
-            "cp config/mpi-macros.m4 config/mpi-macros.m4.orig &&"
+            "cp config/mpi-macros.m4 config/mpi-macros.m4.orig && "
             "sed -i 's/`which /`command -v /g' config/mpi-macros.m4 && "
             )
     else:
@@ -698,9 +701,7 @@ def post_postproc_cuda(self, *args, **kwargs):
     """
 
     # We need to check if we are doing an EESSI-distributed installation
-    eessi_pattern = r"^/cvmfs/[^/]*.eessi.io/versions/"
-    host_injections_location = "/cvmfs/software.eessi.io/host_injections/"
-    eessi_installation = bool(re.search(eessi_pattern, self.installdir))
+    eessi_installation = bool(re.search(EESSI_INSTALLATION_REGEX, self.installdir))
 
     if self.name == 'CUDA' and eessi_installation:
         print_msg("Replacing files in CUDA installation that we can not ship with symlinks to host_injections...")
@@ -752,12 +753,12 @@ def post_postproc_cuda(self, *args, **kwargs):
                         # if it is not in the allowlist, delete the file and create a symlink to host_injections
 
                         # the host_injections path is under a fixed repo/location for CUDA
-                        host_inj_path = re.sub(eessi_pattern, host_injections_location, full_path)
+                        host_inj_path = re.sub(EESSI_INSTALLATION_REGEX, HOST_INJECTIONS_LOCATION, full_path)
                         # CUDA itself doesn't care about compute capability so remove this duplication from
-                        # under host_injections
+                        # under host_injections (symlink to a single CUDA installation for all compute
+                        # capabilities)
                         accel_subdir = os.getenv("EESSI_ACCELERATOR_TARGET")
                         if accel_subdir:
-                            host_inj_path = host_inj_path.replace('/accel/%s' % accel_subdir, '')
                             host_inj_path = host_inj_path.replace("/accel/%s" % accel_subdir, '')
                         # make sure source and target of symlink are not the same
                         if full_path == host_inj_path:
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -684,7 +684,8 @@ def post_sanitycheck_cuda(self, *args, **kwargs):
     Remove files from CUDA installation that we are not allowed to ship,
     and replace them with a symlink to a corresponding installation under host_injections.
     """
-    if self.name == 'CUDA':
+    # Make sure we only do this for CUDA and only if we are doing a CVMFS installation
+    if self.name == 'CUDA' and self.installdir.startswith('/cvmfs/software.eessi.io/versions'):
         print_msg("Replacing files in CUDA installation that we can not ship with symlinks to host_injections...")
 
         # read CUDA EULA, construct allowlist based on section 2.6 that specifies list of files that can be shipped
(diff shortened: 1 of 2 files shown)


ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

eb_hooks.py
```diff
@@ -684,7 +684,8 @@ def post_sanitycheck_cuda(self, *args, **kwargs):
     Remove files from CUDA installation that we are not allowed to ship,
     and replace them with a symlink to a corresponding installation under host_injections.
     """
-    if self.name == 'CUDA':
+    # Make sure we only do this for CUDA and only if we are doing a CVMFS installation
+    if self.name == 'CUDA' and self.installdir.startswith('/cvmfs/software.eessi.io/versions'):
```
ocaisa, line 688: Hmm, this might not be enough of a guard, if we do `--module-only` I think this would still trigger

ocaisa:
@casparvl I've checked this build, and the symlinks do indeed exist point to the location under `host_injections` without `accel/nvidia/cc80`.
The one problem is whether this would fully address the issue you saw with the EEESSI EasyBuild hooks when installing CUDA

eb_hooks.py
boegel: sanity check is also run when using `--module-only`

eb_hooks.py
```diff
@@ -733,6 +734,9 @@ def post_sanitycheck_cuda(self, *args, **kwargs):
                                        basename, full_path)
                         # if it is not in the allowlist, delete the file and create a symlink to host_injections
                         host_inj_path = full_path.replace('versions', 'host_injections')
+                        # CUDA itself doesn't care about compute capability so remove this duplication from
+                        # under host_injections
+                        host_inj_path = re.sub(r"accel/nvidia/cc\d+/", '', host_inj_path)
```
boegel, line 739: Oof, this is hardcoding the `accel/nvidia/cc*` pattern quite deep in our scripts...

Why don't we pick up on `$EESSI_ACCELERATOR_TARGET` here instead, just in case that changes?

boegel (changes requested):

eb_hooks.py
ocaisa: What's the value of `EESSI_ACCELERATOR_TARGET`?

ocaisa: Found it, I'd need `accel/$EESSI_ACCELERATOR_TARGET`

ocaisa: Implemented in https://github.com/EESSI/software-layer/pull/735/commits/087b7d7d42436b07c035b3433d03c9318fd9ec4c

eb_hooks.py
ocaisa: I guess I will just check the build options for `--sanity-check-only` and `--module-only`

ocaisa: Fixed in https://github.com/EESSI/software-layer/pull/735/commits/b8555d1ee64330a0a43df838f4d1ec58a916a4b5

ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

eb_hooks.py
boegel: If it should only happen when an actual install is being done, it belongs in post-install hook?

eb_hooks.py
```diff
@@ -684,7 +684,13 @@ def post_sanitycheck_cuda(self, *args, **kwargs):
     Remove files from CUDA installation that we are not allowed to ship,
     and replace them with a symlink to a corresponding installation under host_injections.
     """
-    if self.name == 'CUDA':
+    # Make sure we only do this for CUDA and only if we are doing a CVMFS installation
+    is_eessi_install = (
+        self.installdir.startswith("/cvmfs/software.eessi.io/versions")
+        and not build_option("sanity_check_only")
```
boegel, line 690: Pfft, not a big fan of this.

Why isn't this done through a post-install hook, is there a reason for that?

Also not a fan of the hardcoded check for `/cvmfs/software.eessi.io`... What if we do a "test" install of CUDA in `dev.eessi.io`?

We want to try and make sure we never ship a CUDA installation that isn't stripped down?

ocaisa: Suggestions welcome. Hook is where it is so that it doesn't impact the installation itself including the sanity check

ocaisa: Obviously I can add an alternate check for dev.eessi.io but I'm sensing general negativity

eb_hooks.py
ocaisa: Afraid not, that would cause the sanity check to (potentially) fail

eb_hooks.py
boegel: 1) W.r.t. this being a post-sanity-check hook: it's both good and bad this it's done there: good because it only kicks in after the sanity check performed by EasyBuild, but bad because we fiddle with the installation after the sanity check has completed (that's a necessary evil I guess).
In general, I'm fairly OK with that, but the additional checks being added now increase the risk that we accidentally won't strip the CUDA installation, imho.

I think we can probably guarantee that the full CUDA installation in `host_injections` will be there when we install the CUDA for the software layer that will be stripped down, so with that in mind it would be better to move this stripping down action to a post-install hook.

2) I'm not sure I fully understand the need for the check "are we installing under EESSI", but rather than having a very strict check like "install path starts with `/cvmfs/software.eessi.io/versions` or `/cvmfs/dev.eessi.io/...`, we could make it a regex check which is a lot less likely to accidentally pass when we don't want it to (like check for `^/cvmfs/[^/]*.eessi.io/`).
Can you clarify why we need that though? You basically want to make sure that this bit isn't triggered when we do the full CUDA install in `host_injections`?
Then why not explicitly check for that (if `/host_injections/` in installpath => skip stripping)?

3) Fixing the CPU duplication issue would be nice, but it's not something we should pursue right now I think, that could be a goal for the next EESSI version (since it's basically just cleanup, no real other impact)?

ocaisa: Checking for `host_injections` is not enough when deciding when to trigger the hook, people may be using the hook via `EESSI-extend` in which case you don't it triggered

ocaisa: The regex with `versions` should do it

ocaisa: I've done an implementation in https://github.com/EESSI/software-layer/pull/735/commits/a3eddb49ef404fa727436f64e59afb68e33d80f6 but now the method contains two hard-coded patterns: one to match to an EESSI-distributed repo and one for the location of `host_injections`, we may want to move these to a global definition

ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

boegel (dismissed):

eb_hooks.py
```diff
@@ -673,18 +685,24 @@ def pre_single_extension_testthat(ext, *args, **kwargs):
...
+def post_postproc_cuda(self, *args, **kwargs):
     """
     Remove files from CUDA installation that we are not allowed to ship,
     and replace them with a symlink to a corresponding installation under host_injections.
     """
-    if self.name == 'CUDA':
+
+    # We need to check if we are doing an EESSI-distributed installation
+    eessi_pattern = r"^/cvmfs/[^/]*.eessi.io/versions/"
+    host_injections_location = "/cvmfs/software.eessi.io/host_injections/"
```
boegel, line 702: Both of these could be top-level constants I guess

eb_hooks.py
```diff
@@ -732,7 +750,15 @@ def post_sanitycheck_cuda(self, *args, **kwargs):
...
-                        host_inj_path = full_path.replace('versions', 'host_injections')
+
+                        # the host_injections path is under a fixed repo/location for CUDA
+                        host_inj_path = re.sub(eessi_pattern, host_injections_location, full_path)
+                        # CUDA itself doesn't care about compute capability so remove this duplication from
+                        # under host_injections
+                        accel_subdir = os.getenv("EESSI_ACCELERATOR_TARGET")
+                        if accel_subdir:
+                            host_inj_path = host_inj_path.replace('/accel/%s' % accel_subdir, '')
+                            host_inj_path = host_inj_path.replace("/accel/%s" % accel_subdir, '')
```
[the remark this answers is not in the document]
ocaisa: Ah, sorry, I was fixing a conflict and accidentally left both

ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen2 accel:nvidia/cc80

ocaisa:
bot: build repo:eessi.io-2023.06-software arch:x86_64/amd/zen3 accel:nvidia/cc80

ocaisa:
Verified the links are as we expect, ready to deploy

bedroge:
Staging PRs have been merged, and the tarballs have been ingested.

bedroge (approved):

PR #752, opened 2024-09-26
Take accelerator builds into account in CI that checks for missing installations

bedroge:


bedroge:
As @boegel pointed out on Slack, this won't solve the failing checks for the CPU targets with an accelerator, as those will still look for CUDA builds (these are in the same easystack file). Closing this for now.

PR #753, opened 2024-09-27
Take accelerator builds into account in CI that checks for missing installations

bedroge:
It doesn't really seem possible to add complex combinations into the workflow matrix, and maybe we don't want that anyway (for instance because the accelerator builds also need the CPU stack). So, I've now defined a CPU->accelerators mapping and use that to:
- first run the "check missing" script for the CPU-only stack
- look up the accelerator targets for this particular CPU target
- loop over all accelerator targets and do the following for each of them:
  - prepend the module dir of this accelerator target to $MODULEPATH
  - run the "check missing" script for every accelerator easystack file
  - remove the module dir from the $MODULEPATH again

### DIFF
--- a/.github/workflows/test-software.eessi.io.yml
+++ b/.github/workflows/test-software.eessi.io.yml
@@ -10,17 +10,10 @@ permissions:
   contents: read # to fetch code (actions/checkout)
 env:
   EESSI_ACCELERATOR_TARGETS: |
-    aarch64/generic:
-    aarch64/neoverse_n1:
-    aarch64/neoverse_v1:
-    x86_64/generic:
     x86_64/amd/zen2:
       - nvidia/cc80
     x86_64/amd/zen3:
       - nvidia/cc80
-    x86_64/amd/zen4:
-    x86_64/intel/haswell:
-    x86_64/intel/skylake_avx512:
 jobs:
   check_missing:
     runs-on: ubuntu-latest
--- a/.github/workflows/test-software.eessi.io.yml
+++ b/.github/workflows/test-software.eessi.io.yml
@@ -73,7 +73,7 @@ jobs:
                   for accel in ${accelerators}; do
                       module use ${EESSI_SOFTWARE_PATH}/accel/${accel}/modules/all
                       echo "checking missing installations for accelerator ${accel} using modulepath: ${MODULEPATH}"
-                      for easystack_file in $(ls easystacks/software.eessi.io/${{matrix.EESSI_VERSION}}/accel/nvidia/eessi-${{matrix.EESSI_VERSION}}-eb-*.yml); do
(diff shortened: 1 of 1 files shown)


.github/workflows/test-software.eessi.io.yml
```diff
@@ -2,11 +2,25 @@
...
+    branches: "*"
   pull_request:
   workflow_dispatch:
 permissions:
   contents: read # to fetch code (actions/checkout)
+env:
+  EESSI_ACCELERATOR_TARGETS: |
+    aarch64/generic:
+    aarch64/neoverse_n1:
+    aarch64/neoverse_v1:
```
boegel, line 15: I would remove entries for which we don't have accelerator builds yet (and likely never will, like haswell)

.github/workflows/test-software.eessi.io.yml
```diff
@@ -56,6 +72,24 @@ jobs:
...
 
+              # now check the accelerator builds for this CPU target
+              accelerators=$(echo "${EESSI_ACCELERATOR_TARGETS}" | yq ".${EESSI_SOFTWARE_SUBDIR_OVERRIDE}[]")
+              if [ -z ${accelerators} ]; then
+                  echo "no accelerator targets defined for ${EESSI_SOFTWARE_SUBDIR_OVERRIDE}"
+              else
+                  for accel in ${accelerators}; do
+                      module use ${EESSI_SOFTWARE_PATH}/accel/${accel}/modules/all
+                      echo "checking missing installations for accelerator ${accel} using modulepath: ${MODULEPATH}"
+                      for easystack_file in $(ls easystacks/software.eessi.io/${{matrix.EESSI_VERSION}}/accel/nvidia/eessi-${{matrix.EESSI_VERSION}}-eb-*.yml); do
```
boegel, line 83: You should use `${accel} here rather than hardcoding `/accel/nvidia`?

boegel (changes requested):

.github/workflows/test-software.eessi.io.yml
bedroge: Done in https://github.com/EESSI/software-layer/pull/753/commits/4a5808c5f343b34190b0fd050023bd892ce23928.

.github/workflows/test-software.eessi.io.yml
bedroge: Done in https://github.com/EESSI/software-layer/pull/753/commits/eaf656a9d38e1af007484589f4fee615c99d5778 by using `dirname ${accel}`.

boegel: Hardcoding `accel/` is fine, but not the `nvidia` part, indeed

bedroge:
https://github.com/bedroge/software-layer/actions/runs/11067076764/job/30749648105
This is an older run where I used a wrong modulepath, it shows that it indeed fails when the accelerator builds cannot be found (hadn't synced my branch yet, so it only complains about LAMMPS).

PR #716, opened 2024-09-18
{2023.06}[foss/2023a] OSU-Microbenchmarks v7.2 w/ CUDA 12.1.1 (rebuild)

casparvl:
requires:
- [x] #720 
- [x] #719 
- [x] #750

PR #715, opened 2024-09-18
{2023.06}[foss/2023a] CUDA-Samples v12.1 w/ CUDA 12.1.1 (rebuild)

casparvl:
requires:
* #720

PR #762, opened 2024-09-30
move CUDA installations in CPU-only paths to dedicated easyconfigs

boegel:
Nothing needs to be installed here, this is just pulling things apart so the `*CUDA*` installations aren't mixed with the CPU-only installations.

The `*CUDA*` easystack files being added here will eventually be removed, since the `CUDA-Samples/12.1-GCC-12.3.0-CUDA-12.1.1` and `OSU-Micro-Benchmarks/7.2-gompi-2023a-CUDA-12.1.1` installation we currently have in the CPU-only installation paths don't belong there, they should go under `accel/nvidia`, see also:
* #715
* #716


laraPPr (approved):
lgtm

PR #765, opened 2024-09-30
Add check for missing Zen4 installations

laraPPr:


.github/workflows/test-software.eessi.io.yml
```diff
@@ -58,6 +59,13 @@ jobs:
               # first check the CPU-only builds for this CPU target
               echo "just run check_missing_installations.sh (should use easystacks/software.eessi.io/${{matrix.EESSI_VERSION}}/eessi-${{matrix.EESSI_VERSION}}-*.yml)"
               for easystack_file in $(ls easystacks/software.eessi.io/${{matrix.EESSI_VERSION}}/eessi-${{matrix.EESSI_VERSION}}-eb-*.yml); do
+                  if [ ${{matrix.EESSI_SOFTWARE_SUBDIR_OVERRIDE}} = "x86_64/amd/zen4" ]; then
+                      if grep -q 2022b <<<"${easystack_file}"; then
```
boegel, line 63: @laraPPr Makes sense, but let's add a comment to explain why we're doing this, both for `2022b` and `CUDA`?

casparvl: I'm still confused, but that might be me. We're also not building CUDA stuff for e.g. haswell right now, but we don't have an explicit skip there, right? Why do we need it for `zen4`?

boegel: We have the CUDA installations of `CUDA-Samples` and `OSU-MicroBenchmarks` installed for all CPU targets in the CPU-only path, but we don't want to do that going forward.

Those installations were split off into separate easystack files in #762 to make this possible.

Once we've made some more progress with the `accel/nvidia` part of the software layer, we'll likely remove those CUDA installations from the CPU path, they don't belong there.

boegel (approved):
lgtm

PR #771, opened 2024-10-02
tweak scripts to support building for `dev.eessi.io` repository

Neves-P:


init/minimal_eessi_env
```diff
@@ -20,4 +20,4 @@ fi
 export EESSI_CPU_FAMILY=$(uname -m)
 
 # set $EPREFIX since that is basically a standard in Gentoo Prefix
-export EPREFIX=$EESSI_PREFIX/compat/$EESSI_OS_TYPE/$EESSI_CPU_FAMILY
+export EPREFIX=$EESSI_COMPAT_LAYER_DIR
```
boegel, line 23: We'll need to be a bit more careful here, we should only use `$EESSI_COMPAT_LAYER_DIR` if it's defined (via `run_in_compat_layer_env.sh` for example), but it's possible it's not set (this script is also used by the EESSI init script), so:

```suggestion
# if $EESSI_COMPAT_LAYER_DIR is defined (for example by run_in_compat_layer_env.sh script), we use that value
if [ ! -z ${EESSI_COMPAT_LAYER_DIR} ]; then
    export EPREFIX=$EESSI_COMPAT_LAYER_DIR
else
    export EPREFIX=$EESSI_PREFIX/compat/$EESSI_OS_TYPE/$EESSI_CPU_FAMILY
fi
```

bot/build.sh
```diff
@@ -21,12 +21,14 @@
 # stop as soon as something fails
 set -e
 
+# Make sure we are referring to software-layer as working directory
+topdir=$(dirname $(dirname $(realpath $0)))
```
boegel, line 25: We should probably give this a more descriptive name, to avoid mistakes, like:

```suggestion
software_layer_dir=$(dirname $(dirname $(realpath $0)))

File: EESSI-install-software.sh [hunk 3]
```diff
@@ -232,7 +237,11 @@ pr_diff=$(ls [0-9]*.diff | head -1)
 # install any additional required scripts
 # order is important: these are needed to install a full CUDA SDK in host_injections
 # for now, this just reinstalls all scripts. Note the most elegant, but works
-${TOPDIR}/install_scripts.sh --prefix ${EESSI_PREFIX}
+
+# Only run install_scripts.sh if not dev.eessi.io for security
+if [[ "${EESSI_CVMFS_REPO}" != /cvmfs/dev.eessi.io ]]; then
```
boegel, line 242: Doing this makes sense, but it will cause fallout below where the `install_cuda_host_injections.sh` script is called.

OK to ignore for now since it's only relevant when we're building CUDA software in `dev.eessi.io`), but we'll probably need to have a construct like this:

```
if [[ "${EESSI_CVMFS_REPO}" == /cvmfs/dev.eessi.io ]]; then
    SCRIPTS_DIR=/cvmfs/software.eessi.io
else
    SCRIPTS_DIR=${EESSI_PREFIX}
    ${TOPDIR}/install_scripts.sh --prefix ${EESSI_PREFIX}
fi
```

and then use `$SCRIPTS_DIR` when calling `install_cuda_host_injections.sh` & co below

boegel (changes requested):

bot/build.sh
Neves-P: Addressed in 211840f

EESSI-install-software.sh
Neves-P: Make this a follow up PR

boegel:
Build done in #782 shows that no breaking changes were introduced here.

boegel (approved):
lgtm

PR #784, opened 2024-10-12
add support for detecting Zen5 (AMD Turin)

boegel:
see also https://github.com/archspec/archspec-json/pull/124

I made this a draft, because we should try and verify that this works first, but I don't know if/where there are AMD Turin-based systems available yet...


laraPPr:
@boege Can you create a new pr for this in https://github.com/EESSI/software-layer-scripts

PR #792, opened 2024-10-16
set `$LMOD_SHORT_TIME` to prevent Lmod from creating a user cache

boegel:
On some systems, Lmod may configured to create a user cache (in `~/.cache/lmod`) after a very short amount of time, for example on Vega:

```
$ ml --config 2>&1 | grep -i LMOD_.*_TIME
User cache valid time(sec) (LMOD_ANCIENT_TIME)                86400
Write cache after (sec) (LMOD_SHORT_TIME)                     2
```

So if a `module` command takes longer than 2 seconds, which can happen if CernVM-FS has to pull in an updated Lmod cache first from the EESSI repository, then a user cache will be auto-generated, which (by default) will be considered valid for 24h.

We should prevent that this happens, since it can easily lead to confusion when a module file is available under `/cvmfs/software.eessi.io`, but it's not visible via `module avail` (and if Lmod is configured with `LMOD_CACHED_LOADS` enabled, you won't be able to load it either, despite it being there).

Draft PR, since this script is not the only place we should set `$LMOD_SHORT_TIME` if we want to take control of this...

### DIFF
--- a/init/eessi_environment_variables
+++ b/init/eessi_environment_variables
@@ -116,7 +116,11 @@ if [ -d $EESSI_PREFIX ]; then
         else
           error "Lmod SitePackage.lua file not found at $lmod_sitepackage_file"
         fi
-       
+
+        # set time until user cache is created to 1 day (86400 seconds),
+        # to prevent that Lmod ever creates one (since it will quickly go stale)
+        export LMOD_SHORT_TIME=86400
+
         if [ ! -z $EESSI_BASIC_ENV ]; then
           show_msg "Only setting up basic environment, so we're done"
         elif [ -d $EESSI_SOFTWARE_PATH ]; then


init/eessi_environment_variables
```diff
@@ -116,7 +116,11 @@ if [ -d $EESSI_PREFIX ]; then
         else
           error "Lmod SitePackage.lua file not found at $lmod_sitepackage_file"
         fi
-       
+
+        # set time until user cache is created to 1 day (86400 seconds),
```
ocaisa, line 120: One day is a long time, what if they have their own modules? Can we set this as part of our own cache configuration instead so it is only relevant for us?

ocaisa: It seems not to be possible...

boegel: The whole point of setting it to 24h is to basically prevent that a user cache can ever be written.

ocaisa: But that's a global setting, so it will also never build cache for modules outside of EESSI

boegel: That's true, but i) only when the EESSI environment is set up, ii) that will "only" lead to longer waiting times when running `module` commands, there's no other negative side effects.

boegel: Maybe there's another way to prevent that specific paths (like `/cvmfs/software.eessi.io/...`) are included in the user cache, not sure.

@rtmclay would know... :)

ocaisa: In the `EESSI/2023.06` module that we ship, this path is protected:

https://github.com/EESSI/software-layer/blob/901a94459bddbd2dc410742883636704cc6201a4/init/modules/EESSI/2023.06.lua#L121-L124

so this wouldn't happen with that approach.

PR #795, opened 2024-10-18
Streamline the checks for missing installations

ocaisa:
This only checks the easystack files that are using the latest releases of EasyBuild (since this is our policy and there is little point in checking older easystacks over and over)


.github/workflows/test-software.eessi.io.yml
```diff
@@ -57,16 +57,14 @@ jobs:
...
-                          # skip the check of installed software on zen4 for foss/2022b builds
-                          continue
-                      elif grep -q CUDA <<<"${easystack_file}"; then
-                          # skip the check of install CUDA software in the CPU path for zen4
-                          continue
-                      fi
+                      # Make a temporary EasyStack file where we clean out all 2022b stuff and use that
+                      new_easystack=$(mktemp pruned_easystackXXX --suffix=.yml)
+                      # first clean out the options then clean out the .eb name
+                      sed '/2022b\|12\.2\.0/,/\.eb/{/\.eb/!d}' "${easystack_file}" | sed '/2022b\|12\.2\.0/d' > $new_easystack
```
ocaisa, line 66: It's a fluke that this passes, it doesn't delete all possible entries. However, the fact that it is needed is a hack anyway for the fact that we don't have the `2022b` gaps filled for Zen4

ocaisa: Rebuilding for 2022b is going to be pretty rare though...

.github/workflows/scripts/only_latest_easystacks.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
trz42, line 1: I'd add a short header (see other .sh scripts) with:
- purpose of the script
- author(s)
- license

.github/workflows/scripts/only_latest_easystacks.sh
```diff
@@ -0,0 +1,42 @@
...
+EESSI_VERSION=${EESSI_VERSION:-"2023.06"}
```
trz42, line 2: If you source `init/eessi_defaults` you don't have to change the version here later.

.github/workflows/scripts/only_latest_easystacks.sh
```diff
@@ -0,0 +1,42 @@
...
+
+directory="easystacks/software.eessi.io/${EESSI_VERSION}"
+# List of example filenames
+files=($(ls "$directory"/*.yml))
```
trz42, line 6: While maybe a rare case, but couldn't happen that one adds a new easystack file with a higher EB version first under `accel/...`? Then you would run tests for the second highest version, and would not use the new easystack file in the check.

.github/workflows/scripts/only_latest_easystacks.sh
```diff
@@ -0,0 +1,42 @@
...
+for item in "${all_latest_easystacks[@]}"; do
+  if [[ "$item" == *"$accel"* ]]; then
+    accel_latest_easystacks+=("$item")
+  else
+    cpu_latest_easystacks+=("$item")
+  fi
+done
+
+# Output the results
+[ -n "$ACCEL_EASYSTACKS" ] && echo "${accel_latest_easystacks[@]}" || echo "${cpu_latest_easystacks[@]}"
```
trz42, line 42: Maybe easier to read if put into standard `if`, `then`, `else`, ...?

trz42 (changes requested):
Looks fine mostly. A few suggestions for the script `only_latest_easystacks.sh`. Might also be worthwhile to put curly brackets around all variable names.

The one rebuild easystack file that was moved, it was not necessary for this PR, just something you noted?

.github/workflows/scripts/only_latest_easystacks.sh
ocaisa: I wanted this to be configurable in CI via an setting the envvar

.github/workflows/scripts/only_latest_easystacks.sh
ocaisa: Done in https://github.com/EESSI/software-layer/pull/795/commits/3ec0157ef11001586de4eb72c0e3e7bbd5ac6732

.github/workflows/scripts/only_latest_easystacks.sh
ocaisa: Done in https://github.com/EESSI/software-layer/pull/795/commits/3ec0157ef11001586de4eb72c0e3e7bbd5ac6732

.github/workflows/scripts/only_latest_easystacks.sh
ocaisa: The value here is actually set in the CI configuration, this is only here as a fall back in case someone explicitly calls this script for some reason

ocaisa:
@trz42 One rebuild easystack was moved because it was in the wrong location, it should have been under the `accel` subdirectory, the other had an inconsistent date stamp (just happened to see it)

.github/workflows/scripts/only_latest_easystacks.sh
trz42: Benefit of sourcing `init/eessi_defaults` is that one has to change the default in a single place only. In CI, one can set `EESSI_VERSION_OVERRIDE` before sourcing the defaults script.

Anyhow, I'll approve the PR.

trz42 (approved):
Looks good.

PR #801, opened 2024-10-30
add easystacks arg and bootstrap improvements

trz42:
Adds command line arg `--easystacks` to install script `EESSI-install-software.sh`. This can be used for interactively building software using one or multiple easystack files (comma-separated). Starting builds for a new architecture could be as simple as
```bash
export EESSI_SOFTWARE_SUBDIR_OVERRIDE=aarch64/neoverse_v2
./run_in_compat_layer_env.sh \
  EESSI-install-software.sh \
    --easystacks ${HOME}/es/eessi-2023.06-eb-4.9.4-001-system.yml,${HOME}/es/eessi-2023.06-eb-4.9.4-2023b.yml
```
provided `/cvmfs/software.eessi.io` is writable as when running `eessi_container.sh --access rw`.

The below is not needed after pulling in previously merged requests.
~Also improves bootstrapping of a new architecture:~
- ~adds `software` directory to `${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR_OVERRIDE}` as its existence is tested by the `EESSI/2023.06` module~
- ~checks if the module `EESSI-extend/${EESSI_VERSION}-easybuild` exists~
  - ~if so, it loads it~
  - ~if not, it doesn't load it but sets both `${EASYBUILD_INSTALLPATH}` and `${EASYBUILD_EXPERIMENTAL}` which are needed to install `EasyBuild/SOME_VERSION` and `EESSI-extend/${EESSI_VERSION}-easybuild`~


EESSI-install-software.sh
```diff
@@ -273,7 +279,16 @@ unset EESSI_PROJECT_INSTALL
...
+eessi_extend_module=EESSI-extend/${EESSI_VERSION}-easybuild
+module avail ${eessi_extend_module} 2>&1 | grep "${eessi_extend_module}"
+ec=$?
+if [ ${ec} -eq 0 ]; then
+    module load ${eessi_extend_module}
+else
+    echo "Did not find ${eessi_extend_module} module; setting EASYBUILD_INSTALLPATH and EASYBUILD_EXPERIMENTAL manually"
+    export EASYBUILD_INSTALLPATH=${EESSI_PREFIX}/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR_OVERRIDE}
+    export EASYBUILD_EXPERIMENTAL=1
+fi
```
ocaisa, line 291: This creates an alternative installation setup that is dependent on whether the `${eessi_extend_module}` is available or not. Rather than do that we should just have a script that ensures this module is in place (something we already have for EasyBuild itself, https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/load_easybuild_module.sh).

I'll look into extending the existing script to also install `EESSI-extend`

ocaisa: Suggestion for this in https://github.com/trz42/software-layer/pull/76

boegel:
@trz42 Can you sync this so we can get it merged?

EESSI-install-software.sh
ocaisa: @trz42 IIRC this part of the PR should no longer be necessary?

trz42:
> @trz42 Can you sync this so we can get it merged?

Updated the PR. It now only includes the changes related to passing an easystack file as argument to the script, then only that is processed instead of easystack files being determined from a `.diff` file.

boegel:
@trz42 Why does CI fail in this PR?

laraPPr:
@trz42 Can you move this pr to https://github.com/EESSI/software-layer-scripts

PR #814, opened 2024-11-14
clean `${tmpdir}` after iterations in `install_cuda_and_libraries.sh`

TopRichard:
The `${tmpdir}` can be cleaned after all iterations are done when having multiple `easystacks/eessi-*CUDA*.yml` files

### DIFF
--- a/scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
+++ b/scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
@@ -239,9 +239,9 @@ for EASYSTACK_FILE in ${TOPDIR}/easystacks/eessi-*CUDA*.yml; do
       echo_green "all installations at ${EESSI_SITE_SOFTWARE_PATH}/software/... succeeded!"
     fi
 
+    # clean up tmpdir content
+    rm -rf "${tmpdir}"/*
+
     # Restore MODULEPATH for next loop iteration
     MODULEPATH=${SAVE_MODULEPATH}
 done
-
-# clean up tmpdir
-rm -rf "${tmpdir}"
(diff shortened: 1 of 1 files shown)


TopRichard:

scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
```diff
@@ -239,9 +239,9 @@ for EASYSTACK_FILE in ${TOPDIR}/easystacks/eessi-*CUDA*.yml; do
...
 
-    # clean up tmpdir
-    rm -rf "${tmpdir}"
-
     # Restore MODULEPATH for next loop iteration
     MODULEPATH=${SAVE_MODULEPATH}
 done
+
+# clean up tmpdir
+rm -rf "${tmpdir}"
```
trz42, line 247: We can do a final clean-up here. However, shouldn't we also do a clean-up at the end of processing an easystack file? I think we may want to either create the tmpdir at the beginning of the loop (and remove it completely at the end) _OR_ remove the contents of the directory, e.g., (original line 243)
```bash
rm -rf "${tmpdir}"/*
```

trz42 (changes requested):
Removing the full `${tmpdir}` inside the loop likely lets the script fail if one processes multiple easystack files. However, the tmpdir directory is only needed for each loop iteration, so we don't need to wait with the clean up until the loop has processed all easystack files.

trz42 (approved):
lgtm

scripts/gpu_support/nvidia/install_cuda_and_libraries.sh
```diff
@@ -239,8 +239,8 @@ for EASYSTACK_FILE in ${TOPDIR}/easystacks/eessi-*CUDA*.yml; do
       echo_green "all installations at ${EESSI_SITE_SOFTWARE_PATH}/software/... succeeded!"
     fi
 
-    # clean up tmpdir
-    rm -rf "${tmpdir}"
+    # clean up tmpdir content
+    rm -rf "${tmpdir}"/*
```
ocaisa, line 243: Shouldn't we then remove `${tmpdir}` after the loop?

TopRichard: https://github.com/EESSI/software-layer/pull/814#discussion_r1846338150

ocaisa: Yes, but that approach only removes the contents of `$tmpdir`, never `$tmpdir` itself. This can be done outside the loop, see #816

PR #820, opened 2024-11-21
catch FATAL: messages when checking build result

trz42:
In some cases a job produces `FATAL: ` (error) messages which likely indicate some problem. For example, `fakeroot` support was missing on the new `zen4` build cluster, hence a software to be rebuilt wasn't actually rebuilt but no other error messages were produced. Thus, one could easily think the job succeeded.

This PR adds a check that determines if the job output contains a line with `FATAL: ` and adds the information to the result overview of a job.


ocaisa (approved):
LGTM

PR #836, opened 2024-12-06
Disable CUDA installation for RISC-V build hosts in `bot/build.sh` (and some other minor enhancements)

bedroge:
This PR
- adds some extra variables for the repository id, name, and version, and uses them throughout the script
- changes the check for the dev repo, as it basically checked if the repo id defined in the bot config equals "dev.eessi.io", but the repo id could in principle be anything. It's better to check against the real repository name.
- puts $SINGULARITY_TMPDIR under $JOB_STORAGE (job's temporary directory) instead of under $PWD (shared job dir), as it will only contain temporary files created by Singularity. This solves an issue that I had on the RISC-V cluster (it couldn't remove some files from the home directory at the end of the container session), and I think it's better to do it like this anyway.

### DIFF
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -125,7 +125,9 @@ else
 fi
 
 # determine repository to be used from entry .repository in ${JOB_CFG_FILE}
-REPOSITORY=$(cfg_get_value "repository" "repo_id")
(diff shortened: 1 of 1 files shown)


trz42 (approved):
lgtm

PR #837, opened 2024-12-06
Always set `$STORAGE` and `$JOB_STORAGE` in `bot/test.sh`

bedroge:
Even when resuming from a previous directory/tarball, `$STORAGE` and `$JOB_STORAGE` are not set, so I think we should always do this. For that reason, I moved them out of the if statement.

Also, this sets `$SINGULARITY_TMPDIR` to a subdir of `$JOB_STORAGE`, similar to the change made in #836.

### DIFF
--- a/bot/test.sh
+++ b/bot/test.sh
@@ -103,25 +103,25 @@ fi
 RESUME_DIR=$(grep 'Using .* as tmp directory' slurm-${SLURM_JOBID}.out | head -1 | awk '{print $2}')
 
 if [[ -z ${RESUME_DIR} ]]; then
-  echo -n "setting \$STORAGE by replacing any var in '${LOCAL_TMP}' -> "
-  # replace any env variable in ${LOCAL_TMP} with its
-  #   current value (e.g., a value that is local to the job)
-  STORAGE=$(envsubst <<< ${LOCAL_TMP})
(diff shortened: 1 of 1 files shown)


bot/test.sh
```diff
@@ -103,34 +103,34 @@ fi
 RESUME_DIR=$(grep 'Using .* as tmp directory' slurm-${SLURM_JOBID}.out | head -1 | awk '{print $2}')
 
 if [[ -z ${RESUME_DIR} ]]; then
```
trz42, line 105: If $RESUME_DIR is empty, we have to assume that we need to create a temporary storage using `$LOCAL_TMP` which can be configured via `app.cfg`.

bedroge: I'm now doing this in either case, so both when $RESUME_DIR is set and when it's not set. Or can we point $JOB_STORAGE to $RESUME_DIR or something in case $RESUME_DIR is set?

trz42 (approved):
OK, this should work. If we cannot find the temp storage directory from a previous run (`$RESUME_DIR` being empty), then we check if we can find a tarball of that run. If not, we exit with an error. If we find it we create the temporary storage.

This is now always available to set `$SINGULARITY_TMPDIR`.

PR #803, opened 2024-11-01
[TEST] RISC-V bot

bedroge:

PR #840, opened 2024-12-10
Let 2023.06 modulefile redirect RISC-V clients to riscv.eessi.io, add symlink for 20240402 version

bedroge:
This allows RISC-V clients (and, in particular, the build bot) to use the same modulefile as we use for software.eessi.io. By adding a symlink `20240402.lua -> 2023.06.lua` we can reuse the exact same modulefile. Note that I haven't added the symlink to `install_scripts.sh`, which ensures that it won't be deployed to any repo (it would be confusing to have this in software.eessi.io, and we don't actually need it in riscv.eessi.io either, as long as it's available in the git repo; the bot sources the file from the git repo). The only slightly confusing thing is that the RISC-V repo will deploy the 2023.06.lua modulefile, so users from that repo will only see `EESSI/2023.06`. But I'm not really sure how to prevent that, except by adding a bunch of specific if statements to `install_scripts.sh`.

Instead of this approach  we could also modify the EESSI-install-software.sh script and do something like this:
```
if [[ ${EESSI_CVMFS_REPO} == "/cvmfs/riscv.eessi.io" ]]; then
    # For the RISC-V repository we don't have a module file, so stick to the bash init script
    source $TOPDIR/init/bash
else
    module use $TOPDIR/init/modules
    module load EESSI/$EESSI_VERSION
fi
```
But I liked the solution with the modulefile a bit more.


bedroge:
Successful build with the RISC-V bot using this approach in https://github.com/EESSI/software-layer/pull/803#issuecomment-2528308258.

The latter still had a hardcoded `eessi_version = "20240402"`, it's now doing another one with the more flexible `eessi_version = os.getenv("EESSI_VERSION_OVERRIDE") or "20240402"`:
https://github.com/EESSI/software-layer/pull/803#issuecomment-2530785792
(from the logs I can already see that the build was successful, it's now running the test step)

trz42 (approved):
looks good!

bedroge:

trz42:
Tarball got ingested. New module file is available via CernVM-FS.

PR #844, opened 2025-01-05
add support for alternative artefacts checks in check-build.sh

smoors:
if `USE_CHECK_BUILD_ARTEFACTS_SCRIPT=1` is set, checking the tarball is skipped. instead, `check-build-artefacts.sh` is sourced, which may do alternative artefacts checks.

by default, `USE_CHECK_BUILD_ARTEFACTS_SCRIPT=0`, thus no change in behavior should be observed

this is useful for installations that are done directly on the system instead of CVMFS.

the changes are rather small if you hide the whitespace differences ;)

@boegel @laraPPr

### DIFF
--- a/bot/check-build.sh
+++ b/bot/check-build.sh
@@ -8,6 +8,7 @@
 # https://github.com/EESSI/software-layer.git
 #
 # author: Thomas Roeblitz (@trz42)
+# author: Samuel Moors (@smoors)
 #
 # license: GPLv2
 #
@@ -59,10 +60,12 @@ display_help() {
   echo " OPTIONS:"
   echo "  -h | --help    - display this usage information [default: false]"
   echo "  -v | --verbose - display more information [default: false]"
+  echo "  -a | --alt-artefacts - alternative artefacts check (sources file check-build-artefacts.sh if exists) [default: false]"
 }
 
 # set defaults for command line arguments
 VERBOSE=0
+ALT_ARTEFACTS=0
 
 POSITIONAL_ARGS=()
 
@@ -74,6 +77,10 @@ while [[ $# -gt 0 ]]; do
       ;;
     -v|--verbose)
       VERBOSE=1
+      shift 1
+      ;;
+    -a|--alt-artefacts)
+      ALT_ARTEFACTS=1
       shift 1
       ;;
     --)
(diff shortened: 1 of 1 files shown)


bot/check-build.sh
```diff
@@ -76,6 +79,10 @@ while [[ $# -gt 0 ]]; do
       VERBOSE=1
       shift 1
       ;;
+    -a|--alt-artefacts)
```
boegel, line 82: @smoors I would rename the option to `--use-check-build-artefacts-script`, which leaves (a lot) less room for confusion.

The `ALT_ARTEFACTS` constant should be renamed accordingly to `USE_CHECK_BUILD_ARTEFACTS_SCRIPT`

bot/check-build.sh
```diff
@@ -417,141 +428,149 @@ success_msg="found message(s) matching <code>${GP_no_missing}</code>"
...
+    echo "${TARBALL}" | sed -e 's/^/    /g' >> ${job_result_file}
+
+    # remove tmpfile
+    if [[ -f ${tmpfile} ]]; then
+        rm ${tmpfile}
+    fi
+
+elif [[ -f "$TOPDIR/check-build-artefacts.sh" ]]; then
+    source "$TOPDIR/check-build-artefacts.sh"
+fi
```
boegel, line 566: @smoors There should be an else here, in case `--use-check-build-artefacts-script` is used but not corresponding script is available:
```suggestion
else
    echo "ERROR: Required script $TOPDIR/check-build-artefacts.sh not found!" >&2
    exit 1
fi

**Review by boegel — CHANGES_REQUESTED** [review]

**Review by boegel — CHANGES_REQUESTED** [review]

**Review by boegel — APPROVED** [review]
lgtm

I think @laraPPr is planning a test PR on top of this to make sure this doesn't cause trouble, but it looks fine to me (didn't test it hands-on).

File: bot/check-build.sh [hunk 3]
```diff
@@ -417,141 +428,152 @@ success_msg="found message(s) matching <code>${GP_no_missing}</code>"
...
-comment_artefacts_fmt="<dt>_Artefacts_</dt><dd>__ARTEFACTS_LIST__</dd>"
-comment_artefacts="${comment_artefacts_fmt/__ARTEFACTS_LIST__/${comment_artefacts_details}}"
-comment_description=${comment_description/__ARTEFACTS_FMT__/${comment_artefacts}}
-
 echo "${comment_description}" >> ${job_result_file}
 
 # add overall result: SUCCESS, FAILURE, UNKNOWN + artefacts
 # - this should make use of subsequent steps such as deploying a tarball more
 #   efficient
 echo "status = ${status}" >> ${job_result_file}
```
laraPPr, line 547: This broke some stuff I think. I also don't know why this is removed?

laraPPr (changes requested):

bot/check-build.sh
laraPPr: See https://github.com/EESSI/software-layer/pull/902#issuecomment-2625346829

smoors: it wasn't removed but moved down. i hoped this would be ok, but apparently the order has to be strictly preserved..

laraPPr: Yes that table thing is tricky always scared of touching it

smoors: @laraPPr new commit hopefully fixes it https://github.com/EESSI/software-layer/pull/844/commits/e5990d213eb27485d0bca990f1d9736d3c75459a

laraPPr (approved):
lgtm

boegel (approved):

PR #866, opened 2025-01-21
{2023.06}[2023a] Rebuild SciPy-bundle 2023.07 with additional patches

bedroge:
These patches were used in https://github.com/EESSI/software-layer/pull/864 to successfully build this for Sapphire Rapids, so let's include the same patches for the other architectures by rebuilding this for all targets.

Issue #556, opened 2024-04-30
Permission denied errors when rebuilding (Python?) bundles with extensions that install binaries


bedroge:
When trying to rebuild Python, hatchling, and Python-bundle-PyPI in #546, we ran into weird permission issues for both hatchling and Python-bundle-PyPI. It's not clear yet what's causing it, but it seems to happen for (Python) bundles that include extensions that install files to the `bin` subdirectory. The removal step seems to work fine and successfully removes the existing installation, but in the build phase the extension will suddenly see the old `bin` directory again (with read-only permissions), and fail with errors like:
```
Successfully built hatchling
Installing collected packages: hatchling
ERROR: Could not install packages due to an OSError.
Consider using the `--user` option or check the permissions.
Traceback (most recent call last):
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/pip/_internal/commands/install.py", line 449, in run
    installed = install_given_reqs(
                ^^^^^^^^^^^^^^^^^^^
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/pip/_internal/req/__init__.py", line 72, in install_given_reqs
    requirement.install(
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/pip/_internal/req/req_install.py", line 800, in install
    install_wheel(
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/pip/_internal/operations/install/wheel.py", line 731, in install_wheel
    _install_wheel(
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen3/software/Python/3.11.3-GCCcore-12.3.0/lib/python3.11/site-packages/pip/_internal/operations/install/wheel.py", line 648, in _install_wheel
    generated_console_scripts = maker.make_multiple(scripts_to_generate)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #871, opened 2025-01-21
Workaround that solves permission issues for rebuilds

bedroge:
Solves #556, tested in #866.

It works around the issue by removing the original directory, then creating a new one with the first-level subdirs (e.g. `bin`, `lib`, `easybuild`, etc), and it makes sure that EB is then run with `--try-amend=keeppreviousinstall=True` to prevent EB from wiping it again (as it would otherwise run into the same issue again).

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -336,7 +336,15 @@ else
         if [ -f ${easystack_file} ]; then
             echo_green "Feeding easystack file ${easystack_file} to EasyBuild..."
 
-            ${EB} --easystack ${easystack_file} --robot
+            if [[ ${easystack_file} == *"/rebuilds/"* ]]; then
+                # the removal script should have removed the original directory and created a new and empty one
+                # to work around permission issues:
+                # https://github.com/EESSI/software-layer/issues/556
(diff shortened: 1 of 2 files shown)


boegel:
Interesting how that dances around the issue...

It almost feels like something isn't being synced fast enough, leading to the error we've seen in #556.

By pre-creating directories we essentially allow for more time to "sync", and then the problem doesn't present itself (or at the very least makes it less likely).

Good that we found a workaround for it.
Relying on `keeppreviousinstall` makes me a bit nervous though, since it was intended for very different purposes, but it should be fine in practice as long as we only have empty directories pre-created.

Can we verify this somehow, maybe in a hook that's triggered when `keeppreviousinstall` is enabled, as a safeguard?
We definitely don't want to have `keeppreviousinstall` enabled when the installation directory has anything other than empty subdirectories...

boegel:
@bedroge This was merged via #866, but it probably makes sense to put that extra check in place in our hooks?

bedroge:
> @bedroge This was merged via #866, but it probably makes sense to put that extra check in place in our hooks?

Done in https://github.com/EESSI/software-layer/pull/881.

PR #876, opened 2025-01-22
{2023.06}[2023b, sapphire_rapids] Bunch of apps from EB 4.9.2 2023b easystack

bedroge:


bedroge:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/sapphire_rapids

bedroge:
The commit used for R-bundle-CRAN includes the older version of Brunsli (which still depends on Highway). https://github.com/EESSI/software-layer/pull/876/commits/e84f1e4ae2d454eac6b88be41cf8a7621a7aa13c should fix this by using the commit that includes the new Brunsli version (without dependency on Highway) for R-bundle-CRAN as well.

Would be nice if we can just deploy this anyway, since all the builds have completed, and the build almost took a day. But maybe the bot won't allow us to do that?

bedroge:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/sapphire_rapids

bedroge:
I've merged the changes done in #880 into this branch, so we can test if the ESPResSO tests do run without issues now. Also removed some larger builds from the easytack, I'll do them in a separate PR.

bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/sapphire_rapids

Neves-P:
> The commit used for R-bundle-CRAN includes the older version of Brunsli (which still depends on Highway). [e84f1e4](https://github.com/EESSI/software-layer/commit/e84f1e4ae2d454eac6b88be41cf8a7621a7aa13c) should fix this by using the commit that includes the new Brunsli version (without dependency on Highway) for R-bundle-CRAN as well. I checked the diff of the R-bundle-CRAN easyconfig between these commits, and there is none, so this should be safe (i.e. doesn't require a rebuild for other architectures).
> 
> Would be nice if we can just deploy this anyway, since all the builds have completed, and the build almost took a day. But maybe the bot won't allow us to do that?

Doesn't look like it was an issue for the deployment :smile:  Is there any other reason why the this could be a problem for the ingestion, given that removing Brunsli makes the bundles identical between architectures, @bedroge ? Otherwise it lgtm and I'll go ahead and ingest it.

bedroge:
> > The commit used for R-bundle-CRAN includes the older version of Brunsli (which still depends on Highway). [e84f1e4](https://github.com/EESSI/software-layer/commit/e84f1e4ae2d454eac6b88be41cf8a7621a7aa13c) should fix this by using the commit that includes the new Brunsli version (without dependency on Highway) for R-bundle-CRAN as well. I checked the diff of the R-bundle-CRAN easyconfig between these commits, and there is none, so this should be safe (i.e. doesn't require a rebuild for other architectures).
> > Would be nice if we can just deploy this anyway, since all the builds have completed, and the build almost took a day. But maybe the bot won't allow us to do that?
> 
> Doesn't look like it was an issue for the deployment 😄 Is there any other reason why the this could be a problem for the ingestion, given that removing Brunsli makes the bundles identical between architectures, @bedroge ? Otherwise it lgtm and I'll go ahead and ingest it.

In the end (discussed with @boegel on Slack) we decided to do a rebuild anyway, and I removed the R bundle from this PR. It's being built separately in #884. So, this one should be good to go. Well, actually, while typing this I realize this includes the code from #880, let me remove that (doesn't affect the tarball/deployment).

bedroge:
@Neves-P Done, this is ready now.

Neves-P (approved):
Lgtm :smile:  Waiting for ingest comment to merge this PR

PR #881, opened 2025-01-23
Add pre-install hook that raises an error in case `keeppreviousinstall` (used for rebuilds) is set but the installation directory still contains files

bedroge:
This is an enhancement for #871 and implements the comment from @boegel in https://github.com/EESSI/software-layer/pull/871#issuecomment-2606523933.

I've tested it locally with `EESSI-extend` for the following scenarios:
- no `keeppreviousinstall` -> this always works, as expected
- with `--try-amend=keeppreviousinstall=True`:
  - empty install dir -> works
  - only some empty subdirs in the install dir -> works
  - file in the root of install dir -> error
  - file in a subdir -> error

You can test it by loading `EESSI-extend`, overriding `EASYBUILD_HOOKS` and setting it to the hooks file of this PR, and doing a small build with or without `--try-amend=keeppreviousinstall=True`.

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -9,7 +9,7 @@ from easybuild.easyblocks.generic.configuremake import obtain_config_guess
 from easybuild.framework.easyconfig.constants import EASYCONFIG_CONSTANTS
 from easybuild.tools.build_log import EasyBuildError, print_msg
 from easybuild.tools.config import build_option, update_build_option
-from easybuild.tools.filetools import apply_regex_substitutions, copy_file, remove_file, symlink, which
+from easybuild.tools.filetools import apply_regex_substitutions, copy_file, dir_contains_files, remove_file, symlink, which
 from easybuild.tools.run import run_cmd
 from easybuild.tools.systemtools import AARCH64, POWER, X86_64, get_cpu_architecture, get_cpu_features
 from easybuild.tools.toolchain.compiler import OPTARCH_GENERIC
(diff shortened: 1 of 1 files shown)


boegel:

eb_hooks.py
```diff
@@ -450,6 +450,16 @@ def pre_fetch_hook_zen4_gcccore1220(self, *args, **kwargs):
         print_msg("Updated build option 'force' to 'True'")
 
 
+def pre_install_hook(self, *args, **kwargs):
+    """Main pre install hook: trigger custom functions based on software name."""
+    if 'keeppreviousinstall' in self.cfg and self.cfg['keeppreviousinstall']:
```
boegel, line 455: @bedroge I think we deserve some context here, where/how we're using `keeppreviousinstall`, and why we only want to use this only when empty directories are available.
Definitely point to https://github.com/EESSI/software-layer/issues/556 + https://github.com/EESSI/software-layer/pull/871

boegel (changes requested):

eb_hooks.py
bedroge: Done: https://github.com/EESSI/software-layer/pull/881/commits/cc79fd7f1ab09e6857f4c894f1d2c69ded67e365

laraPPr:
@bedroge can you move these changes to https://github.com/EESSI/software-layer-scripts

PR #804, opened 2024-11-05
[WIP] Project subdirectories in `dev.eessi.io` install path and EESSI-extend integration

Neves-P:
**Please don't merge for now**

Previous installations to `dev.eessi.io` are landing in `/cvmfs/dev.eessi.io/versions/2023.06/` but should also include a project specific subdirectory such as `/cvmfs/dev.eessi.io/versions/2023.06/ESPResSo/`. Right now, this should be done by setting the environment variable `$EESSI_DEV_PROJECT` in the `dev.eessi.io` bot build script used in each of the projects (example: https://github.com/EESSI/dev.eessi.io-scripts/blob/main/bot/bot-build-dev.eessi.io.slurm), but maybe this can be improved by grabbing the information in the `job.cfg` file which includes a field like this:

``` cfg
[repository]
...
repo_id: repo_name
...
```

For now, however, I kept it simple to reduce points of failure until we are sure the installations land in the right place.

## EESSI-extend

An issue why this PR shouldn't be merged right now: EasyBuild is ignoring the cvmfs repository and installpath overrides and trying to write to `software.eessi.io` (and failing, because we mount this as read-only). Because this was working earlier I _think_ that some wires are getting crossed when we switched to `EESSI-extend` in `EESSI-install-software.sh` (see #790 ) and similarly to what is described at #802. I've not managed to figure out where things are going wrong.

Relevant error from log:

```
== building and installing ESPResSo/4.2.2-foss-2023a-1f672a00e22a9bc5fec1f7070bd28ff7718a706e...
  >> installation prefix: /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/ESPResSo/4.2.2-foss-2023a-1f672a00e22a9bc5fec1f7070bd28ff7718a706e
== FAILED: Installation ended unsuccessfully (build directory: /tmp/bot/easybuild/build/ESPResSo/4.2.2/foss-2023a-1f672a00e22a9bc5fec1f7070bd28ff7718a706e): build failed (first 300 chars): Failed to create lock /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/.locks/_cvmfs_software.eessi.io_versions_2023.06_software_linux_x86_64_amd_zen2_software_ESPResSo_4.2.2-foss-2023a-1f672a00e22a9bc5fec1f7070bd28ff7718a706e.lock: [Errno 38] Function not implemente (took 0 secs)
```

[...]

PR #885, opened 2025-01-24
Add dev.eessi.io support including project subdirectories

Neves-P:
This PR replaces open WIP PR #804 (to merge from a more up to date and tested branch) and adds support for building into the `dev.eessi.io` repository.

Documentation for how to install into `dev.eessi.io` is available in https://www.eessi.io/docs/adding_software/adding_development_software/, with maintainer documentation to come in https://github.com/EESSI/dev.eessi.io-example/pull/12

With this PR, software builds are installed into `/cvmfs/dev.eessi.io/2023.06/$EESSI_DEV_PROJECT/`. `$EESSI_DEV_PROJECT` is determined automatically during the build process and comes from the name of the GitHub repository that the bot is listening according to the format: `dev.eessi.io-EESSI_DEV_PROJECT`. See: https://github.com/EESSI/dev.eessi.io-example

Existing builds to `dev.eessi.io` were tested with the code that is about to be merged as verified in https://github.com/EESSI/dev.eessi.io-example/pull/14


Neves-P:
The CI fails in the detection of the `accel` sub-directory.

I _think_ that this is due to the string matching here: https://github.com/EESSI/software-layer/blob/2f16d6140f576f40e986ec88ae0684cec502e641/.github/workflows/tests_archdetect_nvidia_gpu.yml#L121

And it looks like the `dev.eessi.io` builds are introducing an extra unnecessary `/` which then fails the string matching. See: https://github.com/EESSI/software-layer/actions/runs/12946737807/job/36111860010#step:5:40

Neves-P:
> The CI fails in the detection of the `accel` sub-directory.
> 
> I _think_ that this is due to the string matching here:
> 
> https://github.com/EESSI/software-layer/blob/2f16d6140f576f40e986ec88ae0684cec502e641/.github/workflows/tests_archdetect_nvidia_gpu.yml#L121
> 
> And it looks like the `dev.eessi.io` builds are introducing an extra unnecessary `/` which then fails the string matching. See: https://github.com/EESSI/software-layer/actions/runs/12946737807/job/36111860010#step:5:40

I tried to fix that with 5b47737. Please don't merge yet even if the checks pass, until I've had a chance to test another `dev.eessi.io` build.

Neves-P:
Confirmed with https://github.com/EESSI/dev.eessi.io-example/pull/15 that this still works after 5b47737. I'd say this is ready to review then :smile:

EESSI-install-software.sh
```diff
@@ -293,12 +294,14 @@ if [ ! -z ${EESSI_ACCELERATOR_TARGET} ]; then
 fi
 
 # If in dev.eessi.io, allow building on top of software.eessi.io
-if [[ "${EESSI_CVMFS_REPO}" == /cvmfs/dev.eessi.io ]]; then
+if [[ ! -z ${EESSI_DEV_PROJECT} ]]; then
     module use /cvmfs/software.eessi.io/versions/$EESSI_VERSION/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR_OVERRIDE}/modules/all
+    # Overwrite EASYBUILD_INSTALLPATH to point to dev.eessi.io
+    EASYBUILD_INSTALLPATH=$EESSI_CVMFS_REPO_OVERRIDE/versions/$EESSI_VERSION/$EESSI_DEV_PROJECT/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR_OVERRIDE}
+    echo ">> \$EASYBUILD_INSTALLPATH overwritten to ${EASYBUILD_INSTALLPATH}"
+    module use $EASYBUILD_INSTALLPATH/modules/all
```
ocaisa, line 302: I believe this can all be removed by using
```
if [[ ! -z ${EESSI_DEV_PROJECT} ]]; then
    $EESSI_PROJECT_INSTALL=$EESSI_CVMFS_REPO_OVERRIDE/versions/$EESSI_VERSION/$EESSI_DEV_PROJECT/software/${EESSI_OS_TYPE}/${EESSI_SOFTWARE_SUBDIR_OVERRIDE}
fi
```
before line 255

ocaisa: Actually, 298 can also be removed, so indeed this whole `if` block

Neves-P: I'm working on this now and working out how to implement using `EESSI-extend` as we discussed in the support meeting, and I agree this `if` block isn't needed anymore. 

I'm wondering if we want`$EESSI_PROJECT_INSTALL`, though. Wouldn't we want instead `$EESSI_SITE_INSTALL` to make the installs land in `/cvmfs/dev.eessi.io/versions/$EESSI_VERSION/$EESSI_DEV_PROJECT/...` and available to everyone who mounts `/cvmfs/dev.eessi.io/`?

Neves-P: I've given this a try and was able to make it work, but I don't think we can proceed this way without other changes. The issue boils down to the way `EESSI-extend` handles the `EESSI_PROJECT_INSTALL` variable:

https://github.com/EESSI/software-layer/blob/13577753a4202e0f0b677ec772d0c3a4f5001550/EESSI-extend-easybuild.eb#L119 
In an example `dev.eessi.io` installation: `EESSI_SOFTWARE_PATH=/cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1`  and `EESSI_CVMFS_REPO=/cvmfs/dev.eessi.io`. There is no sensible path that can be passed as `EESSI_PROJECT_INSTALL` that would result in the string substitution going to the desired path of  `/cvmfs/dev.eessi.io/versions/2023.06/example/software/linux/aarch64/neoverse_n1`. 

To conclude that this is the problem I hacked the paths this way: 

https://github.com/Neves-P/software-layer/blob/921145140cfa637e647ecf222053b91d1d2cfba7/EESSI-install-software.sh#L254-L268

This ensures that the path gets substituted correctly, but at this point we might as well override `$EASYBUILD_INSTALLPATH` directly to get the same effect and avoid the obfuscation.

There may be another solution that I am not seeing, but I don't see a way out of this short of changing `EESSI-extend`, which I don't think is a good idea given what depends on that code. 

I'm quite stumped with how to proceed, what do you think @ocaisa ?

Neves-P:
Talked with @bedroge about this and decided to try just adding a simple if statement to EESSI extend that would account for builds targeting the development repository, changes in: ee2bcb1

I'd like to clarify to things:
1. How to test this without potentially breaking working code. I suppose that the change to the EESSI-extend module needs to actually be ingested first so that the new module file is used?
2. Does this actually work, or is there a better or improved way of achieving this.

@ocaisa , would you have any suggestions regarding this?


PS: I'm marking this a draft to prevent accidental merges as this could break production code. This PR also got extremely long with many commits reverting small changes and applying them again. I'm happy to open a new one from a clean branch to prevent polluting the repo's commit history with many trial and error commits.

ocaisa:
> Talked with @bedroge about this and decided to try just adding a simple if statement to EESSI extend that would account for builds targeting the development repository, changes in: [ee2bcb1](https://github.com/EESSI/software-layer/commit/ee2bcb13b2c6a7dad50d53f73a8abc9ebadf7a99)
> 
In general, I think this is the way to go. I would like to target the dev repos just the same as we do project extensions. There will need to be a step that generates the path with the subdirectory, but that could all be done with a single environment variable (if the variable `EESSI_DEV_PROJECT` is set, we are targeting `/cvmfs/dev.eessi.io` and should use the value of the variable to decide the subdirectory). That would mean you set the variable to the project and there is only a very small amount of special treatment to determine `EESSI_PROJECT_INSTALL`

> I'd like to clarify to things:
> 
> 1. How to test this without potentially breaking working code. I suppose that the change to the EESSI-extend module needs to actually be ingested first so that the new module file is used?

EESSI-extend is just an easyconfig that creates a module file, you can install it locally.

> 2. Does this actually work, or is there a better or improved way of achieving this.
> 
> @ocaisa , would you have any suggestions regarding this?
> 

Rather than complicate our lives with
```
/cvmfs/dev.eessi.io/versions/2023.06/example/software/linux/aarch64/neoverse_n1
```
perhaps we should be using
```
/cvmfs/dev.eessi.io/example/versions/2023.06/software/linux/aarch64/neoverse_n1
```
then you have direct mapping `/cvmfs/software.eessi.io` -> `/cvmfs/dev.eessi.io/example`

Neves-P:
Thanks @ocaisa , indeed that makes it a lot simpler. 

In fact, I think we don't need to even touch `EESSI-extend` (or just tweak the check to see if the project directory exists), because we should be able to just pass `/cvmfs/dev.eessi.io/example` as `EESSI_PROJECT_INSTALL` in one go from the`EESSI-install.sh`. I'll give this a try now on the example repo.

EESSI-extend-easybuild.eb
```diff
@@ -106,7 +106,6 @@ elseif (os.getenv("EESSI_SITE_INSTALL") ~= nil) then
   easybuild_installpath = os.getenv("EESSI_SITE_SOFTWARE_PATH")
 else
   -- Deal with user and project installs
-  project_install = os.getenv("EESSI_PROJECT_INSTALL")
```
ocaisa, line 109: This doesn't look right...

ocaisa (changes requested):

EESSI-extend-easybuild.eb
Neves-P: Changed in e012bd4d83b0e61cc0445de7a1c3e3507db0d1d1

EESSI-extend-easybuild.eb
```diff
@@ -106,7 +106,6 @@ elseif (os.getenv("EESSI_SITE_INSTALL") ~= nil) then
...
   project_modulepath = nil
   if (project_install ~= nil) then
     -- Check the folder exists
```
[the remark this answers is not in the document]
ocaisa: You can just make sure it exists before loading  EESSI-extend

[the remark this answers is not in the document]
Neves-P: Perfect, that should work! Trying another build now.

EESSI-extend-easybuild.eb
ocaisa: Still here...

create_tarball.sh
```diff
@@ -21,7 +22,18 @@ os="linux"
...
-software_dir="${cvmfs_repo}/versions/${eessi_version}/software/${os}/${cpu_arch_subdir}"
+# install to $eessi_version/$eessi_dev_project if defined, otherwise install
+# to $eessi_version
+install_prefix_dir=${eessi_version}${eessi_dev_project:+/$eessi_dev_project}
+echo "Setting Install prefix directory to: ${install_prefix_dir}"
+
+if [ -z ${install_prefix_dir} ]; then
+    echo "\${install_prefix_dir} is empty?!" >&2
+    exit 2
+fi
```
ocaisa, line 33: What is this supposed to match? It doesn't look like it could match the new location `/cvmfs/dev.eessi.io/${eessi_dev_project}/versions/...`

From the comment, shouldn't it be `$eessi_dev_project/versions/$eessi_version`

ocaisa: What about keeping the code as was, but updating line 23 such that
```
cvmfs_repo=${EESSI_CVMFS_REPO}${eessi_dev_project:+/$eessi_dev_project}
```

init/minimal_eessi_env
```diff
@@ -8,7 +8,17 @@ EESSI_INIT_DIR_PATH=$(dirname $(readlink -f $BASH_SOURCE))
...
+# If $EESSI_DEV_PROJECT is defined, then we are building for dev.eessi.io
+# and add a subdirectory to $EESSI_PREFIX
+if [ -z ${EESSI_DEV_PROJECT} ]; then
+    export EESSI_PREFIX=$EESSI_CVMFS_REPO/versions/$EESSI_VERSION
+else
+    echo "\$EESSI_DEV_PROJECT: ${EESSI_DEV_PROJECT}; building for dev.eessi.io"
+    export EESSI_PREFIX=$EESSI_CVMFS_REPO_OVERRIDE/versions/$EESSI_VERSION/$EESSI_DEV_PROJECT
+    # Set EESSI_PREFIX to the dev project path for EESSI-extend
+    export EESSI_CVMFS_INSTALL=$EESSI_PREFIX
+fi
```
ocaisa, line 20: Ditto here, you could just tweak the definition of `EESSI_CVMFS_REPO_OVERRIDE`

ocaisa (changes requested):

init/minimal_eessi_env
Neves-P: Agreed, but I think we need to set `$EESSI_CVMFS_REPO_OVERRIDE` before we call `minimal_eessi_env`, in `bot/build.sh`(?). I'm trying this now in acb896a163118d103e62e65566e052ce05c8ccc6

create_tarball.sh
Neves-P: Agreed! Changed in acb896a163118d103e62e65566e052ce05c8ccc6

EESSI-extend-easybuild.eb
Neves-P: You're right...! Sorry about that, hopefully fixed in 82da386f0c96a99f9d3a63739e5d1aad13e75362

Neves-P:
Not there yet, the check for `software_dir_overlay` fails in `create_tarball.sh` but should be close to sorted. I'll pick this up tomorrow.

Neves-P:
I've finished reworking and testing the `dev.eessi.io` build process after the review and taking into account changes that happened upstream in the meantime. For these changes to work, we also had to make small changes to the ingestion procedure, which were added with (merged) PR https://github.com/EESSI/filesystem-layer/pull/212.

I think this is ready for a new review. I tested the changes in this PR using my personal fork of the repository via https://github.com/EESSI/dev.eessi.io-example/pull/. The installations land in the right path: `/cvmfs/dev.eessi.io/example/...` and are available as expected.

@ocaisa , since you reviewed this before could you take another look and see what you think, please?

ocaisa (changes requested):
Few trivial suggestions

bot/build.sh
```diff
@@ -147,7 +147,7 @@ echo "bot/build.sh: EESSI_VERSION_OVERRIDE='${EESSI_VERSION_OVERRIDE}'"
 # determine CVMFS repo to be used from .repository.repo_name in ${JOB_CFG_FILE}
 # here, just set EESSI_CVMFS_REPO_OVERRIDE, a bit further down
 # "source init/eessi_defaults" via sourcing init/minimal_eessi_env
-export EESSI_CVMFS_REPO_OVERRIDE=/cvmfs/${REPOSITORY_NAME}
+export EESSI_CVMFS_REPO_OVERRIDE=/cvmfs/${REPOSITORY_NAME}${EESSI_DEV_PROJECT:+/$EESSI_DEV_PROJECT}
```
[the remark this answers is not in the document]
ocaisa: (maybe worth a comment to explain what is happening, I needed ChatGPT)

ocaisa (approved):
LGTM

PR #897, opened 2025-01-26
{2023.06}[2023a,sapphire_rapids] Remaining apps for EB 4.9.2 2023a easystack (except librosa)

bedroge:
This needs to wait for #896 to get ingested/merged.

```
4 out of 70 required modules missing:

* libxsmm/1.17-GCC-12.3.0 (libxsmm-1.17-GCC-12.3.0.eb)
* libvori/220621-GCCcore-12.3.0 (libvori-220621-GCCcore-12.3.0.eb)
* Libint/2.7.2-GCC-12.3.0-lmax-6-cp2k (Libint-2.7.2-GCC-12.3.0-lmax-6-cp2k.eb)
* CP2K/2023.1-foss-2023a (CP2K-2023.1-foss-2023a.eb)


1 out of 25 required modules missing:

* amdahl/0.3.1-gompi-2023a (amdahl-0.3.1-gompi-2023a.eb)


3 out of 99 required modules missing:

* LLVM/14.0.6-GCCcore-12.3.0-llvmlite (LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb)
* numba/0.58.1-foss-2023a (numba-0.58.1-foss-2023a.eb)
* librosa/0.10.1-foss-2023a (librosa-0.10.1-foss-2023a.eb)


1 out of 41 required modules missing:

* xarray/2023.9.0-gfbf-2023a (xarray-2023.9.0-gfbf-2023a.eb)


8 out of 127 required modules missing:

* xxHash/0.8.2-GCCcore-12.3.0 (xxHash-0.8.2-GCCcore-12.3.0.eb)
* python-xxhash/3.4.1-GCCcore-12.3.0 (python-xxhash-3.4.1-GCCcore-12.3.0.eb)
* Shapely/2.0.1-gfbf-2023a (Shapely-2.0.1-gfbf-2023a.eb)
* pyproj/3.6.0-GCCcore-12.3.0 (pyproj-3.6.0-GCCcore-12.3.0.eb)
* netcdf4-python/1.6.4-foss-2023a (netcdf4-python-1.6.4-foss-2023a.eb)
* Fiona/1.9.5-foss-2023a (Fiona-1.9.5-foss-2023a.eb)
* Cartopy/0.22.0-foss-2023a (Cartopy-0.22.0-foss-2023a.eb)
* SciTools-Iris/3.9.0-foss-2023a (SciTools-Iris-3.9.0-foss-2023a.eb)


2 out of 116 required modules missing:

[...]

PR #898, opened 2025-01-27
{2023.06}[2023a] librosa 0.10.1

bedroge:
The build of librosa with EB 4.9.2 failed in https://github.com/EESSI/software-layer/pull/897#issuecomment-2614568203, but a newer version of the easyconfig is available in 4.9.4, and it should fix this issue (see https://github.com/easybuilders/easybuild-easyconfigs/pull/21434).

Note that this shouldn't be merged yet, as it has dependencies on packages that are still being built in #897 (LLVM and numba), but I'm just giving it a try here already to make sure that it indeed fixes the issue.

If it works, we may want to consider doing a rebuild for the other CPU targets (though I'm not sure if it really changes much).

### DIFF
--- a/easystacks/software.eessi.io/2023.06/sapphire_rapids/eessi-2023.06-eb-4.9.4-2023a.yml
+++ b/easystacks/software.eessi.io/2023.06/sapphire_rapids/eessi-2023.06-eb-4.9.4-2023a.yml
@@ -1,0 +1,2 @@
+easyconfigs:
+  - librosa-0.10.1-foss-2023a.eb


bedroge:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/sapphire_rapids

bedroge:
Based on the contents of the librosa installation directory, I don't think we need to rebuild this for the other architectures. This is in `$EBROOTLIBROSA/0.10.1-foss-2023a/lib/python3.11/site-packages`, and it's the same for the zen3 installation:
```
audioread
audioread-3.0.1.dist-info
lazy_loader
lazy_loader-0.3.dist-info
librosa
librosa-0.10.1.dist-info
__pycache__
resampy
resampy-0.4.3.dist-info
soundfile-0.12.1.dist-info
_soundfile.py
soundfile.py
soxr
soxr-0.3.7.dist-info
```

A `diff -ru` on the `soxr` (which is the extension that actually failed to build in #897) directories mostly shows differences due to the installation prefix, and a change in some binary files (but that's expected, since it's also the case for, for instance, skylake_avx512 vs zen3).

bedroge:
bot: build repo:eessi.io-2023.06-software arch:x86_64/intel/sapphire_rapids

Neves-P (approved):

PR #903, opened 2025-01-31
Test building on Snellius: Zen4/H100

casparvl:
For now, I've set up a personal bot instance to build some experience with bot deployment. This PR is purely to test that instance.

PR #908, opened 2025-02-05
Fix bug when defining multiple modules in app.cfg

casparvl:
Fix bug when multiple modules where defined. In the old syntax, they were interpreted as a single string, with space. I.e. only one iteration of hte for loop was executed. Using bash arrays, we no longer have this issue

See https://github.com/EESSI/software-layer/pull/903#issuecomment-2635259858

### DIFF
--- a/bot/build.sh
+++ b/bot/build.sh
@@ -115,7 +115,8 @@ mkdir -p ${SINGULARITY_TMPDIR}
 
 # load modules if LOAD_MODULES is not empty
 if [[ ! -z ${LOAD_MODULES} ]]; then
-    for mod in $(echo ${LOAD_MODULES} | tr ',' '\n')
+    IFS=',' read -r -a modules <<< "$(echo "${LOAD_MODULES}")"
+    for mod in "${modules[@]}";
     do
         echo "bot/build.sh: loading module '${mod}'"
         module load ${mod}
--- a/bot/test.sh
+++ b/bot/test.sh
@@ -135,7 +135,8 @@ mkdir -p ${SINGULARITY_TMPDIR}
 
 # load modules if LOAD_MODULES is not empty
 if [[ ! -z ${LOAD_MODULES} ]]; then
-    for mod in $(echo ${LOAD_MODULES} | tr ',' '\n')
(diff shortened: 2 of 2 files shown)


trz42:
For me the original version works too :shrug:
```bash
[trz@login1 ~]$ source /cvmfs/software.eessi.io/versions/2023.06/init/lmod/bash
EESSI/2023.06 loaded successfully
[trz@login1 ~]$ ml
Currently Loaded Modules:
  1) EESSI/2023.06
[trz@login1 ~]$ ml av easybuild
----------------------------- /cvmfs/software.eessi.io/versions/2023.06/software/linux/aarch64/neoverse_n1/modules/all ------------------------------
   EasyBuild/4.8.2    EasyBuild/4.9.1    EasyBuild/4.9.3        EESSI-extend/2023.06-easybuild
   EasyBuild/4.9.0    EasyBuild/4.9.2    EasyBuild/4.9.4 (D)
...
[trz@login1 ~]$ export LOAD_MODULES=EasyBuild/4.9.4,EESSI-extend/2023.06-easybuild
[trz@login1 ~]$ for mod in $(echo ${LOAD_MODULES} | tr ',' '\n'); do echo +$mod+; done
+EasyBuild/4.9.4+
+EESSI-extend/2023.06-easybuild+
[trz@login1 ~]$ IFS=',' read -r -a modules <<< "$(echo "${LOAD_MODULES}")"
[trz42@login1 ~]$ for mod in "${modules[@]}"; do echo +$mod+; done
+EasyBuild/4.9.4+
+EESSI-extend/2023.06-easybuild+
```

If this works on our other bot instances, it doesn't hurt to change it.

trz42 (approved):
Works. Can't reproduce the problem with the original code though.

casparvl:
Just as an FYI: I'm not entirely sure why it happened either, I couldn't reproduce it interactively. But _what_ happened was very clear: it was interpreting the `$(echo ${LOAD_MODULES} | tr ',' '\n')` as a single item with a line-break, instead of looping over the modules. It almost must be something related to expansion / quoting /etc. 
In any case, the bash array approach should be more robust. So, thanks for merging :)

PR #910, opened 2025-02-10
Fix behavior for present but failing nvidiasmi

casparvl:
We have some cpu nodes on which the nvidia-smi command is installed, but failing 
```
$ nvidia-smi
NVIDIA-SMI has failed because it couldn't communicate with the NVIDIA driver. Make sure that the latest NVIDIA driver is installed and running.
```
I.e. simply checking for the presence of `nvidia-smi` and then concluding that GPUs are available (as we did prior to this PR) is not very robust. Since `nvidia-smi` does give a non-zero exit code in this case, it's pretty easy to improve the robustness of the check, which is done in this PR.

### DIFF
--- a/EESSI-install-software.sh
+++ b/EESSI-install-software.sh
@@ -271,10 +271,17 @@ fi
 
 # Install NVIDIA drivers in host_injections (if they exist)
 if command_exists "nvidia-smi"; then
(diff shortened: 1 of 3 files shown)


boegel:
Lots of copy-pasting, can we consider moving this check into a utility function so we don't have this duplicate code?

PR #911, opened 2025-02-12
Make sure that the TMPDIR is used as STORAGE if it is set

casparvl:


bot/build.sh
```diff
@@ -90,15 +90,23 @@ if [[ ! -z ${SINGULARITY_CACHEDIR} ]]; then
...
+    STORAGE=${TMPDIR}
+fi
+echo "bot/build.sh: STORAGE='${STORAGE}'"
 
 # make sure ${STORAGE} exists
 mkdir -p ${STORAGE}
 
+# Make sure ${STORAGE} gets bind-mounted
+# This will make sure that any subsequent jobs that create dirs or files under STORAGE have access to it in the container
+export SINGULARITY_BIND="${SINGULARITY_BIND},${STORAGE}"
```
trz42, line 108: I'm a little unsure if bind-mounting that unconditionally is the right thing to do. In https://github.com/trz42/software-layer/pull/79 I do the following (in `eessi_container.sh` not in `bot/build.sh` though):
```bash
# if TMPDIR is not empty and if TMP_IN_CONTAINER is not a prefix of TMPDIR, we need to add a bind mount for TMPDIR
if [[ ! -z ${TMPDIR} && ${TMP_IN_CONTAINER} != ${TMPDIR}* ]]; then
    echo "TMPDIR is not empty (${TMPDIR}) and TMP_IN_CONTAINER (${TMP_IN_CONTAINER}) is not a prefix of TMPDIR: adding bind mount for TMPDIR"
    BIND_PATHS="${BIND_PATHS},${TMPDIR}"
fi
```

This should ensure that if the location pointed to by `$TMPDIR` is not made available inside the container, we add it to the list of bind-mounts. To be fully correct, would probably need to check if any other bind-mount target is a prefix of `$TMPDIR`.

bot/build.sh
```diff
@@ -287,6 +295,9 @@ else
     TARBALL_STEP_ARGS+=("--resume" "${REMOVAL_TMPDIR}")
 fi
 
+# Make sure we define storage, so that the TMPDIR is set to this in eessi_container.sh
+TARBALL_STEP_ARGS+=("--storage" "${STORAGE}")
```
trz42, line 299: This should not be necessary, because we should resume from a previous temporary directory. `--storage` tells `eessi_container.sh` "instead of using a default location for temporary storage such as `$TMPDIR` or simply `/tmp`, use `$STORAGE`"

trz42 (changes requested):
We should limit this (bind-mounting `$TMPDIR`) to cases where it is needed.

The setting of `$STORAGE` in `bot/build.sh` is ok, the bind-mount part should be moved to `eessi_container.sh`

bot/build.sh
casparvl: Hm, I'm trying to figure out why I made this change, but don't see it. I'll first make sure I get a successful build again in https://github.com/EESSI/software-layer/pull/912 and then remove it, and try again.

bot/build.sh
casparvl: I'm fine with moving it to `eessi_container.sh`.

I need to double check that `${TMP_IN_CONTAINER} != ${TMPDIR}*` will resolve the original issue that I had. 
Essentially this check says, if:
```
TMP_IN_CONTAINER=/tmp/eessi_job.XYZ/foo
TMPDIR=/tmp/eessi_job.XYZ
```
It will not bind-mount `/tmp/eessi_job.XYZ`. I guess that's ok, since my original issue was that `mktemp` was failing because the `TMPDIR` didn't exist in the container. And I guess if `/tmp/eessi_job.XYZ/foo` exists, then so does `/tmp/eessi_job.XYZ`, and the `mktemp` _should_ succeed.

Also here: I'll first try to get a successful build in https://github.com/EESSI/software-layer/pull/912 , then I will make this change.

casparvl: Initial build https://github.com/EESSI/software-layer/pull/912#issuecomment-2654300700 succeeded. Now, implemented your suggested change and trying again https://github.com/EESSI/software-layer/pull/912#issuecomment-2654421669

bot/build.sh
casparvl: Initial build https://github.com/EESSI/software-layer/pull/912#issuecomment-2654300700 succeeded. Now, implemented your suggested change and trying again https://github.com/EESSI/software-layer/pull/912#issuecomment-2654421669

casparvl:
Ok, so that worked https://github.com/EESSI/software-layer/pull/912#issuecomment-2654421669 , I'm happy with this. I checked using

```
cat /path/to/slurm-9903139.out | grep \/tmp\/ | grep -v YUVixSjLBO
```
that there weren't any `/tmp` dirs that were _not_ the tempdir created for the job in the `bot-build.slurm`, there weren't. I.e. all where `/tmp/casparl/EESSI/eessi_job.YUVixSjLBO/<foo>`

PR #915, opened 2025-02-12
Add CUDA toolkits for zen4_h100

casparvl:
The build for this willl actually be done in https://github.com/EESSI/software-layer/pull/903#issuecomment-2654446727 . However, since I need several changes to make the bot work on my system, and we are not ready yet to merge those changes I've created this separate PR. The tarballs may then be deployed from #903 , while we will _merge_ this PR (#915) to represent the change in easystack files.


casparvl:
The build for this will actually be done here https://github.com/EESSI/software-layer/pull/903#issuecomment-2654446727 since I need modifications to the rest of the `software-layer` code in order to be able to build on my system. However, none of those changes affect the build of the tarball itself, so that might be ingested from 903, after which _this_ PR can be merged to represent the actual change in easystack files.

bedroge:
Built and deployed via #903.

bedroge (approved):

PR #937, opened 2025-02-24
proof of concept for GitHub action that opens an issue for a PR

trz42:
Making it a draft for now. Just a first attempt to automatically create an issue for a PR that is opened. We might want to add some convention that a PR that needs involvement of the build bot contains some specific keywords in the description or in the title and add some logic to the action such that it not always creates an issue.

Example of its use via https://github.com/trz42/software-layer/pull/81

### DIFF
--- a/.github/workflows/create_issue_on_pr_opened.yml
+++ b/.github/workflows/create_issue_on_pr_opened.yml
@@ -1,0 +1,48 @@
+name: Create Issue on PR Opened
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create an issue
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const title = `New PR opened: #${context.payload.pull_request.number} - ${context.payload.pull_request.title}`;
+            const body = `
+              A new pull request has been opened!
+
+              **Title:** ${context.payload.pull_request.title}
+              **Author:** ${context.payload.pull_request.user.login}
+
+              **Link to PR:** ${context.payload.pull_request.html_url}
+
+              <details><summary>Click for details & help</summary>
+
+              Authorized maintainers may send commands by adding new comments to this issue. A comment can contain multiple commands each starting at the beginning of a line and having the format <code>bot: COMMANDS [ARGS]</code>
+
+              The table below lists the commands that are currently supported:
+              | command | description |
+              | ------- | ----------- |
+              | help | prints short usage information |
+              | show_config | shows config information |
+              | status | shows status information of builds |
+              | build ARGS | instructs to build software as defined by the linked PR and with the one or more of the arguments:<br/>architecture, instance, repository, accelerator, exportvariable |
+
+              For more information see [building software for EESSI](https://www.eessi.io/docs/bot/#build-test-deploy-bot)
+              </details>
+            `;
+
+            const issue = await github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title: title,
+              body: body,
+            });
+
+            console.log(`Created issue: ${issue.data.html_url}`);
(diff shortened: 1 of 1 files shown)


.github/workflows/create_issue_on_pr_opened.yml
```diff
@@ -0,0 +1,48 @@
+name: Create Issue on PR Opened
+
+on:
+  pull_request:
+    types: [opened]
```
ocaisa, line 5: I think @laraPPr has played around with this to only trigger when certain file types are touched (in our case easystack files)

trz42: We should limit that, yes. Only easystack files may be too few, though. Should be easy to figure out for which files bot jobs need to run.

.github/workflows/create_issue_on_pr_opened.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create an issue
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
trz42: Added in https://github.com/EESSI/software-layer/pull/937/commits/47e14cfbc25b3e59a9c46617e464cc58af0bc17f

ocaisa:
I think this will work but I am not sure it is necessary. If the bot has permission to write a comment in an issue, it also has permission to create the issue in the first place.

It may save you from a race between bots to create an issue, but it also adds another step to configuring the bot.

trz42:
> I think this will work but I am not sure it is necessary. If the bot has permission to write a comment in an issue, it also has permission to create the issue in the first place.
> 
> It may save you from a race between bots to create an issue, but it also adds another step to configuring the bot.

Hmm, don't really see the additional configuration step. What happens now is the following:
1. PR opened $\rightarrow$ all running bot instances receive an event and each creates a comment ala _I'm alive. I can build for ..._
2. Maintainer adds a comment to the PR with build commands $\rightarrow$ all running bot instances receive an event, report back that they received the event (create a new comment in the PR), check if they need to launch build jobs and for each submission they create another comment. The latter comments are updated as the job progresses and the tarballs may eventually be uploaded. (Also if an instance encounters a problem in checking out the PR, it reports it with a separate comment to the PR.)
3. Finally, when the PR is closed/merged each bot checks if there is any data to be cleaned up and reports about that in another PR comment.

What has to be changed on the bot side is "only" figuring out which issue corresponds to the PR if any exists. If such an issue is identified it would use that to create comments. If commands are sent from the issue, this is easy. If they are sent from PR commands we need to implement some means to find it, e.g., searching for or accessing issues that reference the PR. There we would have to avoid that other issues referencing the PR might be found ... To avoid such complications, I'd change the bot code such that it doesn't act if the command is sent from a comment to the PR. This needs to be investigated a little more. Ideally one can link the PR and the Issue in a simple easily verifiable way.

.github/workflows/create_issue_on_pr_opened.yml
laraPPr: I was using this action for that https://github.com/dorny/paths-filter. I'm now looking to see if I can strip what we need from in it. So that we don't have to rely on external actions that much.

laraPPr: This pr includes the new filter https://github.com/EESSI/software-layer/pull/949

laraPPr:
@trz42 Can you retarget this pr?

ocaisa:
@trz42 I think the general idea here still has value, especially if we move to GitLab where we can create confidential issues. This would be great way of reporting back on failed builds (with logs analysed by LLMs) without having to worry about exposing sensitive information.

Having this issue write permission in CI though is troublesome in general though since we know these types of permissions are a weak security link. Using this approach via the bot seems like a better approach.

laraPPr:
Closing because @trz42 will not be returning to this one. @ocaisa do you think it is important enough to revive at this moment?

ocaisa:
When it comes to LLM processing of what is happening in a failed build, confidential comments are interesting...but that's not possible in GitHub (unless we make a separate private repo)

PR #948, opened 2025-02-27
Add a script that can sign a file based on an ssh key

ocaisa:
And can also verify a signature based on an allowed signers file.

The allowed signers file has the format
```
identity_1 <public ssh key 1>
identity_2 <public ssh key 2>
```

### DIFF
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -118,7 +118,7 @@ jobs:
           ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.alt.pem -N ""
           # Replace the allowed signers file 
           echo -n "disallowed_identity " > allowed_signers 
-          cat id_rsa.pem.pub >> allowed_signers
+          cat id_rsa.pem.alt.pub >> allowed_signers
           # Make sure signature checking fails in this case
           ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL" && exit 1 || echo "Expected failure for unknown identity"
 
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -105,12 +105,10 @@ jobs:
       run: |
           # Create a PEM format ssh identity
           ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.pem -N ""
-          # Create a file to sign
-          echo "I will sign this" > check_ssh_signing.txt
-          # bind current directory into container as /software-layer
-          export SINGULARITY_BIND="${PWD}:/software-layer"
-          # Sign the file in the container
-          ./eessi_container.sh --mode run --verbose /software-layer/sign_verify_file_ssh.sh sign /software-layer/check_ssh_signing.txt /software-layer/id_rsa.pem
+          # Find a file to sign
+          export TARBALL="$(ls *.tar.gz | tail -1)"
+          # Sign the file
+          ./sign_verify_file_ssh.sh sign "$TARBALL" id_rsa.pem
           # Create an allowed_signers file based on the public key
           echo -n "allowed_signer " > allowed_signers 
           cat id_rsa.pem.pub >> allowed_signers
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -12,6 +12,7 @@ on:
       - run_in_compat_layer_env.sh
       - scripts/utils.sh
       - update_lmod_cache.sh
+      - sign_verify_file_ssh.sh
 
   pull_request:
     paths:
@@ -24,6 +25,7 @@ on:
       - run_in_compat_layer_env.sh
       - scripts/utils.sh
       - update_lmod_cache.sh
+      - sign_verify_file_ssh.sh
 permissions:
   contents: read # to fetch code (actions/checkout)
 jobs:
@@ -88,6 +90,20 @@ jobs:
           # skip installation of CUDA SDKs, since this is too heavy for CI
           sed -i "s@./EESSI-install-software.sh@./EESSI-install-software.sh --skip-cuda-install@g" install_software_layer.sh
           ./eessi_container.sh --mode run --verbose /software-layer/install_software_layer.sh
+
+    - name: test sign_verify_file_ssh.sh script
+      run: |
+          # Create a PEM format ssh identity
+          ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.pem -N ""
+          # Create a file to sign
+          echo "I will sign this" > check_ssh_signing.txt
+          # Sign the file
+          ./sign_verify_file_ssh.sh sign check_ssh_signing.txt id_rsa.pem
+          # Create an allowed_signers file based on the public key
+          echo -n "allowed_signer " > allowed_signers 
+          cat id_rsa.pem.pub >> allowed_signers
+          # Verify the signature
+          ./sign_verify_file_ssh.sh verify check_ssh_signing.txt allowed_signers
 
     - name: test create_directory_tarballs.sh script
       run: |
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -116,10 +116,10 @@ jobs:
           ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL"
           # Make a new signature that does not appear in the allowed signers file
           ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.alt.pem -N ""
-          # Sign the file again
-          rm "$TARBALL".sig
-          ./sign_verify_file_ssh.sh sign id_rsa.alt.pem "$TARBALL"
-          # Make sure signature checking fails
+          # Replace the allowed signers file 
+          echo -n "disallowed_identity " > allowed_signers 
+          cat id_rsa.pem.pub >> allowed_signers
+          # Make sure signature checking fails in this case
           ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL" && exit 1 || echo "Expected failure for unknown identity"
 
     - name: test create_lmodsitepackage.py script
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -108,12 +108,19 @@ jobs:
           # Find a file to sign
           export TARBALL="$(ls *.tar.gz | tail -1)"
           # Sign the file
-          ./sign_verify_file_ssh.sh sign "$TARBALL" id_rsa.pem
+          ./sign_verify_file_ssh.sh sign id_rsa.pem "$TARBALL"
           # Create an allowed_signers file based on the public key
-          echo -n "allowed_signer " > allowed_signers 
+          echo -n "allowed_identity " > allowed_signers 
           cat id_rsa.pem.pub >> allowed_signers
           # Verify the signature
-          ./sign_verify_file_ssh.sh verify "$TARBALL" allowed_signers
+          ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL"
+          # Make a new signature that does not appear in the allowed signers file
+          ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.alt.pem -N ""
+          # Sign the file again
+          rm "$TARBALL".sig
+          ./sign_verify_file_ssh.sh sign id_rsa.alt.pem "$TARBALL"
+          # Make sure signature checking fails
+          ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL" && exit 1 || echo "Expected failure for unknown identity"
 
     - name: test create_lmodsitepackage.py script
       run: |
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -118,7 +118,7 @@ jobs:
           ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.alt.pem -N ""
           # Replace the allowed signers file 
           echo -n "disallowed_identity " > allowed_signers 
-          cat id_rsa.pem.alt.pub >> allowed_signers
+          cat id_rsa.alt.pem.pub >> allowed_signers
           # Make sure signature checking fails in this case
           ./sign_verify_file_ssh.sh verify allowed_signers "$TARBALL" && exit 1 || echo "Expected failure for unknown identity"
 
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -113,7 +113,7 @@ jobs:
           echo -n "allowed_signer " > allowed_signers 
           cat id_rsa.pem.pub >> allowed_signers
           # Verify the signature
-          ./sign_verify_file_ssh.sh verify check_ssh_signing.txt allowed_signers
+          ./sign_verify_file_ssh.sh verify "$TARBALL" allowed_signers
 
     - name: test create_lmodsitepackage.py script
       run: |
--- a/.github/workflows/tests_scripts.yml
+++ b/.github/workflows/tests_scripts.yml
@@ -91,20 +91,6 @@ jobs:
           sed -i "s@./EESSI-install-software.sh@./EESSI-install-software.sh --skip-cuda-install@g" install_software_layer.sh
           ./eessi_container.sh --mode run --verbose /software-layer/install_software_layer.sh
 
-    - name: test sign_verify_file_ssh.sh script
-      run: |
-          # Create a PEM format ssh identity
-          ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.pem -N ""
-          # Create a file to sign
-          echo "I will sign this" > check_ssh_signing.txt
-          # Sign the file
-          ./sign_verify_file_ssh.sh sign check_ssh_signing.txt id_rsa.pem
-          # Create an allowed_signers file based on the public key
-          echo -n "allowed_signer " > allowed_signers 
-          cat id_rsa.pem.pub >> allowed_signers
-          # Verify the signature
-          ./sign_verify_file_ssh.sh verify check_ssh_signing.txt allowed_signers
-
     - name: test create_directory_tarballs.sh script
       run: |
           # bind current directory into container as /software-layer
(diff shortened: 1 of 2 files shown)


ocaisa:
Example usage
```bash
ocaisa@LAPTOP-O6HF2IKC:~/ssh_sign$ ssh-keygen -t rsa -b 4096 -m PEM -f id_rsa.pem -N ""
Generating public/private rsa key pair.
Your identification has been saved in id_rsa.pem
Your public key has been saved in id_rsa.pem.pub
The key fingerprint is:
SHA256:hxgbCt+ZSDkpAci8Q42XCXxNfkYcacWw2xy/3/lgpmU ocaisa@LAPTOP-O6HF2IKC
The key's randomart image is:
+---[RSA 4096]----+
|*o= =..+*.       |
|.* B.+.+..       |
|. * =.++ .       |
| o = =oB+.o      |
|  . + *.So..     |
|         .  .    |
|           .  E  |
|            .*...|
|            .. oo|
+----[SHA256]-----+

ocaisa@LAPTOP-O6HF2IKC:~/ssh_sign$ ~/EESSI/software-layer/sign_verify_file_ssh.sh sign sample.tar id_rsa.pem
Converting SSH key to OpenSSH format...
No existing comment
Comment 'Converted from PEM' applied
Signing the tarball...
Signing file sample.tar
Write signature to sample.tar.sig
Signature created: sample.tar.sig

An allowed signatures file for this key would have contents like:

some_name ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCX7sjB/F5+/JbeWYK3sOCeg53HYTfFRfHuDY7wUd9dqhcVLIa9CVsYnIIzq8T4kmsBKN9rMubkjaKG4UIlQuLcbQu/IB3xXku+8bWfVe38YQBElOXdmkybR94C0DiKryHXfu2fC9bcePmMxuoOL+8WmTEX7lO/7b3zxnTeLuDuxBCxufIuBmsfEdswdixY7aUt8Hhk/zQOVaLHm8LYvwgcd8tpuhJ72fxM18CCPwZELtdxUoA8HoJGGDr+9aoTFwSKo0fvc9eL54zh+LXzitANkjVAl7USv6DMy3NLvLs2Akyd1zLpsaZkGcX9D8iGtlYzJ6Ru/nHHJgfFspmMdTdAkETGu5H8T1GamX2lXXn7BVDA1ox/aKM6aBPVcqKTccnSO21k+81B9bTCzxqRaanhKSzrDZLVtUz8FXEf+WMrQ458I/PY+/aA0AdWLbOWDh9K/CQ/QzUm8C9d6V7QXd9ymUBWP8fIQwtG2qKc8bj5UitNGlNU3akXsMyQUtNiN+JsK9MJeAmtgHlDhvpsR2m6rfJl+zyqCma1m2wNGfYBNGk4ww2tN0dQ1mWYgZXu3H2L5lAivxhi7LYRAU49Oa215X2FZWgZjOlQFZl+dxzbKVJPGYanDOAK/UZ97OZ1rEfYmEekua9mqqpo7f9ewz6dhPTcq2Dvm9vCssYhVjGOQw==

Validating the signature of the file...
Good "file" signature with RSA key SHA256:hxgbCt+ZSDkpAci8Q42XCXxNfkYcacWw2xy/3/lgpmU
- Signature validation successful.

ocaisa@LAPTOP-O6HF2IKC:~/ssh_sign$ ls
id_rsa.pem  id_rsa.pem.pub  sample.tar  sample.tar.sig

ocaisa@LAPTOP-O6HF2IKC:~/ssh_sign$ echo "allowed_signer" "$(cat id_rsa.pem.pub)" > allowed_signers

ocaisa@LAPTOP-O6HF2IKC:~/ssh_sign$ ~/EESSI/software-layer/sign_verify_file_ssh.sh verify sample.tar allowed_signers
Verifying signature for identity: allowed_signer
Good "file" signature for allowed_signer with RSA key SHA256:hxgbCt+ZSDkpAci8Q42XCXxNfkYcacWw2xy/3/lgpmU
Signature is valid for identity: allowed_signer
```

ocaisa:
We can do clever things with the allowed signers file, https://man.openbsd.org/ssh-keygen.1#ALLOWED_SIGNERS

.github/workflows/tests_scripts.yml
```diff
@@ -99,6 +101,20 @@ jobs:
...
+          # Find a file to sign
+          export TARBALL="$(ls *.tar.gz | tail -1)"
+          # Sign the file
+          ./sign_verify_file_ssh.sh sign "$TARBALL" id_rsa.pem
+          # Create an allowed_signers file based on the public key
+          echo -n "allowed_signer " > allowed_signers 
+          cat id_rsa.pem.pub >> allowed_signers
+          # Verify the signature
+          ./sign_verify_file_ssh.sh verify "$TARBALL" allowed_signers
+
```
trz42, line 117: A bit unsure what this **actually** tests? To me it looks like if the script exists in the current directory, is executable and exits with 0 the test will succeed?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+# Usage message
+usage() {
+    echo "Usage: $0 sign <file> <private_key> | verify <file> <public_key> [signature_file]"
```
trz42, line 5: Do we need verification here?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
```
trz42, line 2: Would be nice if the script has some standard header with

- short note about purpose,
- author(s)
- license

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+    usage
+fi
+
+MODE="$1"
+FILE_TO_SIGN="$2"
+
+# Ensure the file exists
+if [ ! -f "$FILE_TO_SIGN" ]; then
+    echo "Error: File '$FILE_TO_SIGN' not found."
+    exit 1
```
trz42, line 20: Maybe named constants for exit values (defined/declared at the beginning) would be nice.

sign_verify_file_ssh.sh
trz42: Usage info is very short. Maybe some information about the supported format of they private key would be good.

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+fi
+
+# Function to securely convert PEM to OpenSSH if needed
+convert_pem_to_ssh() {
+    local key_file="$1"
+    local output_file="$2"
+
+    if grep -q "BEGIN RSA PRIVATE KEY" "$key_file" || grep -q "BEGIN OPENSSH PRIVATE KEY" "$key_file"; then
+        # Ensure cleanup on exit
+        trap 'rm -f "$output_file" "$output_file.pub"' EXIT
```
trz42, line 30: That (`trap ...`) looks a bit fancy to me. First time we use it in any script in EESSI? Maybe not bad ... but then other scripts should be modernised too.

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+
+        # Convert PEM key to OpenSSH format
+        cp "$key_file" "$output_file" && ssh-keygen -c -C "Converted from PEM" -f "$output_file" || {
+            echo "Error: Failed to convert PEM key to OpenSSH."
+            exit 1
+        }
```
trz42, line 36: Again, a bit unusual style looking at other scripts in EESSI. Personally, I think using `if ... else ... fi` makes the code more easy to read.

trz42: What are the options for `ssh-keygen`? A small comment may help others to understand the code without checking a man page.

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+
+        # Generate the public key
+        ssh-keygen -y -f "$key_file" > "$output_file.pub" || {
+            echo "Error: Failed to generate public key from PEM key."
+            exit 1
+        }
```
trz42, line 42: See comment about unusual code style above (this one here is easier to read though).

What happens if the key is protected by a passphrase?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+if [ "$MODE" == "sign" ]; then
+    PRIVATE_KEY_PEM="$3"
+    PRIVATE_KEY="conversion_id"
+    SIG_FILE="${FILE_TO_SIGN}.sig"
+    PUB_KEY="${PRIVATE_KEY}.pub"
+
+    if [ ! -f "$PRIVATE_KEY_PEM" ]; then
+        echo "Error: Private key '$PRIVATE_KEY_PEM' not found."
+        exit 1
+    fi
```
trz42, line 59: Here the traditional style for conditional execution is used. Why not
```bash
[ ! -f "${PRIVATE_KEY_PEM} ] && {
    echo "..."
    exit 1
}
```

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+    echo "Converting SSH key to OpenSSH format..."
+    convert_pem_to_ssh "$PRIVATE_KEY_PEM" "$PRIVATE_KEY"
+
+    # Sign the file
+    echo "Signing the file..."
+    ssh-keygen -Y sign -f "$PRIVATE_KEY" -n file "$FILE_TO_SIGN"
+
+    if [ ! -f "$SIG_FILE" ]; then
+        echo "Error: Signing failed."
+        rm -f "$PRIVATE_KEY" "$PUB_KEY"
```
trz42, line 71: Is this needed? The `trap` should take care of that?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+    echo -e "\nAn allowed signatures file for this key would have contents like:\n"
+    echo -e "some_name $(cat "$PUB_KEY")\n"
+
+    # Verify the signature
+    echo -e "Validating the signature of the file..."
+    if ssh-keygen -Y check-novalidate -n file -f "$PUB_KEY" -s "$SIG_FILE" < "$FILE_TO_SIGN"; then
+        echo "- Signature validation successful."
+    else
+        echo "- Signature validation failed."
+        rm -f "$PRIVATE_KEY" "$PUB_KEY"
```
trz42, line 85: The set up `trap` should do this?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+
+    if [ ! -f "$SIG_FILE" ]; then
+        echo "Error: Signing failed."
+        rm -f "$PRIVATE_KEY" "$PUB_KEY"
+        exit 1
+    fi
+
+    echo "Signature created: $SIG_FILE"
+    echo -e "\nAn allowed signatures file for this key would have contents like:\n"
+    echo -e "some_name $(cat "$PUB_KEY")\n"
```
trz42, line 77: What's this about?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+    echo -e "Validating the signature of the file..."
+    if ssh-keygen -Y check-novalidate -n file -f "$PUB_KEY" -s "$SIG_FILE" < "$FILE_TO_SIGN"; then
+        echo "- Signature validation successful."
+    else
+        echo "- Signature validation failed."
+        rm -f "$PRIVATE_KEY" "$PUB_KEY"
+        exit 1
+    fi
+
+    rm -f "$PRIVATE_KEY" "$PUB_KEY"
```
trz42, line 89: Not handled by the `trap`?

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,128 @@
...
+
+            # Use ssh-keygen to verify the signature (example with RSA)
+            if ssh-keygen -Y verify -f "$ALLOWED_SIGNERS_FILE" -n file -I "$identity" -s "$SIG_FILE" < "$FILE_TO_SIGN"; then
+                echo "Signature is valid for identity: $identity"
+                exit 0  # Exit once we find a valid signature
+            else
+                echo "Invalid signature for identity: $identity"
+            fi
+        fi
+    done < "$ALLOWED_SIGNERS_FILE"
```
trz42, line 120: Is this the equivalent to importing GPG public keys?

trz42 (changes requested):
Overall a solid PR. Thanks @ocaisa !

A few thoughts:
- where should we put "this"? maybe under `bot/sign...` if we put it into the `software-layer repo`
- is this unique to the software-layer repo? how about putting this into the `scripts/...` directory of the `eessi-bot-software-layer` repository?
- should it really have both functions to sign file and to verify the signature of a file? the only reason for not separating them would be that through tests we make sure that both work together, i.e., if we change the signing, it shouldn't break the verification
  - the "bot" only really needs the sign part (unless the signature check uses the verification part)
  - the Stratum-0 needs the verification part
  - the bot could use the script (ideally via `bot/sign...`), but the Stratum-0 would need to get it elsewhere ... another option could be to put this (and maybe other) scripts into a shared repository (EESSI/common-tools)
- good idea to add a test (although I think it doesn't really test much functionality right now)
- a little bit too experimental/exotic with use of "fancy" bash capabilities (`trap`, not really fancy but unusual for EESSI so far; plus it seems what the trap should execute is then repeatedly done explicitly) and unusual syntax for conditional execution (also not consistently used throughout the script)

.github/workflows/tests_scripts.yml
ocaisa: This tests the workflow end-to-end, it creates a key (PEM format, same as a GitHub App key), signs a file with it, and then verifies the signature.

Basically it is testing core functionality, if there is a problem, the code will error out and the CI will fail.

sign_verify_file_ssh.sh
ocaisa: I added it because I don't want to leave these files lying around even if the script is interrupted

sign_verify_file_ssh.sh
ocaisa: It shows what the basic syntax for an allowed signers file is, I will clarify

sign_verify_file_ssh.sh
ocaisa: I haven't used GPG keys in 20 years so I'm not sure 😉 

This cycles through a given authorised signers file, checking to see if the tarball has been signed by any of the identities. Looking it up now, it seems you can give a wildcard `'*'` to the `-I` option, so the loop here may not be necessary.

sign_verify_file_ssh.sh
ocaisa: So there's an implicit assumption here that it is a GitHub App key and therefore has no passphrase. I can add support for an environment variable if there is a passphrase. I'm sure there must be a way to query if the key needs a passphrase...

ocaisa:
About where to put this, the creation of tarballs is done in this repo so I thought it made sense here. Now I think about it though, the bot is the one uploading the tarballs to S3, and the one who is aware of the GitHub App PEM key, so it may indeed be better off there.

.github/workflows/tests_scripts.yml
trz42: The test succeeds if the last run of `sign_verify_file_ssh.sh` exits with `0`, no? All the rest that is done in the test doesn't really matter. Wouldn't be better to use predefined input (data file, pem key) & output (pub key, signature) instead of generating them on the fly? Thus you would have input and output under control and could verify if the created pub key and signature match what you expect.

ocaisa: Not quite, the test will also fail if any of the previous commands fail. The first command tests the signing, the second command tests the verification. The specific data file and PEM key are not particularly relevant (we're not processing the data itself, only the signature file that gets created)

trz42: Maybe call it a unit test :shrug:

sign_verify_file_ssh.sh
ocaisa: I made a stab at this but they are a bit useless unless they are somehow consistent across scripts

sign_verify_file_ssh.sh
ocaisa: Passphrase is now supported

sign_verify_file_ssh.sh
ocaisa: Moved the trap to a global location

sign_verify_file_ssh.sh
ocaisa: Clarified (hopefully)

sign_verify_file_ssh.sh
ocaisa: Wildcards don't work so the loop remains

trz42 (commented):
Looks good. I'd like to test this in action with the bot. This will require a bit coding on my side. It should not really delay the use of this PR because we have to add the signing capability on the bot side anyhow.

sign_verify_file_ssh.sh
```diff
@@ -0,0 +1,154 @@
...
+TEMP_DIR=$(mktemp -d) || { echo "Error: Failed to create temporary directory."; exit $FILE_PROBLEM; }
+chmod 700 "$TEMP_DIR"
+trap 'rm -rf "$TEMP_DIR"' EXIT
+
+# Converts the SSH private key to OpenSSH format and generates a public key
+convert_private_key() {
+    local input_key="$1"
+    local output_key="$2"
+
+    echo "Converting SSH key to OpenSSH format..."
```
trz42, line 74: It seems we need to limit permissions of the converted key (after the addition below, signing worked in my case)
```suggestion
    echo "Converting SSH key to OpenSSH format..."
    # limit file permissions to max 0600 for the converted key
    umask 0177
```

trz42 (changes requested):
Added one more change that was necessary when I tested this. No other changes were needed.

sign_verify_file_ssh.sh
ocaisa: I'll implement this tomorrow, I think I will set this for the context of the whole script (so it also applies to the temporary directory creation where right now I explicitly set permissions)

ocaisa: Done in https://github.com/EESSI/software-layer/pull/948/commits/2be31133d3f7e3d4caf622c8b8e2c14795d3d303

ocaisa:
This is now moved over the Bot repository, see https://github.com/EESSI/eessi-bot-software-layer/pull/304

PR #949, opened 2025-02-27
Add new GPU workflow

laraPPr:
to check if missing dependencies for a CUDA build include cpu builds

This action includes an action that is not allowed yet by the EESSI organisation. Could the following action https://github.com/dorny/paths-filter be added to the allowed action.

BUG FIX: It also includes a fix for the existing missing installation check. Currently it fails after running `check_missing_installations.sh` if their are missing installations. And the workflow is stopped directly after which means the info on which easystack cause the failure is not printed. With this change it is.

### DIFF
--- a/.github/workflows/test-check_easystacks.yml
+++ b/.github/workflows/test-check_easystacks.yml
@@ -1,0 +1,77 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+name: Check for CUDA builds in changed easystacks
+on:
+  push:
+    branches: [ "*-software.eessi.io" ]
+  pull_request:
+  workflow_dispatch:
+permissions:
+  contents: read # to fetch code (actions/checkout)
+env:
+  EESSI_ACCELERATOR_TARGETS: |
+    x86_64/amd/zen2:
+      - nvidia/cc80
+    x86_64/amd/zen3:
+      - nvidia/cc80
+jobs:
+  check_missing:
+    runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        EESSI_VERSION:
+        - 2023.06
+        EESSI_SOFTWARE_SUBDIR_OVERRIDE:
+        - x86_64/amd/zen2
+        - x86_64/amd/zen3
+        - x86_64/amd/zen4
+        - x86_64/intel/haswell
+        - x86_64/intel/skylake_avx512
+        - x86_64/generic
+    steps:
+        - name: Check out software-layer repository
+          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+        - name: Mount EESSI CernVM-FS pilot repository
+          uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
+          with:
+              cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+              cvmfs_http_proxy: DIRECT
+              cvmfs_repositories: software.eessi.io
+        - name: find changed easystacks
+          uses: dorny/paths-filter@v3
+          id: changes
+          with:
+            list-files: shell
+            filters: |
+              changed:
+                - 'easystacks/**'
+        - name: Test check_missing_installations.sh script
+          if: ${{ steps.changes.outputs.changed == 'true' }}
+          run: |
+              export EESSI_SOFTWARE_SUBDIR_OVERRIDE=${{matrix.EESSI_SOFTWARE_SUBDIR_OVERRIDE}}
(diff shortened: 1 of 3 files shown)


.github/workflows/test-check_easystacks.yml
```diff
@@ -0,0 +1,77 @@
...
+          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+        - name: Mount EESSI CernVM-FS pilot repository
+          uses: cvmfs-contrib/github-action-cvmfs@55899ca74cf78ab874bdf47f5a804e47c198743c # v4.0
+          with:
+              cvmfs_config_package: https://github.com/EESSI/filesystem-layer/releases/download/latest/cvmfs-config-eessi_latest_all.deb
+              cvmfs_http_proxy: DIRECT
+              cvmfs_repositories: software.eessi.io
+        - name: find changed easystacks
+          uses: dorny/paths-filter@v3
```
ocaisa, line 42: This would need to use a specific commit from that repo to pass our security checks
```suggestion
          uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36  #v3.0.2
```

laraPPr: Does something else need to be done because I'm still seeing this? https://github.com/EESSI/software-layer/actions/runs/13587202930

hvelab:
- Agreed that we don't need the action but is better to do it ourselves 
- The other option is to accept -> but security concerns

.github/workflows/test-software.eessi.io.yml
```diff
@@ -60,8 +60,7 @@ jobs:
               echo "just run check_missing_installations.sh (should use easystacks/software.eessi.io/${{matrix.EESSI_VERSION}}/eessi-${{matrix.EESSI_VERSION}}-*.yml with latest EasyBuild release)"
               for easystack_file in $(EESSI_VERSION=${{matrix.EESSI_VERSION}} .github/workflows/scripts/only_latest_easystacks.sh); do
                   echo "check missing installations for ${easystack_file}..."
-                  ./check_missing_installations.sh ${easystack_file}
-                  ec=$?
+                  ./check_missing_installations.sh ${easystack_file} || ec=$?
```
ocaisa, line 63: I don't think this will work, it only sets `ec` if the script did not exit successfully

laraPPr: Yes that is what I want otherwise the workflow stops. And than this error message does not get printed  `"missing installations found for ${easystack_file}!"`

laraPPr: I can test again to see if I overlooked something.

laraPPr: Ah no your right

laraPPr: Ok than I'll have to find another way around it

casparvl: can't you just do
```
ec=0
./check_missing_installations.sh ${easystack_file} || ec=$?
```
If it exits succesfully, ec remains 0, if it fails, you get the exit code. That's what you want, right?

casparvl: Or just
```
./check_missing_installations.sh ${easystack_file}
ec=$?
```

ocaisa, line 63:
```suggestion
                  ec=0
                  ./check_missing_installations.sh ${easystack_file} || ec=$?
```
This will default the `ec` to success

.github/workflows/test-software.eessi.io.yml
```diff
@@ -75,8 +74,7 @@ jobs:
                       echo "checking missing installations for accelerator ${accel} using modulepath: ${MODULEPATH}"
                       for easystack_file in $(EESSI_VERSION=${{matrix.EESSI_VERSION}} ACCEL_EASYSTACKS=1 .github/workflows/scripts/only_latest_easystacks.sh); do
                           echo "check missing installations for ${easystack_file}..."
-                          ./check_missing_installations.sh ${easystack_file}
-                          ec=$?
+                          ./check_missing_installations.sh ${easystack_file} || ec=$?
```
ocaisa, line 77:
```suggestion
                          ec=0
                          ./check_missing_installations.sh ${easystack_file} || ec=$?
```

.github/workflows/test-software.eessi.io.yml
laraPPr: yes I will try tomorrow

laraPPr:
Filter now works see https://github.com/EESSI/software-layer/actions/runs/15493894354/job/43625796367?pr=949
And the check fails as expected see https://github.com/EESSI/software-layer/actions/runs/15493419723/job/43624193008?pr=917

ocaisa:
@laraPPr I think we can close this as this check is now being done in our EB hooks

PR #963, opened 2025-03-05
Add OpenMPI host injection script

pfermi:
Bash script that given a path to an OpenMPI installation in the host, makes a copy of the  OpenMPI libraries,
libfabric and libpmix and inject the necessary host libraries with the absolute path. In order to know which libraries must be injected it uses `ldd` from EESSI (`${EESSI_EPREFIX}/usr/bin/ldd`) and the host (`/usr/bin/ldd`).

The script is organized in functions. The function inject_mpi is the one actually performing the injection. There is a download_patchelf function in charge of downloading version v0.17.2 of patchelf, because the patchelf shipped with EESSI could not do the injection successfully.

The initial injection happens in a temporary directory, that can be specified from the command line, but if not specified it uses `mktemp -d`. At the end the temporary directory is removed, unless `--noclean` is specified in command line execution of the script.

The script, by default, does not perform the injection if the directory `.../host_injections/rpath_overrides/OpenMPI/system/lib` is not empty, unless `--force` is given to the script.

### DIFF
--- a/scripts/mpi_support/install_openmpi_host_injection.sh
+++ b/scripts/mpi_support/install_openmpi_host_injection.sh
@@ -149,8 +149,9 @@ inject_mpi() {
     readarray -d '' dlopen_libs < <(find ${MPI_PATH} -mindepth 2 -name "*.so*")
 
     # Get all library names and paths in associative array
+    # If library is libfabric, libpmix, or from the MPI path
+    # modify libpath in assoc array to point to host_injection_mpi_path
     while read -r libname libpath; do
-        # If library is libfabric or from the MPI path, modify libpath in assoc array to point to host_injection_mpi_path
         if [[ ${libname} =~ libfabric\.so ]] && [[ ! -f ${temp_inject_path}/${libname} ]]; then
             local libdir="$(dirname ${libpath})/"     # without trailing slash the find does not work
             find ${libdir} -maxdepth 1 -type f -name "libfabric.so*" -exec cp {} ${temp_inject_path} \;
@@ -188,7 +189,7 @@ inject_mpi() {
                  <(for dlopen in ${dlopen_libs[@]}; do ${system_ldd} ${dlopen}; done) \
             | awk '/=>/ {print $1, $3}' | sort | uniq)
 
-    # Get MPI related lib dependencies not resolved by EESSI ldd
+    # Do library injection to openmpi libs, libfabric and libpmix
     local lib
     while read -r lib; do
         local dep
--- a/scripts/mpi_support/install_openmpi_host_injection.sh
+++ b/scripts/mpi_support/install_openmpi_host_injection.sh
@@ -153,22 +153,22 @@ inject_mpi() {
     local lib
 
     while read -r lib; do
-    local dep
+        local dep
 
-    ${PATCHELF_BIN} --set-rpath "" ${lib}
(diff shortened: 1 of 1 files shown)


ocaisa:
@pfermi You should be able to create CI for this since you can install OpenMPI from the OS.

TopRichard:
Tested the script on OpenMPI-5.0.7 build with EasyBuild(draft easyconfig) using eessi_container and the following was noted:
As the results shows: 
```
patchelf: open: Permission denied
patchelf: open: Permission denied
patchelf: open: Permission denied
patchelf: open: Permission denied
patchelf: open: Permission denied
MPI injection was successful
``` 

1. The `patchelf` command failed to replace the needed libraries, and thus the MPI injections should not be successful
2. The library files lack write mode, which can be fixed by `chmod u+w` for the library files within the tmp dir before trying to `patchelf` 
3. After applying the change mentioned above, the script injects the following in host_injections:
`ls host_injections/2023.06/software/linux/aarch64/generic/rpath_overrides/OpenMPI/system/lib/
libfabric.so  libfabric.so.1  libfabric.so.1.26.0  libpmix.so  libpmix.so.2  libpmix.so.2.8.0
` 
            -  should **libmpi.so.40** be also injected? 
            -   some of the above are symbolic links :
 ```
libfabric.so -> libfabric.so.1.26.0
libfabric.so.1 -> libfabric.so.1.26.0
libfabric.so.1.26.0
libpmix.so -> libpmix.so.2.8.0
libpmix.so.2 -> libpmix.so.2.8.0
libpmix.so.2.8.0
```

ocaisa:
@TopRichard Can you show the command line you used so it's clear where you pointed the script to? Indeed, the final contents seem to be missing critical libraries.

@pfermi The script should only apply `patchelf` if it is actually required (and indeed it will need write permissions on the copy). This sounds like a good second CI test for what I hope will be a common use case: someone builds a custom OpenMPI with EESSI as a base and then injects that. For CI, I would take a specific existing installation of OpenMPI from within EESSI and run the injection on that. In this case, if there is no need to `patchelf` anything, I would just create symlinks (this would also allow a rebuild to be automatically captured).

TopRichard:
> @TopRichard Can you show the command line you used so it's clear where you pointed the script to? Indeed, the final contents seem to be missing critical libraries.

>{EESSI 2023.06} Apptainer> /cvmfs/software.eessi.io/versions/2023.06/scripts/mpi_support/install_openmpi_host_injection.sh -t /tmp/mpi --noclean --mpi-path /p/project1/ceasybuilders/MPI-rt/Extra/software/OpenMPI/5.0.7-GCC-12.3.0/ 

> @pfermi The script should only apply `patchelf` if it is actually required (and indeed it will need write permissions on the copy). This sounds like a good second CI test for what I hope will be a common use case: someone builds a custom OpenMPI with EESSI as a base and then injects that. For CI, I would take a specific existing installation of OpenMPI from within EESSI and run the injection on that. In this case, if there is no need to `patchelf` anything, I would just create symlinks (this would also allow a rebuild to be automatically captured).

agimenog:
Hi @ocaisa,

While we're currently working on the issues we have found with the script injections, could you kindly clarify what you meant with the first CI test? To be able to trigger the script via GitHub actions or you mean something else? 

Regards,
Arturo.

PR #972, opened 2025-03-20
Add handling of the same-name libraries on different locations for link_nvidia_host_libraries.sh

Darkless012:
Check for duplicates when reading ldconfig directories.


.github/workflows/tests_link_nvidia_host_libraries.yml
```diff
@@ -152,9 +154,9 @@ jobs:
             exit 1
           fi
           
-          # Verify it points to our mock library in /tmp/nvidia_libs
-          if [[ "$target" != "/tmp/nvidia_libs/$lib"* && "$target" != *"/tmp/nvidia_libs/"* ]]; then
-            echo "Error: Symlink $lib_path points to $target, which is not in our mock directory"
+          # Verify it points to our mock library in /tmp/nvidia_libs or /tmp/nvidia_libs_duplicate
+          if [[ "$target" != "/tmp/nvidia_libs/$lib"* && "$target" != "/tmp/nvidia_libs_duplicate/$lib"* ]]; then
```
ocaisa, line 158: Shouldn't we know exactly which one it should point to? And there should be something in the output that says the other one was filtered

scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
```diff
@@ -402,8 +402,41 @@ find_cuda_libraries_on_host() {
...
+                echo "Checking library $lib_name for duplicates"
+                
+                # Check if we already have this library filename in our matched libraries
+                duplicate_found=0
+                for existing_lib in "${MATCHED_LIBRARIES[@]}"; do
+                    existing_name=$(basename "$existing_lib")
+                    if [ "$existing_name" = "$lib_name" ]; then
+                        log_verbose "Duplicate library found: $lib_name (existing: $existing_lib, new: $lib_path)"
+                        # Prioritize libraries in standard locations if possible
+                        if [[ "$lib_path" == "/usr/lib"* || "$lib_path" == "/lib"* ]]; then
```
ocaisa, line 423: I'm not sure if this will catch all cases. Why not just take the first option? That is what the linker itself would do.

scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
```diff
@@ -484,12 +517,30 @@ symlink_mode () {
...
+            fi
     
             # Create a symlink in the current directory
             # and check if the symlink was created successfully
             if ! ln -s "$library" .
             then
-                fatal_error "Error: Failed to create symlink for library $library in $PWD"
+                echo_yellow "Warning: Failed to create symlink for library $library in $PWD"
+                # Continue instead of fatal_error to make the script more robust
+                continue
```
ocaisa, line 543: And what happens if the creation of all symlinks fails? I don't think this helps with robustness, to me this is an error.

ocaisa (changes requested):

scripts/gpu_support/nvidia/link_nvidia_host_libraries.sh
Darkless012: Reverted it back to fatal_error.

boegel:

boegel (approved):

boegel (approved):

bedroge:
I'm going to merge this, the CI is still failing due to issues with `lit`, and that's not related to this PR. Not merging this will make all current builds pick up the old version of this file.

PR #946, opened 2025-02-26
{2023.06}[2023b,sapphirerapids] Add EB 4.9.3 easystack for 2023b

bedroge:
```
1 out of 60 required modules missing:

* GROMACS/2024.3-foss-2023b (GROMACS-2024.3-foss-2023b.eb)


15 out of 118 required modules missing:

* archspec/0.2.2-GCCcore-13.2.0 (archspec-0.2.2-GCCcore-13.2.0.eb)
* Voro++/0.4.6-GCCcore-13.2.0 (Voro++-0.4.6-GCCcore-13.2.0.eb)
* ffnvcodec/12.1.14.0 (ffnvcodec-12.1.14.0.eb)
* kim-api/2.3.0-GCC-13.2.0 (kim-api-2.3.0-GCC-13.2.0.eb)
* x264/20231019-GCCcore-13.2.0 (x264-20231019-GCCcore-13.2.0.eb)
* Yasm/1.3.0-GCCcore-13.2.0 (Yasm-1.3.0-GCCcore-13.2.0.eb)
* x265/3.5-GCCcore-13.2.0 (x265-3.5-GCCcore-13.2.0.eb)
* MDI/1.4.29-gompi-2023b (MDI-1.4.29-gompi-2023b.eb)
* tbb/2021.13.0-GCCcore-13.2.0 (tbb-2021.13.0-GCCcore-13.2.0.eb)
* SDL2/2.28.5-GCCcore-13.2.0 (SDL2-2.28.5-GCCcore-13.2.0.eb)
* FFmpeg/6.0-GCCcore-13.2.0 (FFmpeg-6.0-GCCcore-13.2.0.eb)
* ScaFaCoS/1.0.4-foss-2023b (ScaFaCoS-1.0.4-foss-2023b.eb)
* PLUMED/2.9.2-foss-2023b (PLUMED-2.9.2-foss-2023b.eb)
* VTK/9.3.0-foss-2023b (VTK-9.3.0-foss-2023b.eb)
* LAMMPS/29Aug2024-foss-2023b-kokkos (LAMMPS-29Aug2024-foss-2023b-kokkos.eb)

```

PR #992, opened 2025-03-31
{2023.06}[2023b,grace] apps originally built with EB 4.9.3

trz42:
This PR adds GROMACS and LAMMPS. It needs to use `--from-commit` because PRs for easyconfigs were only merged after EB release 4.9.4. For LAMMPS it also uses a commit for an updated easyblock that strictly speaking may not be necessary because the PR was targeting building for Sapphire Rapids.

### DIFF
--- a/easystacks/software.eessi.io/2023.06/grace/eessi-2023.06-eb-4.9.4-2023b.yml
+++ b/easystacks/software.eessi.io/2023.06/grace/eessi-2023.06-eb-4.9.4-2023b.yml
@@ -149,3 +149,19 @@ easyconfigs:
 #        # we use a commit from the Brunsli PR here to get rid of the Highway dependency
 #        from-commit: 1736a123b1685836452587a5c51793257570bb2d
   - R-bundle-CRAN-2024.06-foss-2023b.eb
+# from here on easyconfigs were originally built with EB 4.9.3
+  - GROMACS-2024.3-foss-2023b.eb:
+      options:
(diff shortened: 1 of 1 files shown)


trz42:
Building on JURECA-DC may not work until some issue with the system has been resolved.

trz42:
bot: build instance:trz42-GH200-jr repository:eessi.io-2023.06-software architecture:aarch64/nvidia/grace exportvariable:SKIP_TESTS=yes

TopRichard (changes requested):
Shouldn't the commit `8e4d509711db42200c1f4e4cfc88bcef7aee80b7` be used for  GROMACS-2024.3-foss-2023b.eb ?

trz42:
> Shouldn't the commit `8e4d509711db42200c1f4e4cfc88bcef7aee80b7` be used for GROMACS-2024.3-foss-2023b.eb ?

Why that one? Because it's the merge commit? Technically no difference to the one I used (which is also the same that was used in #946

TopRichard:
> > Shouldn't the commit `8e4d509711db42200c1f4e4cfc88bcef7aee80b7` be used for GROMACS-2024.3-foss-2023b.eb ?
> 
> Why that one? Because it's the merge commit? Technically no difference to the one I used (which is also the same that was used in #946

Yes because it is the merge commit, but technically in this case there is no difference.

TopRichard (approved):
lgtm

TopRichard:
Tarball ingested, accessible

PR #994, opened 2025-04-02
Add zstd support for faster tarball creation or extraction in eessi_container.sh

casparvl:
When --resume and/or --save is used and zstd is available.

The builds in this PR show the benefit on a system that has zstd available. For SAMtools, the total time for the build job goes down from 12 to 4 minutes (tarballs for saving the container state of the build_step are something like 1.7GB for this build). For pmt, it goes down from 22 to 3.5 minutes (tarballs for saving the container state of the build step are around 3.2 GB for this build). Note also that the tarball size goes down (2.7 GB for pmt for the build_step), so `zstd` is not only faster, it also compresses better. Could also be nice for our file size footprint on the AWS storage.

### DIFF
--- a/eessi_container.sh
+++ b/eessi_container.sh
@@ -408,19 +408,6 @@ else
   echo "Using ${EESSI_HOST_STORAGE} as tmp directory (to resume session add '--resume ${EESSI_HOST_STORAGE}')."
 fi
 
-# if ${RESUME} is a file, unpack it into ${EESSI_HOST_STORAGE}
-if [[ ! -z ${RESUME} && -f ${RESUME} ]]; then
-  if [[ "${RESUME}" == *.tgz ]]; then
-    tar xf ${RESUME} -C ${EESSI_HOST_STORAGE}
-  # Add support for resuming from zstd-compressed tarballs
-  elif [[ "${RESUME}" == *.zst && -x "$(command -v zstd)" ]]; then
-    zstd -dc ${RESUME} | tar -xf - -C ${EESSI_HOST_STORAGE}
-  elif [[ "${RESUME}" == *.zst && ! -x "$(command -v zstd)" ]]; then
-    fatal_error "Trying to resume from tarball ${RESUME} which was compressed using zstd, but zstd command not found"
-  fi
-  echo "Resuming from previous run using temporary storage ${RESUME} unpacked into ${EESSI_HOST_STORAGE}"
-fi
-
 # if ${RESUME} is a file (assume a tgz), unpack it into ${EESSI_HOST_STORAGE}
 if [[ ! -z ${RESUME} && -f ${RESUME} ]]; then
   tar xf ${RESUME} -C ${EESSI_HOST_STORAGE}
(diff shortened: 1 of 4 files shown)


casparvl:
Hmmm, I'm not sure if I need more of the changes from #903 to make things work on Snellius. Nor if I can actually build some CPU-only stuff there. But let's try :)

bot: build instance:eessi-bot-surf repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90

casparvl:
bot: build instance:eessi-bot-surf repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90

casparvl:
bot: build instance:eessi-bot-surf repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90

casparvl:
bot: build instance:eessi-bot-surf repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90

casparvl:
bot: build instance:eessi-bot-surf repo:eessi.io-2023.06-software arch:zen4 accel:nvidia/cc90

trz42 (commented):
Nice! I'll give it a try later today.

eessi_container.sh
```diff
@@ -865,17 +878,30 @@ if [[ ! -z ${SAVE} ]]; then
...
   #   of these aspects to where the script is used
+  # Compression with zlib may be quite slow. On some systems, the pipeline takes ~20 mins for a 2 min build because of this.
+  # Check if zstd is present for faster compression and decompression
   if [[ -d ${SAVE} ]]; then
     # assume SAVE is name of a directory to which tarball shall be written to
     #   name format: tmp_storage-{TIMESTAMP}.tgz
     ts=$(date +%s)
-    TGZ=${SAVE}/tmp_storage-${ts}.tgz
+    if [[ -x "$(command -v zstd)" ]]; then
+      TARBALL=${SAVE}/tmp_storage-${ts}.zst
```
boegel, line 888: Is `.zst` the standard extensions for zstd-compressed tarballs?
I've never seen that anywhere...

eessi_container.sh
```diff
@@ -408,6 +408,19 @@ else
   echo "Using ${EESSI_HOST_STORAGE} as tmp directory (to resume session add '--resume ${EESSI_HOST_STORAGE}')."
 fi
 
+# if ${RESUME} is a file, unpack it into ${EESSI_HOST_STORAGE}
+if [[ ! -z ${RESUME} && -f ${RESUME} ]]; then
+  if [[ "${RESUME}" == *.tgz ]]; then
+    tar xf ${RESUME} -C ${EESSI_HOST_STORAGE}
+  # Add support for resuming from zstd-compressed tarballs
+  elif [[ "${RESUME}" == *.zst && -x "$(command -v zstd)" ]]; then
```
boegel, line 416: Why not check once if `zstd` is supported, define a constant, and use that everywhere:
```shell
if [[ -x $(command -v zstd) ]]; then
    USE_ZSTD=true
else
    USE_ZSTD=false
fi

casparvl: well, you can, but then you have to check if [[ `USE_ZSTD` == 'true' ]] (remember, this is bash, not python, there is no boolean in bash.

Since you have to do a test anyway, the current is explicit, equally short, and execution time of `command -v` is negligible. No point in storing it in an environment variable, then having to remember what the value to check for is (true or True or whatever).

eessi_container.sh
casparvl: yes, it is. https://en.wikipedia.org/wiki/Zstd

I believe if your tar is build with zstd support, you can do -xaf to inflate and it will infer from the extension that it needs to use zstd. But: I guess you could have zstd on your system and _not_ have your tar compiled with zstd support, so my current implementation is more portable.

trz42 (approved):
Works as intended.

Tested with tarballs of about 14 GB size. Resuming with gzip2 took about 3-4 minutes, saving about 12-13 minutes. With `zstd` resuming took about 2 minutes, saving about 2-3 minutes.

Nice! 👍

PR #1001, opened 2025-04-04
Update install_apptainer_ubuntu.sh to fix ci

ocaisa:


trz42 (approved):
Seems to work in CI again and looks much simpler than before.

PR #1023, opened 2025-04-12
experiment with checks api

trz42:
Let's see...


trz42:
Should experiment in personal repo first.

PR #1035, opened 2025-04-22
[WIP] Scripts for generating easystacks based on an existing stack

bedroge:
See https://gitlab.com/eessi/support/-/issues/145.

The first script, `stack_to_json.sh`, allows you to generate a json file with all the build details of an existing stack (paths to easyconfigs, easyblocks, EB version used for the builds, etc), ordered by build time. It will output a list with items that look like:
```
[
  {
    "build_time": "2023-11-26T20:27:55",
    "build_duration_minutes": "1",
    "name": "EasyBuild",
    "version": "4.8.2",
    "easybuild": "4.9.4",
    "toolchain": "SYSTEM",
    "easyconfig": "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/EasyBuild/4.8.2/easybuild/reprod/EasyBuild-4.8.2.eb",
    "easyblocks": "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/EasyBuild/4.8.2/easybuild/reprod/easyblocks/*.py"
  }
]
```
You can adjust the json as required, e.g. in order to move rebuilt apps up in the list (to match the original build time of the application, which we cannot determine automatically at the moment).

The second script, `json_to_easystacks.sh`, can then be used to generate a sequence of easystack files based on the json file, named like `001-eb-X.Y.Z.yml`.

### DIFF
--- a/scripts/generate_easystacks_for_existing_stack/stack_to_json.sh
+++ b/scripts/generate_easystacks_for_existing_stack/stack_to_json.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+DEBUG=0
+BASE_STACK=/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software
+EB_BOOTSTRAP=4.9.4
+
+declare -A gcc_to_foss=( ["12.2.0"]="2022b" ["12.3.0"]="2023a" ["13.2.0"]="2023b" )
+
+if [[ ! -d ${BASE_STACK} ]]; then
+  echo "The given base stack (${BASE_STACK}) is not a directory."
+  exit 1
+fi
+
+apps=$(find ${BASE_STACK} -mindepth 2 -maxdepth 2 -type d | head -n 3)
+
+json_output="["
+for app_dir in $apps; do
+  app_version=$(basename ${app_dir})
+  app_name=$(basename $(dirname ${app_dir}))
+
+  if [[ ${app_name} == "EESSI-extend" ]]; then
+    # Skip EESSI-extend, as it will be installed automatically.
+    continue
+  fi
+
+  easyblocks=${app_dir}/easybuild/reprod/easyblocks/*.py
+  easyconfig=${app_dir}/easybuild/reprod/${app_name}-${app_version}.eb
+  if [[ ! -f ${easyconfig} ]]; then
+    echo "ERROR: cannot find easyconfig for ${app_name}/${app_version}"
+  fi
+  # If rebuilds would not remove the original log file, we should take the build time from the first log.
+  # As we cannot guarantee that at the moment, we are cautious and use the last one.
+  log_file=$(ls -1 ${app_dir}/easybuild/easybuild-${app_name}*.log* | tail -n 1)
+  build_time_start=$(bzcat ${log_file} | head -n 1 | awk '{print $2 "T" $3}' | cut -d, -f1)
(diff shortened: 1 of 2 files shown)


casparvl (changes requested):
As a general comment: the script should get some proper option parsing and --help. E.g. something like https://github.com/casparvl/convenience_scripts/blob/30e9be6d89cf74c6118152b180552a551f163fbe/git_cleanup#L10

casparvl:
From what I've seen in https://github.com/EESSI/software-layer/pull/1037#issuecomment-2822320519 all the bootstrapping done by `load_easybuild_module.sh` and `load_eessi_extend_module.sh` makes it _really_ hard to enforce that e.g. EB 4.8.2 is installed with EB 4.8.2. Or even that EB 4.9.0 is installed with EB 4.8.2.

It means that even regardless of what you'd set as `EB_bootstrap` it is _very_ likely that it will just use the latest EB release anyway, as that's the one being installed in all the bootstrap procedures. 

I think your `EB_bootstrap` approach is reasonable, and we should keep it. We should just use the latest EB version as bootstrap. That will ensure that regardless of whether the bootstrap procedure from the `load_*` scripts is used, or whether it is an EB version actually based on the EB version in the EasyStack name, it'll be _the same_ (namely latest) EB version.

scripts/generate_easystacks_for_existing_stack/stack_to_json_parallel.sh
```diff
@@ -0,0 +1,90 @@
...
+(
+  app_version=$(basename "${app_dir}")
+  app_name=$(basename "$(dirname "${app_dir}")")
+
+  if [[ ${app_name} == "EESSI-extend" ]]; then
+    exit 0
+  fi
+
+  easyblocks=${app_dir}/easybuild/reprod/easyblocks/*.py
+  easyconfig=${app_dir}/easybuild/reprod/${app_name}-${app_version}.eb
```
bedroge, line 40: We shouldn't use the easyconfig from the `reprod` dir, that's causing too many issues with missing parameters/patches.
```suggestion
  easyconfig=${app_dir}/easybuild/${app_name}-${app_version}.eb
```

scripts/generate_easystacks_for_existing_stack/stack_to_json.sh
```diff
@@ -0,0 +1,84 @@
...
+  app_version=$(basename ${app_dir})
+  app_name=$(basename $(dirname ${app_dir}))
+
+  if [[ ${app_name} == "EESSI-extend" ]]; then
+    # Skip EESSI-extend, as it will be installed automatically.
+    continue
+  fi
+
+  easyblocks=${app_dir}/easybuild/reprod/easyblocks/*.py
+  easyconfig=${app_dir}/easybuild/reprod/${app_name}-${app_version}.eb
```
bedroge, line 27: We shouldn't use the easyconfig from the reprod dir, that's causing too many issues with missing parameters/patches.

```suggestion
  easyconfig=${app_dir}/easybuild/${app_name}-${app_version}.eb
```

casparvl:
I'm now in the stage where I have to play catch-up with the rest of the architectures after we've ingested all the initially generated EasyStacks. My main issue is that I essentially have to regenerate the easystacks, figure out from the last few what is still missing, then separate those out. It's not a big deal, but it could be nice if we can improve the script to _also_ look at e.g. the `icelake` prefix and then generate EasyStacks _only_ for those. Then, if we could even indicate what _number_ it should use to start numbering new EasyStacks... that'd make it super smooth and easy.

Issue #845, opened 2025-01-10
supporting Arm Neoverse V2 CPUs: NVIDIA Grace, AWS Graviton 4, Google Axion


boegel:
While looking into implementing support in `archdetect` for detecting the Neoverse V2-based Graviton 4, I noticed that the CPUs that implement this microarchitecture only partially overlap (based on https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html):

| instructions | Amazon Graviton 4 | NVIDIA Grace | Google Axion |
| ------------- | ------------- | ------------- | ------------- |
| `sve2` | x | x | x |
| `paca` | x |  | x |
| `pacg`| x |  | x |
| `rng` | x |  | x |
| `sm3` | | x | x |
| `sm4` | | x | x |
| `svesm4` | | x | x |
| `ssbs` | x | x | |

Google Axion not supported `ssbs` is particularly interesting, since that meanw that our `aarch64/neoverse_v1` installations may not even work there...

ocaisa:
Seems to be possible: https://github.com/pkgw/elfx86exts (says it works for x86 and ARM)

ocaisa:
We could inspect the instructions required by the software, and have CI complain if it goes outside the common options.

boegel:
> Seems to be possible: https://github.com/pkgw/elfx86exts (says it works for x86 and ARM)

Step 0:
* https://github.com/easybuilders/easybuild-easyconfigs/pull/22145

ocaisa:
Sample output from `elfx86exts`:
```bash
$ elfx86exts $(which elfx86exts)
File format and CPU architecture: Elf, X86_64
MODE64 (call)
CMOV (cmova)
BMI2 (mulx)
AVX (vpxor)
NOVLX (vpxor)
AVX2 (vpbroadcastq)
BMI (tzcnt)
SSE2 (pause)
SSE1 (xorps)
BWI (vpbroadcastb)
VLX (vpbroadcastb)
AVX512 (kortestw)
DQI (kshiftrb)
Instruction set extensions used: AVX, AVX2, AVX512, BMI, BMI2, BWI, CMOV, DQI, MODE64, NOVLX, SSE1, SSE2, VLX
CPU Generation: Unknown
```

ocaisa:
Instructions set extensions cover a range of assembly instructions relevant to different categories (security, performance, AI/ML, Crypography,...). We could consider keeping a catalogue of what assembly instructions are required for each software package.

[...]

PR #1042, opened 2025-04-23
Add support in archdetect for detecting NVIDIA/Grace

TopRichard:
This PR should not be merged until NVIDIA/Grace stack is ready

### DIFF
--- a/init/arch_specs/eessi_arch_arm.spec
+++ b/init/arch_specs/eessi_arch_arm.spec
@@ -7,3 +7,4 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
+"aarch64/nvidia/grace"	"0x41"		"svesm4"		# NVIDIA Grace
--- a/init/arch_specs/eessi_arch_arm.spec
+++ b/init/arch_specs/eessi_arch_arm.spec
@@ -7,4 +7,4 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
-"aarch64/nvidia/grace"	"0x41"		"svesm4"		# NVIDIA Grace
+"aarch64/nvidia/grace"	"0x41"		"ssbs svesm4"		# NVIDIA Grace
--- a/init/arch_specs/eessi_arch_arm.spec
+++ b/init/arch_specs/eessi_arch_arm.spec
@@ -7,4 +7,5 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
-"aarch64/nvidia/grace"	"0x41"		"ssbs svesm4"		# NVIDIA Grace
+"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
+"aarch64/nvidia/grace"	"0x41"		"sve2 sm3 sm4 svesm4"		# NVIDIA Grace
--- a/init/arch_specs/eessi_arch_arm.spec
+++ b/init/arch_specs/eessi_arch_arm.spec
@@ -1,5 +1,6 @@
 # ARM CPU architecture specifications (see https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html for guidance)
 # CPU implementers: 0x41 (ARM), 0x46 (Fujitsu) - also see https://github.com/hrw/arm-socs-table/blob/main/data/socs.yml
+# Flipping the order of nvidia/grace and google/axion results in nvidia/grace being the "top" on Axion systems
 
 # Software path in EESSI 	| 'Vendor ID' or 'CPU implementer' 	| List of defining CPU features
 "aarch64/a64fx"		"0x46"		"asimdhp sve"		# Fujitsu A64FX
--- a/init/arch_specs/eessi_arch_arm.spec
+++ b/init/arch_specs/eessi_arch_arm.spec
@@ -7,5 +7,5 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
-"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
 "aarch64/nvidia/grace"	"0x41"		"sve2 sm3 sm4 svesm4"		# NVIDIA Grace
+"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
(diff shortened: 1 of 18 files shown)


init/arch_specs/eessi_arch_arm.spec
```diff
@@ -7,3 +7,4 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
+"aarch64/nvidia/grace"	"0x41"		"svesm4"		# NVIDIA Grace
```
boegel, line 10: This doesn't uniquely identify NVIDIA Grace, could also be Google Axion, see https://github.com/EESSI/software-layer/issues/845

init/arch_specs/eessi_arch_arm.spec
```diff
@@ -7,3 +7,4 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
+"aarch64/nvidia/grace"	"0x41"		"ssbs svesm4"		# NVIDIA Grace
```
boegel, line 10: @TopRichard Based on the `/proc/cpuinfo` that @trz42 grabbed from a Google Axion system (see https://github.com/EESSI/software-layer/issues/845#issuecomment-2827629221), `ssbs` is not a discriminator between Grace & Axion...

One way to deal with that would be to add another entry there, that's considered first, to first check if it's a Google Axion, and if not fall through to the Grace case?
It looks like only Axion support `rng`?

So something like:

```
"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
"aarch64/nvidia/grace"	"0x41"		"sve2 sm3 sm4 svesm4"		# NVIDIA Grace

Comment (TopRichard) [reply init/arch_specs/eessi_arch_arm.spec]: Suggestion applied

Comment (boegel) [reply init/arch_specs/eessi_arch_arm.spec]: This should work correctly on a Grace system now (it's not detect it as `axion`, and fall through to the last entry and find a hit there), but on an Axion system is doesn't actually fully work correctly yet with this, since it'll "prefer" `grace` over `axion`:

```
$ init/eessi_archdetect.sh cpupath
aarch64/nvidia/grace

$ init/eessi_archdetect.sh -a cpupath
aarch64/nvidia/grace:aarch64/google/axion:aarch64/neoverse_v1:aarch64/neoverse_n1:aarch64/generic
```

So to make it correct on Axion we need a bit of work in `archdetect` itself...
I think that for Grace, we need to be able to express "has `sve2 sm3 sm4 svesm4`, does not have `rng`".

With that in mind, we should probably just remove the `axion` entry again here, and get this merged, since at least the detection on `grace` will be correct.

We can deal with correctly detecting `axion` in a follow-up PR, since that's not really relevant yet at this point.

File: init/arch_specs/eessi_arch_arm.spec [hunk 3]
```diff
@@ -7,3 +7,5 @@
 "aarch64/neoverse_n1"	"0x41"		"asimddp"		# AWS Graviton2
 "aarch64/neoverse_v1"	"ARM"		"asimddp svei8mm"
 "aarch64/neoverse_v1"	"0x41"		"asimddp svei8mm"	# AWS Graviton3
+"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
+"aarch64/nvidia/grace"	"0x41"		"sve2 sm3 sm4 svesm4"		# NVIDIA Grace
```
boegel, line 11: @TopRichard My earlier suggestion was incorrect, since this leads to incorrect order on Axion systems, resulting in `nvidia/grace` being the "top" result while it should be `google/axion`.

Just flipping these is enough to make it correct, since that's more specific (also has `rng`):

```suggestion
"aarch64/nvidia/grace"	"0x41"		"sve2 sm3 sm4 svesm4"		# NVIDIA Grace
"aarch64/google/axion"	"0x41"		"sve2 rng sm3 sm4 svesm4"	# Google Axion
```

```
$ init/eessi_archdetect.sh -a cpupath
aarch64/google/axion:aarch64/nvidia/grace:aarch64/neoverse_v1:aarch64/neoverse_n1:aarch64/generic
```

So basically this means that binaries optimized for Grace can also be used on Axion, but will be "suboptimal" (since `rng` instructions will not be used).

To make sure we're not overlooking something, we should also add tests for `google/axion`, using the `/proc/cpuinfo` output available in https://github.com/EESSI/software-layer/issues/845#issuecomment-2827629221

boegel (changes requested):

init/arch_specs/eessi_arch_arm.spec
boegel: Nevermind, it's actually much simpler than that: Axion is a superset of Grace, since it has `rng` extra on top of everything that Grace has (incl. `ssbs`)

init/arch_specs/eessi_arch_arm.spec
trz42: Might be good to add a comment, why the order is important. There may be other ways to implement this, but they would require significantly more code.

init/arch_specs/eessi_arch_arm.spec
```diff
@@ -1,9 +1,12 @@
 # ARM CPU architecture specifications (see https://gpages.juszkiewicz.com.pl/arm-socs-table/arm-socs.html for guidance)
 # CPU implementers: 0x41 (ARM), 0x46 (Fujitsu) - also see https://github.com/hrw/arm-socs-table/blob/main/data/socs.yml
+# Flipping the order of nvidia/grace and google/axion results in nvidia/grace being the "top" on Axion systems
```
boegel, line 3: @TopRichard I don't think we're doing anything special here for either grace or Axion.

When CPU target A is a subset of CPU target B, then A should be listed before B to ensure a correct order produced by `archdetect`.

That's also the case for `neoverse_n1` vs `neoverse_v1`, for example.

So maybe this should be worded a bit more generally, to point out the importance of the order below without making it specific to particular CPU targets?

boegel (changes requested):

boegel (changes requested):
@TopRichard The extra test case for `nvidia/grace` is only used when the corresponding GitHub Actions workflow is also updated.

Also, it's worth also adding a test case for `google/axion`.

Both tackled in https://github.com/TopRichard/software-layer/pull/17

boegel (approved):
Looks good to go for me now.

@trz42 Can you double check, and perhaps manually test on both `grace` and `axion`?
Should be fine though, since it's verified through the extra test cases.

We'll also need to trigger a deploy before merging.

trz42:
> Looks good to go for me now.
> 
> @trz42 Can you double check, and perhaps manually test on both `grace` and `axion`? Should be fine though, since it's verified through the extra test cases.
> 
> We'll also need to trigger a deploy before merging.

Works on Google Axion
```bash
[@google-axion software-layer]$ cat /proc/cpuinfo | awk '/^processor/{p++} p==1' && init/eessi_archdetect.sh -a cpupath
processor	: 0
BogoMIPS	: 2000.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb dcpodp sve2 sveaes svepmull svebitperm svesha3 svesm4 flagm2 frint svei8mm svebf16 i8mm bf16 dgh rng
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd4f
CPU revision	: 1

aarch64/google/axion:aarch64/nvidia/grace:aarch64/neoverse_v1:aarch64/neoverse_n1:aarch64/generic
```
and also on NVIDIA Grace/Hopper
```bash
[jrc0900 software-layer]$ cat /proc/cpuinfo | awk '/^processor/{p++} p==1' && init/eessi_archdetect.sh -a cpupath
processor	: 0
BogoMIPS	: 2000.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb dcpodp sve2 sveaes svepmull svebitperm svesha3 svesm4 flagm2 frint svei8mm svebf16 i8mm bf16 dgh
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd4f
CPU revision	: 0

aarch64/nvidia/grace:aarch64/neoverse_v1:aarch64/neoverse_n1:aarch64/generic
```

TopRichard:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:x86_64/generic

bedroge (approved):
Note that the tarball includes `eb_hooks.py` as we forgot to deploy that in #1046.

bedroge:
Tarball has been ingested.

PR #1043, opened 2025-04-24
{2023.06}[NVHPC/25.1-CUDA-12.6] add hook for nvhpc

adammccartney:
This adds a pre_configure_hook for NVHPC. It performs some search and replace operations on the "localrc" file used by NVHPC to detect information about the system. In particular it points the sysroot flag at the eessi eprefix variable, and appends two variables definitions about where to look for system libraries.

The content of the hook is extracted from:
https://github.com/ComputeCanada/easybuild-computecanada-config/blob/main/2023/cc_hooks.py#L544-L547

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -744,6 +744,26 @@ def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
     else:
         raise EasyBuildError("LAMMPS-specific hook triggered for non-LAMMPS easyconfig?!")
 
+def pre_configure_hook_nvhpc_x86_64(self, *args, **kwargs):
+    """
+    pre-configure hook for nvhpc
+    - search and replace operations in the ec dict
+    """
+    if self.name == "NVHPC":
+        eprefix = get_eessi_envvar("EPREFIX")
+        new_opts = f'''installdir=%(installdir)s/Linux_x86_64/%(version)s
+            EPREFIX={eprefix}
+            EBROOTGENTOO={eprefix}/usr
+            sed -i "s@\(set LDSO=.*\);@\\1 --sysroot=$EPREFIX;@" $installdir/compilers/bin/localrc
+            echo "set DEFLIBDIR=$EBROOTGENTOO/lib64;" >> $installdir/compilers/bin/localrc
+            echo "set DEFSTDOBJDIR=$EBROOTGENTOO/lib64;" >> $installdir/compilers/bin/localrc'''
+        old_opts = self.cfg['postinstallcmds']
+        self.cfg.update('postinstallcmds', new_opts)
+        print_msg("Updated postinstall cmds from %s to %s", old_opts, new_opts)
+    else:
+        raise EasyBuildError("NVHPC-specific hook triggered for non-NVHPC easyconfig?!")
+
+
 
 def pre_test_hook(self, *args, **kwargs):
     """Main pre-test hook: trigger custom functions based on software name."""
@@ -1218,6 +1238,7 @@ PRE_CONFIGURE_HOOKS = {
     'GROMACS': pre_configure_hook_gromacs,
     'libfabric': pre_configure_hook_libfabric_disable_psm3_x86_64_generic,
     'MetaBAT': pre_configure_hook_metabat_filtered_zlib_dep,
+    'NVHPC': pre_configure_hook_nvhpc_x86_64,
     'OpenBLAS': pre_configure_hook_openblas_optarch_generic,
     'WRF': pre_configure_hook_wrf_aarch64,
     'LAMMPS': pre_configure_hook_LAMMPS_zen4,


hvelab:
Hi @adammccartney , 

As in the PR there is no build generate with its corresponding tests, could you kindly share the steps you did to test this so we can also test and reproduce it? 

Thank you!

eb_hooks.py
```diff
@@ -744,6 +744,26 @@ def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
...
+def pre_configure_hook_nvhpc_x86_64(self, *args, **kwargs):
+    """
+    pre-configure hook for nvhpc
+    - search and replace operations in the ec dict
+    """
+    if self.name == "NVHPC":
+        eprefix = get_eessi_envvar("EPREFIX")
+        new_opts = f'''installdir=%(installdir)s/Linux_x86_64/%(version)s
+            EPREFIX={eprefix}
+            EBROOTGENTOO={eprefix}/usr
```
ocaisa, line 756: The naming here is misleading, while Compute Canada have a Gentoo module, we do not. I think you could get away without setting it at all and just replace the usage below with `$EPREFIX/usr/lib64`

eb_hooks.py
```diff
@@ -744,6 +744,26 @@ def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
     else:
         raise EasyBuildError("LAMMPS-specific hook triggered for non-LAMMPS easyconfig?!")
 
+def pre_configure_hook_nvhpc_x86_64(self, *args, **kwargs):
+    """
+    pre-configure hook for nvhpc
+    - search and replace operations in the ec dict
+    """
+    if self.name == "NVHPC":
```
ocaisa, line 752: This is good for now, but there is quite a bit of discussion right now about changing the naming in the EasyBuild context. We are currently naming the _compilers only_ `NVHPC` but we should perhaps be defining a toolchain hierarchy for `NVHPC` since they also contain MPI and some math libraries. This may lead to the compilers being called something like `nvidia_compilers` or `NVHPC` becoming a fatter toolchain.

ocaisa: You're also going to need some versioning clauses here for things you have actually tested.

This is definitely a corner case at present as EESSI itself will not ship his toolchain (currently), and I wonder if we should not allow an environment variable to force the use of the hook, something like
```suggestion
    if self.name == "NVHPC":
        force_nvhpc_hook = 'EESSI_FORCE_NVHPC_HOOK'
        if self.version in [] or os.getenv(force_nvhpc_hook, False):
            ...
        else:
            print_msg(f"Not using exisiting hook for {self.name}/{self.version}, if you wish to force this please set the envvar {force_nvhpc_hook}")
```

adammccartney:
> Hi @adammccartney ,
> 
> As in the PR there is no build generate with its corresponding tests, could you kindly share the steps you did to test this so we can also test and reproduce it?
> 
> Thank you!

Sure, would be happy to. Would you mind giving a few points of guidance? Let me know what would be useful to see.
I haven't had the time to look into ReFrame at all yet, so there are no tests (yet) apart from the sanity checks in the EasyBuild. The build last week was done in the eessi-container from the "software layer" repo, which was slightly adapted to suit our own build environment.  The build command looks like the following:

```
#!/bin/bash

project_root="$(realpath $(dirname $(dirname $(dirname $(dirname $BASH_SOURCE)))))"

eb "${project_root}/easyconfigs/2025/NVHPC-25.1-CUDA-12.6.0.eb" \
    -r --cuda-compute-capabilities=9.0 \
    --configfiles="${project_root}/easybuild-asc-config/2025/config.cfg" \
    --hooks="${project_root}/easybuild-asc-config/2025/eb_hooks.py"
```

As you can see we are referencing an explicit config for easybuild and I think the easyconfig for NVHPC is slightly adapted to include the "accept-eula" variable or whatver. I'll backport this to a "vanilla" eessi-extend environment today that can be used to install stuff on host-injections. I guess it would be useful to have a command that can be run in the standard container started by `eessi_container.sh` ?

eb_hooks.py
```diff
@@ -744,6 +744,25 @@ def pre_configure_hook_LAMMPS_zen4(self, *args, **kwargs):
...
+        eprefix = get_eessi_envvar("EPREFIX")
+        new_opts = f'''installdir=%(installdir)s/Linux_x86_64/%(version)s
+            EPREFIX={eprefix}
+            sed -i "s@\(set LDSO=.*\);@\\1 --sysroot=$EPREFIX;@" $installdir/compilers/bin/localrc
+            echo "set DEFLIBDIR=$EPREFIX/usr/lib64;" >> $installdir/compilers/bin/localrc
+            echo "set DEFSTDOBJDIR=$EPREFIX/usr/lib64;" >> $installdir/compilers/bin/localrc'''
```
ocaisa, line 758: Thinking about this a bit more, the logic here could be added directly to the NVHPC easyblock (and used conditionally based on whether the EB build option `--sysroot` is set).

adammccartney:
So, interestingly the sanity check now fails if I try to build this directly on a compute node (x86_64/amd/zen4 is the architecture by the way) . The initial build was done in the eessi container as I mentioned, set up to use the `EESSI_PROJECT_INSTALL` variable pointing at a writeable `/cvmfs/software.asc.ac.at` directory. 
The build now fails when I try to use `EESSI_SITE_INSTALL`. Maybe there is something leaking in via the ld cache on the host as was previously observed. Makes me wonder about how usable the compiler is if we load it from the custom cvmfs repo...

adammccartney:
bot: build inst:eessi-bot-mc-azure arch:x86_64/amd/zen4 repo:eessi.io-2023.06-software accelerator:nvidia/cc90

PR #502, opened 2024-03-14
{2023.06}[foss/2023a] Z3 4.12.2

ocaisa:
Via https://github.com/easybuilders/easybuild-easyconfigs/pull/20050 , the different Z3 easyconfigs have been consolidated. This means the version used as a dependency for PyTorch has been "corrected" so we need to dance around that for our CI purposes

```
 1 out of 14 required modules missing:

* Z3/4.12.2-GCCcore-12.3.0 (Z3-4.12.2-GCCcore-12.3.0.eb)
```

PR #1045, opened 2025-04-24
{2023.06}[2023a,grace] Z3 v4.12.2

TopRichard:
This is an outdated version of the Z3-4.12.2-GCCcore-12.3.0 easyconfig, before unifying Z3 4.12.2 easyconfigs into a single one with Python bindings in https://github.com/easybuilders/easybuild-easyconfigs/pull/20050/files , adding it to keep sync with the other stacks

### DIFF
--- a/easystacks/software.eessi.io/2023.06/grace/eessi-2023.06-eb-4.9.0-2023a.yml
+++ b/easystacks/software.eessi.io/2023.06/grace/eessi-2023.06-eb-4.9.0-2023a.yml
@@ -1,3 +1,4 @@
 easyconfigs:
 #This is an outdated version of the Z3-4.12.2-GCCcore-12.3.0 easyconfig, adding it to keep sync with the other stacks
+#see https://github.com/easybuilders/easybuild-easyconfigs/pull/20050
   - Z3-4.12.2-GCCcore-12.3.0-Python-3.11.3.eb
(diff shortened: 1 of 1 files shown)


TopRichard:
bot: build inst:rt-Grace-jr arch:aarch64/nvidia/grace repo:eessi.io-2023.06-software

boegel:
@TopRichard In what sense is this Z3 outdated? Was it updated in a more recent version of EasyBuild?
Would be good to be more clear on that I think

TopRichard:
> @TopRichard In what sense is this Z3 outdated? Was it updated in a more recent version of EasyBuild? Would be good to be more clear on that I think

Clarification added

boegel:
> > @TopRichard In what sense is this Z3 outdated? Was it updated in a more recent version of EasyBuild? Would be good to be more clear on that I think
> 
> Clarification added

Thanks!

Just to clear: for the other CPU targets, Z3 was never rebuilt to pick up the changes made in https://github.com/easybuilders/easybuild-easyconfigs/pull/20050?

Should we open an issue to do that when the time is right?

boegel (approved):

TopRichard:
> > > @TopRichard In what sense is this Z3 outdated? Was it updated in a more recent version of EasyBuild? Would be good to be more clear on that I think
> > 
> > 
> > Clarification added
> 
> Thanks!
> 
> Just to clear: for the other CPU targets, Z3 was never rebuilt to pick up the changes made in [easybuilders/easybuild-easyconfigs#20050](https://github.com/easybuilders/easybuild-easyconfigs/pull/20050)?
> 
> Should we open an issue to do that when the time is right?

It was done in: #502

boegel:
@TopRichard Hmm: `transfer of eessi-2023.06-software-linux-aarch64-nvidia-grace-1745483027.tar.gz to S3 bucket failed`

boegel:
staging PR merged, so merging this PR too

PR #1046, opened 2025-04-24
use Arm GitHub Actions runner when checking for missing installations for `aarch64/*` CPU targets

boegel:


bedroge (approved):
Nice! We may run into a similar issue again when we'll add RISC-V for a next version, but we van deal with that then (maybe we could use Qemu?).

PR #1071, opened 2025-05-04
{2023.06}[icelake,cascadelake] EasyStacks batch 15: 058 - 059a

casparvl:
This can be synced with upstream once https://github.com/EESSI/software-layer/pull/1070 is merged. Then, it can be built.

PR #1072, opened 2025-05-04
{2023.06}[icelake,cascadelake] EasyStacks batch 16: 060 - 068

casparvl:
This can be synced with upstream once https://github.com/EESSI/software-layer/pull/1071 is merged. Then, it can be built.

### DIFF
--- a/easystacks/software.eessi.io/2023.06/icelake_cclake/060-eb-4.9.3.yml
+++ b/easystacks/software.eessi.io/2023.06/icelake_cclake/060-eb-4.9.3.yml
@@ -1,0 +1,5 @@
+# 060-eb-4.9.3.yml: total build duration = 42 minutes
+easyconfigs:
+  - /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/haswell/software/GROMACS/2024.3-foss-2023b/easybuild/GROMACS-2024.3-foss-2023b.eb:
+      options:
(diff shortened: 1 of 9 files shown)


bedroge:
I don't think this depends on ReFrame (which still has to be built in #1080, split off from #1071), so it should be okay to build this one already.

bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:icelake
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:cascadelake

bedroge:
For cascade lake, one GROMACS 2024.4 test failed:
```
The following tests FAILED:
          2 - GmxapiMpiTests (Failed)
```

bedroge:
And we may want to check if LAMMPS picked the right Kokkos CPU. Looks like it used `SKX` (skylake + avx512) for both, while it seems that it could have used ICL for Icelake:
https://github.com/lammps/lammps/blob/stable_29Aug2024/lib/kokkos/cmake/kokkos_arch.cmake#L55

bedroge:
I haven't really looked into the failing GROMACS test, but I think we've seen before that it may just work if we try it again, so let's do another build overnight...

bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:cascadelake

bedroge (approved):

PR #1087, opened 2025-05-20
Add CUDA software check to stack comparison CI

ocaisa:


ocaisa:
This will have to remain a Draft PR until the CUDA stacks are in sync

ocaisa:
@casparvl This is now working, but it is comparing against `generic` (which is not _really_ a problem since they are all supposed to be the same)

ocaisa:
Retriggering CI

ocaisa:
Retriggering this to see where we are

laraPPr:
It needs this one to be merged first https://github.com/EESSI/software-layer-scripts/pull/23

laraPPr:
Should i move this one then to this pr because it also need this but I have not received answer to my question in their? https://github.com/EESSI/software-layer-scripts/pull/23

PR #1086, opened 2025-05-20
{2023.06}[2023a,a64fx] PyTorch 2.1.2 originally built with EB 4.9.0

trz42:
Adds remaining app - PyTorch - that was originally built with EB 4.9.0

Other apps originally built with EB 4.9.0 were added via #1038 and #1040

PR #1094, opened 2025-05-27
run the `check_missing_installations.sh` script with the EB version from the easystack filename

bedroge:
By not specifying the version, it would always use the latest EB version. That could lead to annoying issues, for instance the issue observed in #1086: it started running the checks for some targets with EB 5.1.0 (which just got ingested), and in this version Java 21 was updated to point to Java 21.0.7 instead of 21.0.5. By always sticking to the EB version that's encoded in the easystack filename, this should not happen. We already do the same in `EESSI-installsoftware.sh`, see https://github.com/EESSI/software-layer/blob/2023.06-software.eessi.io/EESSI-install-software.sh#L362.

### DIFF
--- a/.github/workflows/test-software.eessi.io.yml
+++ b/.github/workflows/test-software.eessi.io.yml
@@ -74,7 +74,7 @@ jobs:
               env | grep ^EESSI | sort
 
               # first check the CPU-only builds for this CPU target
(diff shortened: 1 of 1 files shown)


ocaisa (dismissed):
LGTM

PR #1093, opened 2025-05-27
Added hook to restore filtered deps for LLVM builds

Crivella:
Possible fix for 

- #1092 

A possible alternative fix could be:

- #1095

PR #1095, opened 2025-05-27
Adding general way to handle filtered deps when installing software

Crivella:
Right now every time a software that relies on calling `get_software_root` for a filtered deps is called requires and ad-hoc fix via an hook.

This is a proposed way to automatically point the respective `$EBROOT...` variable to the compat layer for every filtered deps.

## TODO / Ideas

- Would need also to remove the customization done in other hooks
- Possibly recognize only deps filtered specifically by EESSI instead of relying on `build_option('filter_deps')
- Alternatively or together to previous point, implement better logic for `filter_deps` as in [framework](https://github.com/easybuilders/easybuild-framework/blob/8e6079e329a5a3b977203b3012771a27fa073578/easybuild/framework/easyconfig/easyconfig.py#L1153-L1179)
- Could/Should this be performed at an earlier step the `pre_configure`?

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -661,27 +661,6 @@ def pre_configure_hook_gromacs(self, *args, **kwargs):
         raise EasyBuildError("GROMACS-specific hook triggered for non-GROMACS easyconfig?!")
 
 
-def pre_configure_hook_llvm(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


boegel:
This is a pretty big hammer, I would rather see us handle this case by case, as is being done in #1093

PR #1099, opened 2025-05-28
fix cudnn hook for non eessi installation

pescobar:
while doing local installations we got this error installing cuddn

```
== 2025-05-28 11:41:04,652 hooks.py:249 INFO Running 'post_postproc_hook' hook function (args: [<easybuild.easyblocks.cudnn.EB_cuDNN object at 0x7fe233fd2450>], keyword args: {})...
== 2025-05-28 11:41:04,816 build_log.py:226 ERROR EasyBuild encountered an error (at easybuild/tools/build_log.py:166 in caller_info): cuDNN-specific hook triggered for non-cuDNN easyconfig?! (at easybuild/eb_hooks.py:1073 in post_postproc_cudnn)
```

Reviewing the hook we noticed that if line `if self.name == 'cuDNN' and eessi_installation:`  returns `False` (like in our case) then you always end up in `raise EasyBuildError("cuDNN-specific hook triggered for non-cuDNN easyconfig?!")`

I modified the cudnn hook taking as reference the cuda hook above it so the logic is 

```
if eessi_install
    run_hook
else
    print_msg(f"EESSI hook to respect cuDDN license not triggered for installation path {self.installdir}")
```

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -390,7 +390,7 @@ def parse_hook_freeimage_aarch64(ec, *args, **kwargs):
                 ec['toolchainopts'] = {}
             ec['toolchainopts']['pic'] = True
             ec['toolchainopts']['extra_cflags'] = '-DPNG_ARM_NEON_OPT=0'
-            print_msg("Changed toolchainopts for %s: %s", ec.name, ec['toolchainopts']) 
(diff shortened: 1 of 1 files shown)


ocaisa:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws arch:x86_64/amd/zen2

PR #1089, opened 2025-05-23
Fix EB4 related warning for hooks

ocaisa:
This removes the deprecation warning when using `EESSI-extend`:
```
== Running post-ready hook...

WARNING: Deprecated functionality, will no longer work in EasyBuild v6.0: Easyconfig parameter 'parallel' is deprecated, use 'max_parallel' or the parallel property instead.; see
https://docs.easybuild.io/deprecated-functionality/ for more information

== ... (took < 1 sec)
```

PR #1104, opened 2025-05-28
fix issue that parallel gets set to None with EB 5+

trz42:
Fixes an issue caused by the post_ready_hook trying to obtain the parallelism via

```
self.cfg['max_parallel'] 
```

which is not set by `EasyBlock.set_parallel`. The latter always (for EB 4 and EB 5) sets

```
self.cfg['parallel'] 
```
_or_
```
self.cfg.parallel 
```

### DIFF
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -133,10 +133,11 @@ def post_ready_hook(self, *args, **kwargs):
     # Check whether we have EasyBuild 4 or 5
     parallel_param = 'parallel'
     if EASYBUILD_VERSION >= '5':
-        # parallel_param = 'max_parallel'  # EasyBlock.set_parallel sets 'parallel'
-        parallel_param = 'parallel'
+        parallel_param = 'max_parallel'
     # get current parallelism setting
     parallel = self.cfg[parallel_param]
+    if parallel == None:
+        return  # self.cfg doesn't contain 'parallel' or 'max_parallel'
     if parallel == 1:
         return  # no need to limit if already using 1 core
 
--- a/eb_hooks.py
+++ b/eb_hooks.py
@@ -133,7 +133,8 @@ def post_ready_hook(self, *args, **kwargs):
     # Check whether we have EasyBuild 4 or 5
     parallel_param = 'parallel'
     if EASYBUILD_VERSION >= '5':
-        parallel_param = 'max_parallel'
+        # parallel_param = 'max_parallel'  # EasyBlock.set_parallel sets 'parallel'
+        parallel_param = 'parallel'
     # get current parallelism setting
     parallel = self.cfg[parallel_param]
     if parallel == 1:


eb_hooks.py
```diff
@@ -133,7 +133,8 @@ def post_ready_hook(self, *args, **kwargs):
     # Check whether we have EasyBuild 4 or 5
     parallel_param = 'parallel'
     if EASYBUILD_VERSION >= '5':
-        parallel_param = 'max_parallel'
+        # parallel_param = 'max_parallel'  # EasyBlock.set_parallel sets 'parallel'
+        parallel_param = 'parallel'
```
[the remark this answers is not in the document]
trz42: Is the warning really appearing for anyone or only when we adjust the parallelism in the post_ready_hook?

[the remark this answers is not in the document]
ocaisa: Why don't we just use the parallel property (I'm guessing that is `self.parallel`) then in the EB5 scenario?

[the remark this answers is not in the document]
ocaisa: > Is the warning really appearing for anyone or only when we adjust the parallelism in the post_ready_hook?

I think yes, I fixed it because it bothered me to see it all the time while preparing the CI tutorial

eb_hooks.py
```diff
@@ -136,6 +136,8 @@ def post_ready_hook(self, *args, **kwargs):
         parallel_param = 'max_parallel'
```
boegel, line 136: A better with EasyBuild 5.x would be to use `self.cfg.parallel` instead of `self.cfg['parallel']`

ocaisa: In general, I think the information gathering here is in the wrong place. You only need to do this if you plan to change the parallelism, so this whole block could be moved to when you need it, right before `if new_parallel != parallel:`

eb_hooks.py
```diff
@@ -136,6 +136,8 @@ def post_ready_hook(self, *args, **kwargs):
...
     # get current parallelism setting
     parallel = self.cfg[parallel_param]
+    if parallel == None:
+        return  # self.cfg doesn't contain 'parallel' or 'max_parallel'
```
ocaisa, line 140: This exits too early, it may be the case that this is `None` but you still want to limit the parallelism. What we really want is to call https://github.com/easybuilders/easybuild-framework/pull/3811/files#diff-8136c25c2706ba344aa622b3441a997a04037c1f5e6acadc006ae27f132554f4R1833 since that covers all cases.

That's an open framework PR though so I can be pragmatic. In general this query is triggered for _every_ easyconfig, but in reality you only need to know it if you plan on changing it. The query should happen inside the `if` block. In that case you'll only see the warning when it affects what you are trying  to build (right now you see it on every call to `eb`). I'm ok with reverting https://github.com/EESSI/software-layer/pull/1089 as long as the querying the parallelism is moved into the `if self.name in PARALLELISM_LIMITS:` block

I also think we need to do this properly  now, I was not familiar with the new parallel property but I have since had a look. In the case of EasyBuild 5 a different approach seems to be needed.

ocaisa: I think `getattr(self, 'parallel', self.cfg['parallel'])` will work for both scenarios without the need for a version check.

boegel:
This is now being fixed via https://github.com/EESSI/software-layer-scripts/pull/17, so let's close?

PR #1115, opened 2025-06-10
Allow for CUDA compute capability fallbacks when initialising EESSI

ocaisa:
This also expects/enforces that site installations follow the same structure as EESSI


ocaisa:
Switched to https://github.com/EESSI/software-layer-scripts/pull/5

PR #737, opened 2024-09-24
{2023.06}[foss/2023a] Clang v16.0.6

casparvl:
I've seen that it failed in this PR https://github.com/EESSI/software-layer/pull/637#issuecomment-2216870134 but that's been a while, so let me just retry.

Also: on our system we also build with RPATH, and have no issues installing this Clang version... It'll be interesting to see what might be different in EESSI.

PR #637, opened 2024-07-06
{2023.06}[foss/2023a] Single-cell-python-bundle 2024.02

boegel:
```
83 out of 303 required modules missing:

* gawk/5.3.0-GCC-12.3.0 (gawk-5.3.0-GCC-12.3.0.eb)
* mm-common/1.0.6-GCCcore-12.3.0 (mm-common-1.0.6-GCCcore-12.3.0.eb)
* sparsehash/2.0.4-GCCcore-12.3.0 (sparsehash-2.0.4-GCCcore-12.3.0.eb)
* libsigc++/3.6.0-GCCcore-12.3.0 (libsigc++-3.6.0-GCCcore-12.3.0.eb)
* Boost.Python/1.82.0-GCC-12.3.0 (Boost.Python-1.82.0-GCC-12.3.0.eb)
* tqdm/4.66.1-GCCcore-12.3.0 (tqdm-4.66.1-GCCcore-12.3.0.eb)
* PyCairo/1.25.0-GCCcore-12.3.0 (PyCairo-1.25.0-GCCcore-12.3.0.eb)
* cairomm/1.16.2-GCC-12.3.0 (cairomm-1.16.2-GCC-12.3.0.eb)
* PyGObject/3.46.0-GCCcore-12.3.0 (PyGObject-3.46.0-GCCcore-12.3.0.eb)
* LLVM/14.0.6-GCCcore-12.3.0-llvmlite (LLVM-14.0.6-GCCcore-12.3.0-llvmlite.eb)
* numba/0.58.1-foss-2023a (numba-0.58.1-foss-2023a.eb)
* Pyomo/6.7.3-foss-2023a (Pyomo-6.7.3-foss-2023a.eb)
* statsmodels/0.14.1-gfbf-2023a (statsmodels-0.14.1-gfbf-2023a.eb)
* scikit-misc/0.3.1-foss-2023a (scikit-misc-0.3.1-foss-2023a.eb)
* anndata/0.10.5.post1-foss-2023a (anndata-0.10.5.post1-foss-2023a.eb)
* pydot/2.0.0-GCCcore-12.3.0 (pydot-2.0.0-GCCcore-12.3.0.eb)
* openpyxl/3.1.2-GCCcore-12.3.0 (openpyxl-3.1.2-GCCcore-12.3.0.eb)
* umap-learn/0.5.5-foss-2023a (umap-learn-0.5.5-foss-2023a.eb)
[...]

PR #1133, opened 2025-06-27
{2023.06}[foss/2023a] Clang v16.0.6

laraPPr:


laraPPr:
Update of https://github.com/EESSI/software-layer/pull/737/files after split of software-layer repos.

laraPPr:
Ah I saw that this one was merged in EasyBuild 5 https://github.com/easybuilders/easybuild-easyblocks/pull/3373. SO thought that it was picking everything up. Which other changes should be picked up?

Crivella:
This and all the successive PRs are for the LLVM easyblock.
Before `18.1.6` clang still has to use the `Clang` one which will be deprecated afterwards in of switching Clang everywhere to LLVM.

I've made one EC for LLVM with the idea of it ending up on EESSI but we still have some easyblock PRs open on that front which would make sense for them to be merged before we try it https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/LLVM/LLVM-20.1.5-GCCcore-13.2.0.eb

ocaisa:
@laraPPr I don't think we want to pursue this version of Clang, better to use the really recent LLVM work of @Crivella and friends

laraPPr:
allright. It's the depency this one got stuck on, https://github.com/EESSI/software-layer/pull/637. But I'm ok with updating that Single-cell-python bundle and than relying on a newer clang.

PR #1148, opened 2025-08-08
Draft: Software licenses tracking 

hvelab:
Re-opened here as per https://github.com/EESSI/software-layer-scripts/pull/55

Added new logic and the scripts used to try to pre-fetch the license before requiring human interaction 

Needs further testing and refinement


hvelab:
just note that i put the scripts for fetching here but as soon as the testing is done I will move them to software-layer-scripts repo

PR #1149, opened 2025-08-08
Determine missing license information for packages to be installed

hvelab:
New PR as per https://github.com/EESSI/software-layer-scripts/pull/55 

Needs further testing and refinement, as soon as scripts are tested I will move them to software-layer-scripts

### DIFF
--- a/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a-CUDA.yml
+++ b/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a-CUDA.yml
@@ -1,5 +1,2 @@
 easyconfigs:
   - OSU-Micro-Benchmarks-7.2-gompi-2023a-CUDA-12.1.1.eb
-  - OTF2-3.0.3-GCCcore-12.3.0.eb
-  - XlsxWriter-3.1.3-GCCcore-12.3.0.eb
-  - spdlog-1.11.0-GCCcore-12.3.0.eb


.github/workflows/check_licenses.yml
```diff
@@ -0,0 +1,125 @@
...
+permissions:
+  contents: read # we dont need to write
+
+jobs:
+  license_update:
+    runs-on: ubuntu-latest
+
+    steps:
+      -  uses: actions/checkout@v4
+      -  uses: eessi/github-action-eessi@v3
```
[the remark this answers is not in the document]
ocaisa: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/LAMMPS/LAMMPS-29Aug2024_update2-foss-2024a-kokkos.eb#L60-L65

.github/workflows/check_licenses.yml
```diff
@@ -0,0 +1,125 @@
...
+
+          for easystack_file in $file_list; do
+              eb_version=$(echo ${easystack_file} | sed 's/.*eb-\([0-9.]*\).*.yml/\1/g')
+              echo "check missing installations for ${easystack_file} with EasyBuild ${eb_version}..."
+              module purge
+              module load EasyBuild/${eb_version}
+              module load EESSI-extend/${EESSI_VERSION}-easybuild
+              which eb
+              ${EB:-eb}  --version
+              ${EB:-eb}  --missing --easystack ${easystack_file} 2>&1 | tee ${eb_missing_out}
```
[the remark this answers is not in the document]
hvelab: In the last commits I applied "--robot" and the output is the same

.github/workflows/check_licenses.yml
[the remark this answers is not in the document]
hvelab: i've checked in all easyconfigs and LAMMPS is the only one that adds dependencies depending on the arch, the rest just change some compilation options which is not relevant here, anyway i fixed the matrix (with only x86 and a64fx) and will add what you mentioned to add the EESSI version

licenses/parse_licenses.py
```diff
@@ -0,0 +1,239 @@
...
+        if isinstance(license_info, list):
+            license_info = ", ".join(license_info) if license_info else "not found"  # Convert list to string
+
+        license_info_normalized = license_info.lower() if isinstance(license_info, str) else license_info
+
+        if license_info_normalized in SPDX_LICENSES:
+            spdx_details = SPDX_LICENSES[license_info_normalized]
+            is_redistributable = spdx_details["isOsiApproved"] or spdx_details["isFsfLibre"]
+
+        results[module_name] = {
```
hvelab: change to yaml

licenses/parse_licenses.py
```diff
@@ -0,0 +1,239 @@
...
+            "Permission to redistribute": is_redistributable,
+            "Retrieved from": url
+        }
+    return results
+
+
+def save_license_results(results, output_file="licenses_test.json"):
+    """Saves license information to a JSON file."""
+    with open(output_file, "w") as f:
+        json.dump(results, f, indent=4)
```
hvelab: dump in yaml

easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-4.9.0-2023a-CUDA.yml
```diff
@@ -1,2 +1,3 @@
 easyconfigs:
   - OSU-Micro-Benchmarks-7.2-gompi-2023a-CUDA-12.1.1.eb
+
```
trz42: Is this extra empty line needed?

licenses/spdx.py
```diff
@@ -0,0 +1,100 @@
...
+
+# Configure the logging module
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
+
+
+def get_spdx_license_list():
+    """
+    Download JSON file with current list of SPDX licenses, parse it, and return it as a Python dictionary.
+    """
+    global spdx_license_list
```
trz42: Maybe, instead of using a global variable, add an argument to the function.

licenses/spdx.py
```diff
@@ -0,0 +1,100 @@
...
+
+def license_info(spdx_id):
+    """Find license with specified SPDX identifier."""
+ 
+    spdx_license_list = get_spdx_license_list()
+
+    licenses = spdx_license_list['licenses']
+    for lic in licenses:
+        if lic['licenseId'] == spdx_id:
+            return lic
```
trz42: Maybe there are not many licenses, but it seems a dict data structure could be a better fit. Food for future work.

trz42 (changes requested):
Overall this PR adds a workflow that determines for which packages license information is missing and supplies a patch that can be used to add the missing information (manual step). At the moment it succeeds when license information is missing. We may want to change this in the future. To my understanding it does not verify if all packages (to be added) allow redistribution, but this was likely not the intention of the PR. Nevertheless it's an important step towards checking licenses and thus we should not delay start using this workflow.

Went through this and tested it in https://github.com/trz42/software-layer/pull/94 and https://github.com/trz42/software-layer/pull/95

I suggest to only include the changes that are used in https://github.com/trz42/software-layer/pull/95 and will make a PR to your, so you can easily apply the changes.

Before we can merge this PR, the easystack file needs to be removed.

PR #1152, opened 2025-08-13
{2023.06}[2023b,a64fx] apps originally built with EB 4.9.3

agimenog:
Based on #992


boegel:
`M4/1.4.19` being missing in EESSI 2025.06 has nothing to do with this PR, so ignoring that and merging this PR

PR #1146, opened 2025-08-05
{2025.06} GCC 13.3.0 + 14.2.0

boegel:
With https://github.com/EESSI/software-layer-scripts/pull/45 merged, we should be ready to start building for EESSI 2025.06...

PR #1158, opened 2025-08-21
{2025.06} foss/2024a

boegel:
Will need merge conflict fix after merging of:
- https://github.com/EESSI/software-layer/pull/1146

### DIFF
--- a/easystacks/software.eessi.io/2025.06/eessi-2025.06-eb-5.1.1-2024a.yml
+++ b/easystacks/software.eessi.io/2025.06/eessi-2025.06-eb-5.1.1-2024a.yml
@@ -1,0 +1,3 @@
+easyconfigs:
+  - GCC-13.3.0.eb
+  - foss-2024a.eb


bedroge:
> So it's basically not happy with the `automake` we have in the EESSI 2025.06 compat layer... _sigh_

See https://gitlab.com/eessi/support/-/issues/56#note_2551021911.

bedroge:
> So either we run `autogen.pl` via `preconfigopts` to fix the problem, or we stop filtering `Automake` & co and rely on them as build dependencies provided by EasyBuild rather than coming from the compat layer.
> 
> @bedroge Which option do you prefer?

No strong preference, but I think I'd remove it from the filtered dependencies and rely on the versions provided by EB.

boegel:
> > So either we run `autogen.pl` via `preconfigopts` to fix the problem, or we stop filtering `Automake` & co and rely on them as build dependencies provided by EasyBuild rather than coming from the compat layer.
> > @bedroge Which option do you prefer?
> 
> No strong preference, but I think I'd remove it from the filtered dependencies and rely on the versions provided by EB.

PR to change this:
* https://github.com/EESSI/software-layer-scripts/pull/74

PR #1159, opened 2025-08-21
{2025.06} foss/2025a

boegel:
Will need merge conflict fix after merging of:
- https://github.com/EESSI/software-layer/pull/1146


boegel:
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-mc-aws arch:x86_64/intel/sapphirerapids
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-jsc architecture:aarch64/nvidia/grace

boegel:
Same problem in both builds: some tests that are being run (during the `build` step) are failing:
```
LD_LIBRARY_PATH=/tmp/eessibot/easybuild/build/Python/3.13.1/GCCcore-14.2.0/Python-3.13.1 ./python -m test --pgo --timeout=
...
0:00:13 load avg: 1.33 [16/44] test_embed
test test_embed failed
0:00:14 load avg: 1.33 [17/44] test_float -- test_embed failed (71 failures)
```

ocaisa:
We need to add a hook to enable --enable-ipv6 for OpenMPI (or make a PR upstream and use that)

boegel:
However, using this in the `Python` easyconfig does *not* actually fix the problem, so maybe the problem isn't that `$LIBS` set after all...

```python
prebuildopts = "unset LIBS && "
```

Indeed, also when trying via the interactive debug shell, just the act of removing `Programs/_testembed` and then re-running `make` fixes the problem (even without unsetting `$LIBS`)...

boegel:
One more data point: the problem doesn't occur when I'm using `eb --disable-rpath` on top of `EESSI-extend` (without using the build container), but does happen when RPATH linking is left enabled.

That suggests it may be a bug of some kind in the EasyBuild RPATH wrappers?

ocaisa:
@boegel I wonder if there is a chance this might be fixed with rpath wrapper updates that will come with EB 5.1.2?

boegel:
> @boegel I wonder if there is a chance this might be fixed with rpath wrapper updates that will come with EB 5.1.2?

I did a quick test using current `develop` branch for EasyBuild framework, and that didn't help.

I'll try to dig a bit more in the coming days, I'll also open a dedicated issue on this since there's probably some kind of lesson to learn from this one...

PR #1161, opened 2025-08-22
{2025.06} gfbf/2025a + FFTW 3.3.10

boegel:
One step closer to `foss/2025a`.

Not including `OpenMPI` yet, because that build failed in https://github.com/EESSI/software-layer/pull/1159, need to figure out what the problem is...

Will need merge conflict fix after merging of https://github.com/EESSI/software-layer/pull/1146


boegel:
Failing build in #1159 was `Python`, which is also required here (build dep for FlexiBLAS), so will need to take a closer look at that first...

PR #1162, opened 2025-08-22
also check on EESSI 2025.06 w.r.t. missing installations and software layer being in sync across CPU targets

boegel:


.github/workflows/test_compare_stacks.yml
```diff
@@ -21,6 +21,7 @@ jobs:
       matrix:
         EESSI_VERSION:
         - 2023.06
+        - 2025.06
```
ocaisa, line 24: Are we ignoring A64FX for this scenario? (It is likely to be a bit of a problem case, so I'm fine with that, but it is also nice to be able to see how it is not consistent)

ocaisa (approved):
LGTM

PR #1172, opened 2025-09-01
{2025.06}[foss/2024a] VASP 6.5.0

Atisom:


ocaisa:
@Atisom The restricted licence on VASP means we cannot include it in EESSI.

Atisom:
OK, thank you!

PR #1215, opened 2025-10-03
{2023.06}[2022b, a64fx] Rebuild Rust 1.65.0 and Python 3.10.8

bedroge:
Requires https://github.com/EESSI/software-layer-scripts/pull/97.

This replaces the dummy module that only prints an LmodError. Also rebuilds Python, as it has a build dependency on that particular version of Rust.


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-deucalion for:arch=aarch64/a64fx

bedroge:
Cancelled that job, just noticed that Python needs to be built with a different EB version, so I'll split that easystack.

ocaisa:
Staging bundle merged

PR #1221, opened 2025-10-08
enable "compare stacks" CI for A64FX

bedroge:
We're at ~77% now (836/1089). The CI will obviously still fail, but can be used to find out what's still missing/ I've left the original code as comments, figured that they could serve as an example for excluding targets.


boegel:
With the deployment of GROMACS 2024.1 being triggered (cfr. https://github.com/EESSI/software-layer/pull/1308), A64FX is now fully in sync with other CPU targets for EESSI 2023.06, so this check should go green really soon now...

ocaisa (approved):
LGTM, @bedroge for the win

PR #1236, opened 2025-10-15
{2025.06}[2024a] TurboVNC 3.1.2

ocaisa:
This is not so useful until we have a window manager that we can start up for vnc


ocaisa:
bot: build repo:eessi.io-2025.06-software instance:eessi-bot-jsc for:arch=aarch64/nvidia/grace

PR #1238, opened 2025-10-16
{2023.06} Rebuild Python 3.11.3 + 3.11.5 for ctypes fix

casparvl:


boegel:
@casparvl Failure?

casparvl:
Interesting, the failure is for GCCcore-12.2.0 on zen4 specifically, which isn't supported there. I do see that all the steps get skipped (as expected), as it should just happily regenerate the module file that prints the `LmodError` that this combination isn't supported. It doesn't. 

One way forward is to split off the `GCCcore-12.2.0` rebuild, and simply not do that on zen4 (and keep the other three rebuilds in this PR and do it for all archs including zen4). There's no point for that particular rebuild on zen4 anyway, nothing will change.

boegel:
@casparvl Is there a particular reason you prefer using the SURF bot to build for `zen2`?
We better keep that consistent, and use the usual AWS bot (although end result shouldn't matter)?

casparvl:
Well, that's a possibility, but wouldn't get you the last point of actually noticing problems that would be caused by building on different infra. It should not matter, and if it does, we should fix that. I guess opinions differ on what is nice and not nice https://github.com/EESSI/software-layer/pull/1251#issuecomment-3429576957 :D

I'll raise an issue in the support channel so we can discuss it in the next sync meeting.

easystacks/software.eessi.io/2023.06/rebuilds/20251016-eb-5.1.2-Python-ctypes-improved-patch.yml
```diff
@@ -0,0 +1,20 @@
...
+  - Python-3.11.3-GCCcore-12.3.0:
+      options:
+        # See https://github.com/easybuilders/easybuild-easyblocks/pull/3860
+        include-easyblocks-from-commit: 3fa0df7414d7d8415fe34f7758649f966322472e
+        from-commit: 08aa521b52c9f83d8df345d083e46d6f304d5641
+  - Python-3.11.5-GCCcore-13.2.0:
+      options:
+        # See https://github.com/easybuilders/easybuild-easyblocks/pull/3860
+        include-easyblocks-from-commit: 3fa0df7414d7d8415fe34f7758649f966322472e
+        from-commit: 08aa521b52c9f83d8df345d083e46d6f304d5641
```
bedroge:
```suggestion
  - Python-3.11.3-GCCcore-12.3.0:
      options:
        # See https://github.com/easybuilders/easybuild-easyblocks/pull/3860
        # and https://github.com/easybuilders/easybuild-easyblocks/pull/3965
        include-easyblocks-from-commit: 3fa0df7414d7d8415fe34f7758649f966322472e
        # See https://github.com/easybuilders/easybuild-easyconfigs/pull/23499
        from-commit: 08aa521b52c9f83d8df345d083e46d6f304d5641
  - Python-3.11.5-GCCcore-13.2.0:
      options:
        # See https://github.com/easybuilders/easybuild-easyblocks/pull/3860
        # and https://github.com/easybuilders/easybuild-easyblocks/pull/3965
        include-easyblocks-from-commit: 3fa0df7414d7d8415fe34f7758649f966322472e
        # See https://github.com/easybuilders/easybuild-easyconfigs/pull/23499
        from-commit: 08aa521b52c9f83d8df345d083e46d6f304d5641
```

casparvl:
bot:status last_build

bedroge:
All tarballs have been ingested.

PR #1266, opened 2025-10-25
{2023.06}[2023b] LINC v5.0

casparvl:


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-mc-aws for:arch=x86_64/amd/zen2

casparvl:
Asked one of my AI friends... It pointed out that the current limit for `ulimit -l` is 8192 KB. It also pointed out that on the host, (where an EESSI-extend installation worked), it's probably set to unlimited - which I checked and is indeed true.

It probably doesn't fail on AWS because UCX never goes down this code path: UCX is trying to setup an RDMA connection here, which I guess it doesn't have support for on the AWS interconnect. There, it probably just goes for TCP, which means it never hits this error.

The solution is simple:
- Confirm that the `ulimit -l` in the container is indeed 8192 KB
- Raise the ulimit in the container (although it is a bit unclear why this is lower - the same AI friend gives ambiguous answers :)) - but when and where? Can we do this in a way that the _default_ is different? Do we have to do it as part of the install scripts?

casparvl:
bot:status last_build

bedroge (approved):

PR #1273, opened 2025-10-29
{2023.06}[2023a,2023b] Rebuild UCX v1.14.1, v1.15.0 to fix performance issues

casparvl:

PR #1274, opened 2025-10-29
{2023.06}[2022b] Rebuild for UCX 1.13.1

casparvl:
This was split off from https://github.com/EESSI/software-layer/pull/1273 because the zen4 rebuilds fail. However, we don't _need_ to rebuild for zen4, we can just skip those here: the 2022b toolchain isn't supported there anyway.


casparvl:
bot:status last_build

casparvl:
All good, the zen4 one is failure, but that shouldn't be deployed anyway.

bedroge (approved):

PR #1302, opened 2025-11-13
{2023.06}[2023b, a64fx] GROMACS 2024.3 + 2024.4

bedroge:


bedroge:
bot: build repo:eessi.io-2023.06-software instance:eessi-bot-deucalion for:arch=aarch64/a64fx

bedroge:
> @bedroge Can we remove the exclusion from CI now?

Almost, there's one missing GROMACS version left (2024.1). That one had many double precision test failures. I'll check if there's an easy fix for that (2024.3 does work fine, so something got fixed), but otherwise I'd say that we just install a dummy modulefile that prints "this version is not supported on A64FX".

PR #1312, opened 2025-11-20
{2023.06}[2023a+2023b] Rebuild UCX 1.14.1 + 1.15.0 without `--with-sysroot`

bedroge:
See https://github.com/EESSI/software-layer-scripts/pull/133. Rebuilds the 2023a+2023b versions for all CPU targets, using the commit that was used for the current build (with the bistro fix).


ocaisa:
Before we merge this, did you check that rdma is still being picked up?

PR #1337, opened 2025-12-12
{2025.06}[2024a] LAMMPS v29Aug2024_update2

laraPPr:


easystacks/software.eessi.io/2025.06/eessi-2025.06-eb-5.1.2-2024a.yml
```diff
@@ -100,3 +100,9 @@ easyconfigs:
   - R-bundle-Bioconductor-3.20-foss-2024a-R-4.4.2.eb
   - tmux-3.4-GCCcore-13.3.0.eb
   - JupyterLab-4.2.5-GCCcore-13.3.0.eb
+  - LAMMPS-29Aug2024_update2-foss-2024a-kokkos.eb:
+      options:
+        # See https://github.com/easybuilders/easybuild-easyconfigs/pull/23719
+        from-commit: 379c0f01a29109a3c3db1e1807838f50074af143
+        # See https://github.com/easybuilders/easybuild-easyblocks/pull/3894
+        include-easyblocks-from-commit: 60af21ba2173dd8d4ce7390aa60bd16718a28546
```
lorisercole:
```suggestion
        include-easyblocks-from-commit: 433d723ceb5492ab22bf45566e40296d01db87c2
```
This is the correct merge commit of https://github.com/easybuilders/easybuild-easyblocks/pull/3894

lorisercole:
The MPI ones seem to be due to a MPI configuration problem. The tests apparently require 4 MPI slots, but not enough are available

ocaisa:
Ah, I knew we would hit this at some point, we need to set the environment variables to allow oversubscription for OpenMPI 5

laraPPr:
I think I had at least 4 cores in my allocation but I will have to double check. Because I think I had exact 4 cores and maybe it needs at least 5. Also should we do a check that these are not run when there are to little cores. Or add the `--oversubscribe` like @ocaisa sugested.
I'll try to get a clearer view on why the `KimCommands` test is failing next week.

PR #1353, opened 2026-01-13
Clone specific software-layer-commit and implement CI to check merged status

casparvl:
This PR is an initial step in creating a workflow where we can use PRs to `software-layer-scripts` directly, and then once they are merged, just update the SHA checksum to a (Github-signed) merge commit, rebuild, and be done :)

Edit: better description...

This PR contains three changes:
1. `bot/build.sh` now clones a particular commit, which is specified in `bot/commit_sha`
2. CI job `Verify software-layer-scripts / check_bot_build_checksum` verifies the checksum of `bot/build.sh` against a reference hardcoded in the workflow file. This way, a malicious contributor would have to modify both `bot/build.sh` and the workflow file, which would (hopefully) stand out to a reviewer.
3. CI job `Verify software-layer-scripts / check_software_layer_scripts_commit` which check that `bot/commit_sha` is a commit that is part of the upstream https://github.com/EESSI/software-layer-scripts , is on the `main` branch (i.e. has been merged), is a merge commit, and is signed with the public Github GPG for the web interface.

We might need to update the `commit_sha` already (not sure if there have been more merges to software-layer-scripts since I started this) before we merge this to make sure this PR doesn't actually revert us to using an _older_ version.


casparvl:
Ok, as expected, both `Verify software-layer-scripts / check-bot_build_checksum` and `Verify software-layer-scripts / check_software_layer_scripts_commit` pass on https://github.com/EESSI/software-layer/pull/1353/commits/bce9bbc441c22aaf3eefaf4996df8285d649cb0b

Let's test again by changing the sha checksum...

casparvl:
Ok, on https://github.com/EESSI/software-layer/pull/1353/commits/bee1d296ce80f8ba3a8ae3e94cc304cac09f2879 we again have the expected failure:

> Commit c0a3ff09a3a38737af5a922fdf581aa7b2dd6c88 is NOT merged into origin/main.
Error: Process completed with exit code 1.

Let's change the sha checksum back, and now change something in `bot/build.sh` to test the other CI job.

bot/commit_sha
bedroge: I would make it clear that this sha belongs to a software-layer-scripts commit, so maybe rename it to `software-layer-scripts.commit_sha` or something?

casparvl:
Discussed in support meeting. One issue came up - though it doesn't affect _this_ specific PR, nor the CI checks in it.

The issue is related to how we might _use_ the commit_sha introduced in this PR to make the bot refuse to deploy. The rationale is: if a bot has tarballs with a `commit_sha` that _do not_ match the `bot/commit_sha`, those are invalid and should not be deployed. However, what may happen is this:
- software-layer-scripts PR 1 introduces a change (e.g. to `eb_hooks`), commit `X`
- software-layer PR 1 uses it to build some software, and proves that the new `eb_hooks` solves some issue
- software-layer-scripts PR 1 is merged, creating a merge-commit `Y`
- software-layer PR 1 updates `bot/commit_sha` to `Y` and rebuilds all tarballs
- software-layer-scripts PR 2 introduces a different change, which gets accepted and merged with merge commit `Z`
- software-layer PR 2 uses merge commit `Z` to build all tarballs
- software-layer PR 2 gets deployed and merged
- software-layer PR 1 gets synced with `main`. This means `bot/commit_sha` should be updated to `Z` (as `Z` happened after `Y`, and we don't want a merge of software-layer PR 1 to 'downgrade' the `bot/commit_sha` to an older version than the one on `main`). This now invalidates all builds done with `bot/commit_sha`=`Y` - even though those are perfectly valid builds!

I'm not yet sure how to avoid this. It is not possible to retain `Y` in SW PR1, as a merge of PR1 would then cause `main` to contain `bot/commit_sha=Y`, which we don't want if `Z` is newer than `Y`. So we have to come up with a way that if we then accept `Z` in `bot/commit_sha` in the PR 1, that builds with `Y` are still considered 'valid'.

.github/workflows/test_software_layer_scripts.yml
```diff
@@ -0,0 +1,127 @@
...
+    steps:
+        - name: Check out software-layer repository (shallow)
+          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+          with:
+              fetch-depth: 1  # We only need the current revision to read bot/commit_sha
+        - name: Checkout software-layer-scripts (full history)
+          uses: actions/checkout@v4
+          with:
+            repository: EESSI/software-layer-scripts
+            path: upstream-scripts
```
boegel: Why make it difficult, just call this `EESSI-software-layer-scripts` ?

boegel (dismissed):

laraPPr:
@casparvl Where is this stuck?

casparvl:
At a technical level, I don't know if it is even stuck - I think this worked. But back then I felt we lacked the commitment to actually adopt this approach - though it keeps popping up that our existing workflow is annoying. I would like to get back to this, so please leave it open.

I think at a technical level, the one thing that would be needed is an accompanying bot PR that makes the bot print _which_ checksum was used. I think we discussed at some point how to do this in a generic way, and the plan - if I remember correctly - was to have a configurable item in app.cfg that points to a file in the repo that - if it exists - is read, and the information is added to the bot's feedback in the PR (and then also to the summary table).

casparvl:
Made a PR for the bot to report the commit SHA of the `software-layer` clone https://github.com/EESSI/eessi-bot-software-layer/pull/386 . Together with this PR (#1353) , which defines _which commit_ is cloned from `software-layer-scripts` by means of the `bot/software_layer_scripts_commit` file, we get full traceability: the `software-layer` commit sha refers to a certain version of `bot/software_layer_scripts_commit` which defines which `software-layer-scripts` commit was checked out to do the build process.

casparvl:
TODO: 

- Implement a CI that is only run _after_ the deploy, which checks that the `bot/software_layer_scripts_commit` points to the current tip of the `software-layer-scripts` main branch. This is so that whoever creates a new feature branch does so based on the latest commit on `software-layer-scripts`' main. 
 
This means that if a PR changed `bot/software_layer_scripts_commit` explicitely, it _will_ have to be updated to the tip of the main `software-layer-scripts` right before the PR is merged, to ensure that the `bot/software_layer_scripts_commit` in the `main` branch of `software-layer` always points to the latest `software-layer-scripts` commit on `main`.

One caveat: if someone forks, and doesn't sync with main, builds _will_ happen with a somewhat older version of `software-layer-scripts`. I think however that this is desirable: currently, we sometimes have builds in `software-layer` that are done 5 minutes apart, which use different versions of `software-layer-scripts`, just because someone has happened to deploy something from there in the meantime. That is currently completely invisible, and therefore completely undesirable. It is exactly the type of behavior we want to avoid...

Note that even after the above TODO item is tackled, it is still the reviewers that are responsible for checking that they only set a deploy label _after_ the CI "Verify software-layer-scripts / check_software_layer_scripts_commit" passes, as this ensures that the builds are done from a merged `software-layer-scripts` commit. I won't be implementing a hard check on the bot side right now. We have discussed that having functionality in the bot for a pre-deploy check, that would check the status of certain CIs (for which the names are then configurable) would be nice. But that should not block adopting this new workflow: including the above TODO, it is already an improvement over the current workflow.

casparvl:
I thought about it some more, and maybe we can do better.

- We make a CI in `software-layer` that updates the SHA checksum to the tip of `software-layer-scripts/main` upon merges into main.

- Additionally, we can make a CI in `software-layer` for which the trigger is a CI in `software-layer-scripts` that runs on that main branch, and sends the new SHA as a payload. According to my AI:


=======

Create trigger on repo A (software-layer):

- Go to Settings > CI / CD > Pipeline triggers in repo A.
- Click Add trigger → give it a name (e.g., update‑sha‑trigger).
- Note the Trigger URL and the generated Trigger token.

You’ll use these in repo B to start a pipeline in repo A.

CI on repo B (software-layer-scripts):
```
stages:
  - notify

# Run on every push to the main branch
update_repo_a_sha:
  stage: notify
  only:
    - main
  script:
    # Optional: fetch the new SHA and pass it as a variable
    - NEW_SHA=$(git rev-parse HEAD)
    - echo "New SHA is $NEW_SHA"
    # Trigger repo A pipeline, passing the SHA as a CI variable
    - >
      curl --request POST
      --form token=TRIGGER_TOKEN   # ← replace with the token from repo A
      --form ref=main
      --form "variables[NEW_SHA]=$NEW_SHA"
      https://gitlab.com/api/v4/projects/PROJECT_ID/trigger/pipeline
```

CI on repo A (software-layer)

```
stages:
  - update_sha

update_sha_file:
  stage: update_sha
  only:
    - triggers        # run only when started by a trigger
  variables:
    GIT_STRATEGY: fetch   # ensure we have the full history
  script:
    # 1. Get the SHA passed from repo B (fallback: fetch via API)
    - |
      if [ -z "$NEW_SHA" ]; then
        # No variable – fetch the latest SHA from repo B via API
        B_PROJECT_ID=YOUR_B_PROJECT_ID
        NEW_SHA=$(curl --silent "https://gitlab.com/api/v4/projects/${B_PROJECT_ID}/repository/branches/main" \
                         --header "PRIVATE-TOKEN:${CI_JOB_TOKEN}" |
                         jq -r .commit.id)
      fi
      echo "Using SHA: $NEW_SHA"

    # 2. Replace the old SHA in the target file (e.g., src/commit_ref.txt)
    - FILE=path/to/your/file.txt
    - |
      # Assume the file contains only the SHA; otherwise adjust the sed pattern.
      echo "$NEW_SHA" > $FILE

    # 3. Commit and push the change back to repo A
    - git config user.email "ci-bot@example.com"
    - git config user.name "CI Bot"
    - git add $FILE
    - git commit -m "Update repo B SHA to $NEW_SHA [ci skip]"
    # Use a personal access token (PAT) with write_repo permission
    - git push "https://${CI_JOB_TOKEN}@gitlab.com/namespace/repo-a.git" HEAD:main
```

=======

The downside is that this requires tokens that have direct write access to your main branch. It should be possible to do this in a controlled way in the GH CI space (masked & protected tokens), but if that makes us uncomfortable, the alternative would be:

- [ ] A CI that is triggered when the `bot:deploy` label is set (typically all the building is done then, so we can freely update the SHA commit) , and which then either fails if the commit SHA is not pointing to the head of `software-layer-scripts` _or_ even has write permission and proactively pushes the latest commit SHA to the feature branch. The advantage is that this the only happens on the feature branch, and is still 'in view' of a reviewer before actually merging it.
- [ ]  A second CI that is triggered on commits to `software-layer-scripts/main` that creates a PR to `software-layer/main` to update the SHA commit.

That solution only requires write tokens on feature branches.

casparvl:
Some more interaction with my AI tells me that I need to use `GITHUB_TOKEN` instead of `CI_JOB_TOKEN` if we want it to have write access. Also, the branch protection rules should be modified in order for the `GITHUB_TOKEN` to be able to do this (.e.g  '“Allow GitHub Actions to bypass branch protection rules”`). We need to discuss if we want this.

casparvl:
Ok, to avoid stalling this over discussion, let's phase the approach:

First I'll implement:
- A CI that runs only after the `bot:deploy` label has been added, and checks if the `bot/software_layer_commit_sha` points to the tip of `software-layer-scripts/main`
- A Github Action that runs periodically (every few minutes or so) and checks if the `bot/software_layer_commit_sha` on `software-layer/main` still points to the tip of `software-layer-scripts/main`. And if not, creates a PR.

This should be totally safe, and only requires the 2nd GH action to have write access to a dedicated feature branch which I'll call `ci-update-sha`.

casparvl:
Perfect, https://github.com/EESSI/software-layer/actions/runs/31018598477/job/92349130424?pr=1353 goes green immediately (and takes only 6s to run). Let me change the checksum to an older merge commit.

casparvl:
Excellent, CI https://github.com/EESSI/software-layer/actions/runs/31018815032/job/92349881283?pr=1353 goes red immediately, reporting:

```
ERROR: bot/software_layer_scripts_commit does not match the latest commit on main of EESSI/software-layer-scripts.
Stored commit:  83169938a52e22849226308cf3010e496f267430
Latest commit:  62b8e332fb041a9ba598dbed3ed781a3416407c0
Please update bot/software_layer_scripts_commit to 62b8e332fb041a9ba598dbed3ed781a3416407c0 and push the change.
Error: Process completed with exit code 1.
```

Now, let's move to an unmerged commit, check that our other CI also goes red in that case.

casparvl:
Perfect, the other CI also fails https://github.com/EESSI/software-layer/actions/runs/31019123923/job/92350942842?pr=1353 with:

```
Commit 89e670f2336a610827ef899ecad520af78b0a848 is NOT merged into origin/main.
```

Exactly as itended. This CI always runs, and provides reviewers with a red flag that this thing is NOT ready to deploy, as it's not pointing to a merge commit.

.github/workflows/check_software_layer_scripts_commit.yml
```diff
@@ -0,0 +1,61 @@
...
+#
+# This workflow checks if the commit SHA in bot/software_layer_scripts_commit
+# matches the latest commit on the main branch of EESSI/software-layer-scripts.
+# It only runs when the 'bot:deploy' label is present on the PR. The reason is that 
+# before deployment, we typically want to keep using a fixed version of EESSI/software-layer-scripts
+# so that we don't (unintentionally) change build behavior half-way through a PR.
+# However, after deploy, we want the `bot/software_layer_script_commit to be updated
+# so that on software-layer/main this always refers to as recent a commit SHA 
+# from EESSI/software-layer-scripts as possible
+name: Check software-layer-scripts commit is up to date
```
casparvl, line 18: Discussed with Bob. We should adapt this so that it only runs on feature branches that have actually changed `bot/software_layer_scripts_commit`. Otherwise, contributors that have no idea about this file will be asked to update it.

Instead, we should make a github action that automatically makes PRs to update the `bot/software_layer_scripts_commit`.

.github/workflows/test_software_layer_scripts.yml
```diff
@@ -0,0 +1,127 @@
+# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
+#
+# This workflow verifies that the correct version of software-layer-scripts is used.
+# 
+# First, check_bot_build_checksums checks if the bot/build.sh code that clones software-layer-scripts is untouched,
+# as this normally shouldn't change (a change could mean a contributor is trying to inject something
```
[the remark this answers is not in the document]
casparvl: I think this is the first change since we've split off `software-layer-scripts` from `software-layer`, well over a year ago. Plus, the wording is 'could', it doesn't say it _is_ malicious.

The point is: there is very little reason to modify `bot/build.sh` going forward. Builds from feature branches are already possible, as long as they are open PRs to `EESSI/software-layer-scripts`. If `bot/build.sh` is modified, we want it to stand out _really_ well, since it _is_ a potential risk to do builds on a PR that uses some arbitrary `bot/build.sh`. 

I know never to say never, so it can still happen - and that's ok. It'll just require a checksum update in the CI as well :)

.github/workflows/check_software_layer_scripts_commit.yml
casparvl: Done

.github/workflows/check_software_layer_scripts_commit.yml
```diff
@@ -0,0 +1,70 @@
...
+# matches the latest commit on the main branch of EESSI/software-layer-scripts.
+# It only runs when the 'bot:deploy' label is present on the PR AND if the bot/software_layer_scripts_commit
+# has been changed. This ensures that:
+# - Contributors that don't know about bot/software_layer_scripts_commit (i.e.: haven't updated it) don't get
+# a CI that complains about the version not being the latest. That's fine, we will update that file in other PRs
+# and if users don't change the file in their PR, there's no risk of overwriting with an older version
+# - Contributors that have changed bot/software_layer_scripts_commit can build with a fixed version (commit)
+# of EESSI/software-layer-scripts, ensuring that builds for all targets are done in a coherent manner.
+# After all builds are deployed, the CI is run, checking that this is the latest commit. If not, contributors
+# will have to update that before merging. This ensures that mergin their PR does not cause software-layer/main's
```
bedroge:
```suggestion
# will have to update that before merging. This ensures that merging their PR does not cause software-layer/main's
```

bot/build.sh
```diff
@@ -3,7 +3,19 @@
 # give up as soon as any error occurs
 set -e
 
-git clone https://github.com/EESSI/software-layer-scripts
+TOPDIR=$(dirname $(realpath $0))
+
+# Clone a the commit from software-layer-script that corresponds to `bot/software_layer_scripts_commit`
```
bedroge:
```suggestion
# Clone the commit from software-layer-scripts that corresponds to `bot/software_layer_scripts_commit`
```

.github/workflows/test_software_layer_scripts.yml
```diff
@@ -0,0 +1,127 @@
...
+# malicious). Having this CI means that a change in bot/build.sh should at least be accompanied by
```
bedroge: I don't mind using the current sentence (as it says `could`), but if @boegel does not like it, maybe we could use something like:

```suggestion
# as this normally shouldn't change, so let's be really cautious here. Having this CI means that a change in bot/build.sh should at least be accompanied by
```

bedroge (changes requested):
Few tiny typos/suggestions

.github/workflows/check_software_layer_scripts_commit.yml
```diff
@@ -0,0 +1,70 @@
...
+    paths:
+      - 'bot/software_layer_scripts_commit'
+permissions:
+  contents: read # to fetch code (actions/checkout)
+jobs:
+  check_latest_commit:
+    if: >
+      (
+        contains(github.event.pull_request.labels.*.name, 'bot:deploy')
+        || contains(github.event.pull_request.labels.*.name, 'force-ci-checks')
```
casparvl:
```suggestion
        || contains(github.event.pull_request.labels.*.name, 'force-sw-layer-commit-ci')
```

.github/workflows/test_software_layer_scripts.yml
```diff
@@ -0,0 +1,126 @@
...
+          run: |
+            # Print clear error if file doesn't exist at all
+            if [[ ! -f bot/build.sh ]]; then
+              echo "ERROR: File bot/build.sh not found!"
+              exit 1
+            fi
+
+            # Reference checksum
+            # UPDATE THIS CHECKSUM IF AND ONLY IF WE ACTUALLY WANT TO CHANGE bot/build.sh
+            EXPECTED_CHECKSUM="94df53fe4af9c4f28f5c3dc16bf1377e2ba5c7c0df65699d50984b538a1eeab1"
```
bedroge: Because a previous commit (based on my suggestion) changed the comment at the top of this file, the checksum is different now: 
```suggestion
            EXPECTED_CHECKSUM="622e0b4d39543aaf70ee912edd0660c3746d7a1ecb176e0d41df598155e50e21"
```

bedroge (approved):
Lgtm

PR #1355, opened 2026-01-13
{2025.06}[2024a,2024b] Rebuild poetry with new dependency

ocaisa:


bedroge:
For rebuilds we don't use `--robot` (https://github.com/EESSI/software-layer-scripts/blob/main/EESSI-install-software.sh#L407), so I think you will have to add that new dependency first.

bedroge:
Staging PR merged.

PR #1451, opened 2026-03-19
{2025.06}[foss/2025a] ITK 5.4.5

vladaceban:


casparvl:
Discussed in support meeting, Pedro will follow up. Some options:

- It's related to something `netlib`, that doesn't seem to make sense. Maybe that can just be turned off in the build options?
- Could be patched (i.e. a patch to the ITK code), if it's not too much work
- As a last resort, we could build it only for x86_64, but it's used a lot as dependency so this would be extremely annoying

bedroge:
I've opened an issue here: https://github.com/InsightSoftwareConsortium/ITK/issues/6058.

bedroge:
Gave it another try (interactive job) with 5.4.6, but it has the same issue.

PR #1472, opened 2026-04-15
{2025.06}[GCC/12.3.0] pv 1.7.24

fullama:


bedroge:
First of all, thanks for your contribution to EESSI!

The EESSI versions (2023.06, 2025.06, etc) are currently split by toolchains: 2023.06 supports EasyBuild toolchains `foss` 2022b, 2023a, and 2023b, while EESSI 2025.06 only supports `foss` 2024a, 2025a, and 2025b. Since this version of `pv` uses GCC 12.3.0, which is part of toolchain `foss/2023a`, it would have to go into EESSI 2023.06.

If you prefer to have a `pv` installation in EESSI 2025.06, we would first need to get an easyconfig for a newer toolchain merged into EasyBuild. Looking at https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/pv/pv-1.7.24-GCCcore-12.3.0.eb, this should be quite straightforward, since it does not really have any dependencies. At the same time the version of `pv` itself could also be bumped to the latest one. Would you be up for opening such a PR? I can help in getting that merged into EasyBuild.

fullama:
Ah ok! The toolchain stuff had me stumped a bit - I tried to figure it out but obviously I didn't quite get there. I'll try the easybuild PR and see how it goes - thanks for the pointers!

boegel:
@fullama Do you have concrete suggestions for how we could improve our docs to make things clearer?

fullama:
I was thinking about that and am making notes as I go to see if it can help there..

fullama:
ok so here are the unclear parts for me while following https://www.eessi.io/docs/adding_software/opening_pr/:

1) In this part of the docs :https://www.eessi.io/docs/adding_software/opening_pr/#software_layer_pull_request
 it says `Determine the correct easystack file to change` But there was no explaination of how to determine that.
 I picked one of the newest ones because `We strongly prefer that the latest release of EasyBuild that is available at the time is used to add software to EESSI.` should i instead have added it to `eessi-2023.06-eb-5.2.0-2023b.yml`? to me it looked like 2023.06 was deprecated and you wouldnt want to add anything there anymore?

 Could help to add a sentence on how to pick?

2) > ` 2023.06 supports EasyBuild toolchains foss 2022b, 2023a, and 2023b, while EESSI 2025.06 only supports foss 2024a, 2025a, and 2025b`

this i didnt know at all - how does one find that out?

3) > `GCC 12.3.0, which is part of toolchain foss/2023a`
 
 to be fair i should have noticed everything else in the yaml was `GCCcore-14.3.0.` so thats my bad there. Would you look at this for example to know this?: https://docs.easybuild.io/common-toolchains/#common_toolchains_overview_foss if so maybe adding a link could help there?

laraPPr:
@fullama we have noticed that your request has gone stale on our side? We would like to ask if you will still like this. Did you look into EasyBuild for a newer toolchain?

PR #1505, opened 2026-05-20
Restrict missing software check to after deploy label is set

ocaisa:


.github/workflows/test-software.eessi.io.yml
```diff
@@ -110,6 +110,7 @@ jobs:
           echo "EOF" >> "$GITHUB_OUTPUT"
 
   check_missing:
+    if: contains(github.event.pull_request.labels.*.name, 'bot:deploy')
```
ocaisa: Hmm, I think this is wrong, and has no effect

ocaisa:
To whomever reviews this, you can see that the the missing installation test is currently skipped. To force the check to run you add the label  `force-ci-checks`. 

However, I would like to also make sure that we require the missing installations check before a PR is merge-able (which it seems like is not quite the status)

ocaisa:
Actually, it seems the the required checks are indeed held back, this PR should not be mergable even with a positive review

ocaisa:
I'm going to take this further, I plan to

- Only run the consistency check after the missing installations check has succeeded
- Reduce the number of CI jobs to architecture families only (our missing installations and consistency checks only require pure python or shell commands so are actually architecture independent to that level)

.github/workflows/test-software.eessi.io.yml
```diff
@@ -4,10 +4,51 @@ on:
...
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
   workflow_dispatch:
 permissions:
   contents: read # to fetch code (actions/checkout)
 env:
+  # list below should correspond with table @ https://eessi.io/docs/software_layer/cpu_targets/
```
ocaisa, line 12: This really should be a file in the repository that is read so that it can be easily reused by other CI

.github/workflows/test-software.eessi.io.yml
```diff
@@ -4,10 +4,51 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
```
casparvl: Any reason not to add reopened here as well? Also, you might as well one-line this.

```suggestion
    types: [opened, synchronize, reopened, labeled]
```

.github/workflows/test-software.eessi.io.yml
```diff
@@ -106,10 +147,15 @@ jobs:
           
           # Use GITHUB_OUTPUT heredoc correctly
           echo "EESSI_VERSIONS<<EOF" >> "$GITHUB_OUTPUT"
-          echo "$EESSI_VERSIONS" >> "$GITHUB_OUTPUT"
+          echo "2025.06" >> "$GITHUB_OUTPUT"
```
casparvl: I'm confused, why is this version hard-coded now? Or was that part of a test and should that change be reverted?

.github/workflows/test-software.eessi.io.yml
ocaisa: That's a large change that needs to be consistently implemented across all the CI files...should be a follow-up PR

.github/workflows/test-software.eessi.io.yml
ocaisa: Ah, I remember, there are no changed easystacks, I needed to manually set this value to match the condition later

ocaisa: Needs to be reverted before merge

.github/workflows/test-software.eessi.io.yml
casparvl: Agreed.

casparvl:
Ugh, there's a conflict. Could you look into it? I don't want to accidentally select the wrong parts in your PR :)

Edit: nvm, I thought about it a bit, and it was pretty obvious after all. So resolved it.

casparvl (commented):
In my opinion, this PR is almost ready to go. It'll be very helpful to prevent these checks from being run _before_ the `bot:deploy` label is added, since it'll make it much easier for contributors and reviewers to see if a PR is ready to be build/deployed (i.e. all CI should be green - currently that's not the case since all the 'missing' checks are expected to fail in current `main`).

One question we need to resolve: currently we have branch protection rules that state that these 'missing' checks _have_ to pass before a PR can be merged. We'll need to adapt the names, as they have changed (that's why there are 4 CIs marked as 'expected' right now). The other question is: do we just update the names of the CIs that are required to pass? Or do we remove this entirely? The problem with making these new CI names required is that while they run on normal software deployment PRs, the don't normally run on, say, a PR that only updates a workflow file.

I'm wondering if we can somehow solve that by running these CIs in any case, but simply _changing their behavior_ after a deploy (and retriggering them on labels, so that they are rerun after the deploy lable is added). If that's possible, we can simply run an essentially 'no-op' CI before the deploy label is added, and only do the real check once we have a deploy label. Note that the downside is that we'll still incur some latency until these CIs start (even if they are a no-op), and having more CIs run in general might cause our CI to slow down...

I'm honestly not sure what should weigh heavier: being able to enforce this CI to be succesful (at the cost of more CI runs) or being able to reduce the amount of CIs being run. I have a small tendency towards the second: I don't expect any of the reviewers to merge PRs with failing CI anyway...

ocaisa:
The new CI listens to the label `force-ci-checks` as well, so it's always possible to trigger even when a deploy is not required. I would definitely prefer to keep the checks mandatory.

ocaisa:
Hmm, that label is there right now but the CI hasn't run...I will need to check if there is a logic flaw

ocaisa:
Sorry, I am slowly coming back to remembering what this PR is actually doing. Yes, the CI that we require to pass will need to change, they will become names like [`check_missing (ubuntu-24.04, 2023.06)`](https://github.com/EESSI/software-layer/actions/runs/31678271504/job/94377641715#logs) (for all EESSI versions and their Arm equivalents). The CI already covers when these are are not actually needed, they just become skipped, and skipped is equivalent to a pass for a merge status

casparvl:
So, you mean your idea is that on any PR that doesn't change, we'd have to add the `force-ci-checks` label? 

N.B. I think I used that name somewhere as well, but we should then change that name wherever I used it - those don't need to run, I just wanted to give them a know to force them 'just in case' we ever wanted that (e.g. it was useful to test the PR that committed my CI itself).

ocaisa:
I'm sparring with AI to figure the least wasteful way to do this. I need to block PR merges when the checks have not run and I don't want to spin up VMs to do that. I think the right approach is to fail when a force is required:
```
- name: Check whether force-ci-checks is required
  if: >
    needs.check_EESSI_version_changed_files.outputs.EESSI_VERSIONS == ''
    && !contains(github.event.pull_request.labels.*.name, 'force-ci-checks')
  run: |
    echo "::error::No EasyStack changes detected, so missing-installation CI was skipped."
    echo "::error::All PRs require missing-installation CI, so a maintainer needs to add the 'force-ci-checks' label."
    exit 1
```

ocaisa:
Ok, I think this is good to go now. We can now require specific matrix entries in order for CI to be merge-able. For the case where an easystack is _not_ changed, CI will fail and tell you someone with triage rights needs to add the label. This is the rare case and I think it is ok.

casparvl (approved):

PR #1564, opened 2026-08-10
Add CI that leverages `bot:status last_build` to see if all targets have (appropriately) been built for

casparvl:
Add a CI that can check the 'bot:status last_build' output to see if it is complete in the sense that it supports all architectures, and all builds are done from the same software-layer commit.

TODO: 
- actually list the reference architectures it should check for
- make sure it handles multi-architecture GPU builds correctly


ocaisa:
In https://github.com/EESSI/software-layer/blob/main/.github/workflows/test-software.eessi.io.yml there's a maintained list of all the architectures and accelerators for each EESSI version. I think it makes sense to have a source of truth for that list (a yaml file?) and to also use it for this CI.

PR #1576, opened 2026-08-13
Reduce the number of CI jobs when checking for discrepancies, increase consistency with missing installations check, add support for 2026.06 everywhere

ocaisa:


.github/workflows/test_compare_stacks.yml
```diff
@@ -1,89 +1,175 @@
-# documentation: https://help.github.com/en/articles/workflow-syntax-for-github-actions
 name: Check for discrepancies between software stacks in software.eessi.io
+
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
+  workflow_run:
+    workflows:
+      - Check for missing software installations in software.eessi.io
```
ocaisa, line 6: This condition only works in a PR once this CI file is on the default branch, but I manually triggered it at https://github.com/ocaisa/software-layer/actions/runs/31696213226

ocaisa:
For this to work, we support for 2026.06 in the EESSI GitHub Action. That requires that the module is exposed under `/cvmfs/software.eessi.io/init/modules/EESSI` (but it can be hidden in there by adding `hide_version("EESSI/2026.06")` to `/cvmfs/software.eessi.io/init/modules/EESSI/.modulerc.lua`) 

Requires https://github.com/EESSI/github-action-eessi/pull/38 (and a release of the action)

ocaisa:
Closing and reopening to retrigger ci

ocaisa:
This should be good to go now, and I also trigger the discrepancy checker in https://github.com/ocaisa/software-layer/actions/runs/32032422282

bedroge (approved):
Lgtm. There's no `riscv64/generic` in the list of supported CPU targets for 2026.06, but that can be added later (and it's less relevant anyway, since we may want to skip certain installations for RISC-V).

PR #1582, opened 2026-08-18
Fix consistency check for CI in PRs

ocaisa:


ocaisa:
Now we are really shrinking the CI costs!

@bedroge The easiest thing to do to fix the architecture consistency check was to put it into the same workflow. That works quite easily now since I made them pretty consistent anyway...very little extra time cost

ocaisa:
Next target is the licences check, that can probably also be run as part of the same workflow (but in another PR!)

ocaisa:
Also fixed all the Annotation warnings for all CI

bedroge (approved):
🏎️❤️

PR #1389, opened 2026-02-16
{2025.06}[2024a] PyTorch 2.9.1

bedroge:

PR #973, opened 2025-03-20
{2023.06}[2023a] PyTorch v2.1.2 with CUDA/12.1.1

trz42:
New (final?) attempt to build PyTorch/2.1.2 with CUDA/12.1.1

This PR should replace previous attempts:
- #586
- #718
- #825

The PR is based on extensive testing / debugging / analysis on a VM with Haswell CPUs and NVIDIA L40S vGPUs (CUDA compute capability 8.9). It benefits from recently rebuilt CUDA/12.1.1 modules (#919) that added a directory with needed libraries to `$LIBRARY_PATH` to the module files such that the RPATH wrappers used for building software in EESSI add the necessary arguments to the linker command. After that nearly 100 tests of the PyTorch test-suite (which contains about 207k tests) still failed. Most of these 100 tests failed with an error such as
```
Could not load library libcudnn_cnn_train.so.8. Error: libcudnn_cnn_train.so.8: cannot open shared object file: No such file or directory
```
because the library was dynamically loaded by another library of the cuDNN SDK. Fixing this issue would normally be achieved by adding the directory containing `libcudnn_cnn_train.so.8` to `$LD_LIBRARY_PATH`. Because we don't want to do that in EESSI and neither want to change the binary distribution of the cuDNN SDK (which would likely violate cuDNN SDK's license), we chose to work around this by adding a dependency for the above library to `libtorch_cuda.so` which is built when installing PyTorch. This work-around is implemented in commit https://github.com/EESSI/software-layer/commit/4cc89fdd965d75dc0f0f525da7f028289f56cbb1

After that fix, 9+1 tests still failed. These failing tests are
```
dynamo/test_functions 1/1 (1 failed, 167 passed, 2 rerun)
dynamo/test_dynamic_shapes 1/1 (2 failed, 2065 passed, 14 skipped, 33 xfailed, 4 rerun)
distributed/elastic/utils/distributed_test 1/1 (3 failed, 4 passed, 6 rerun)
distributed/test_c10d_common 1/1 (1 unit test(s) failed)
distributed/test_c10d_gloo 1/1 (1 unit test(s) failed)
distributed/test_c10d_nccl 1/1 (1 unit test(s) failed)
+ test_cuda_expandable_segments
```

All failing tests were analysed individually:
- by trying to rerun the tests in the (EasyBuild) build environment, _or_
- by running essential code of the tests in the (EasyBuild) build environment.

[...]

PR #1625, opened 2026-09-07
{2023.06}[foss/2023a] RELION-5.0.0-foss-2023a-CUDA-12.1.1.eb

jpecar:
This one depends on all the goodies: cuda, pytorch, ... expecting problems. Want to use it to learn how deal with these.

### DIFF
--- a/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-5.3.1-2023a.yml
+++ b/easystacks/software.eessi.io/2023.06/eessi-2023.06-eb-5.3.1-2023a.yml
@@ -1,2 +1,3 @@
 easyconfigs:
   - octave-kernel-0.36.0-foss-2023a.eb
+  - RELION-5.0.0-foss-2023a-CUDA-12.1.1.eb


bedroge:
For CUDA builds the bot requires that all CPU dependencies are in place first, because they would otherwise end up in the wrong prefix. Quite a few are still missing for this one, so those would have to be added first.

Furthermore, we usually recommend to do big/complex dependencies (like PyTorch) first (in separate PR). But for PyTorch we will definitely run into issues, we've tried to build this particular version a while ago: https://github.com/EESSI/software-layer/pull/973. In the end we decided to give up and move the effort to a more recent version of PyTorch, see #1389. That version also got somewhat stuck on a few remaining Arm issues, but maybe we can get that ingested soonish with some final pushes. Then we do still need to build a CUDA version as well, which will probably have additional issues.

There's a still open PR for RELION 5.1.0 with a more recent toolchain and PyTorch 2.9.1, so maybe we can ultimately try to get that one built? https://github.com/easybuilders/easybuild-easyconfigs/pull/26470

jpecar:
Ah, wasn't aware of PR for 5.1. Yes, if we already have that, lets go with that one.

